PR #106, opened 2013-05-02
Option for escaping replacement

anderslemke:
Hi jamuhl,

Here at Issuu we're exited about using i18next. Thanks for the great work.

We're proposing this change, based on these assumptions:
- We, as developers, have control of what goes into the dictionaries.
- We, as developers, do not always have control of the values inserted into translations. I.e. it might be from user input.

To avoid the risk of XSS we would like to have the inserted values escaped by default, with an option for not escaping by explicit indication.

This feature is turned off in i18next by default, but you can turn it on, on init or by options with

```
escapeInterpolation: true
```

When turned on, given the following keys in the dictionary

```
"key1": "Not escaped __nameHTML___"
"key2": "Escaped __name___"
```

these are the results

```
t('key1', {name: '<tag>'}); // => Not escaped <tag>
t('key2', {name: '<tag>'}); // => Escaped &lt;tag&gt;
```

What do you think?

### DIFF
--- a/spec/translate/translate.interpolation.spec.js
+++ b/spec/translate/translate.interpolation.spec.js
@@ -116,7 +116,8 @@ describe('default i18next way - with escaping interpolated arguments per default
       translation: {                      
         interpolationTest1: 'added __toAdd__',
         interpolationTest5: 'added __toAddHTML__',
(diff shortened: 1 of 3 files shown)


adrai:
https://github.com/i18next/i18next/discussions/2199

Issue #307, opened 2014-10-14
Default variables


moeriki:
I'm working on a web app with multiple front-ends. (eg. default and business)

Is there currently a way to have a default list of variables. In my case I'd like to have a **SITENAME** I can use anywhere in my localisations.

PR #311, opened 2014-10-25
Added defaultVariables option and tests for it

sergii-trotsiuk:
Implementation of global variable replace #307

### DIFF
--- a/src/i18next.translate.js
+++ b/src/i18next.translate.js
@@ -134,6 +134,10 @@ function _translate(potentialKeys, options) {
         options = options || {};
     }
 
+    if (typeof o.defaultVariables == 'object') {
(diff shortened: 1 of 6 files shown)


src/i18next.translate.js
```diff
@@ -134,6 +134,10 @@ function _translate(potentialKeys, options) {
         options = options || {};
     }
 
+    if (typeof o.defaultVariables == 'object') {
+        options = f.extend({}, o.defaultVariables, options);
```
watadarkstar: This does not allow you to have default nested values. i.e. 

```
defaultVariables: {
      foo: { bar:  'something' },
    },
```

jamuhl: PR welcome for any typescript related fixes / improvements

Issue #334, opened 2014-12-15
Add option to get resource bundle from store


jcbvm:
For my current project it would be nice to have an option to get a resource bundle from the store. 

To be in line with the other API calls ```hasResourceBundle```, ```addResourceBundle``` and ```removeResourceBundle```, I was thinking of something like ```getResourceBundle(lng, ns)``` which will return the current translations from the store for the given language and namespace (or default namespace if omitted).

I can add a PR if needed.

danicomas:
news?

danicomas:
https://github.com/i18next/i18next/pull/335

jamuhl:
PR was merged

PR #335, opened 2014-12-15
Added getResourceBundle to API. #334

jcbvm:
I've added a new API function 'getResourceBundle' which will return the resources of a specific bundle.

The function takes two arguments, a language and an optional namespace. When no namespace is provided, is will use the default namespace taken from the options. This is consistent with other functions like addResourceBundle and removeResourceBundle. It will return the bundle resources or an empty object when the bundle does not exist. It returns a clone of the bundle, to prevent the bundle from being changed.

I've added tests for it.

### DIFF
--- a/spec/init.spec.js
+++ b/spec/init.spec.js
@@ -8,6 +8,8 @@ describe('Initialisation', function() {
 
     //= init/init.addResourceOrResourceBundle.spec.js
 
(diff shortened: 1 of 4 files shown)


danicomas:
When will be released the next version? @jamuhl

jamuhl:
@danicomas not yet planned a next version...might be begin of march.

PR #343, opened 2015-01-02
Allow multiple post-processors

Strajk:
Hi, first of all, thanks for great work on this project... :)

I think it would be good to accept more ( == array) of post-processors. I needed in current project so I have implemented it quickly (but also naively) – IMHO it needs review and nice-fy before potential merge.

It should allow:
```
$.t('example_key', {postProcess: ['markdownify', 'javascriptify']})
```

```
'Ruby is __awesome__'
-> (markdownify) =>
'Ruby is <b>awesome</b>'
-> (javascriptify) => 
'Javascript is <b>awesome</b>'  
```

Of course, it should not break current API and allow one post-processor as string.

What do you think?


jamuhl:
sure that's very welcome...will check this when back from holidays.

PR #346, opened 2015-01-11
Save resources to localStorage when useLocaleStore is true (fixes #345)

### LINKED ISSUE #345: resources added through addResource or addResourceBundle not saved to localStorage
When adding resources through `addResourceBundle` or `addResource`, these resources are not saved to localStorage... 

As a result, calling `setLng` has the unfortunate side effect of blowing them all away, even when using the `useLocalStorage` option as per the suggestion #342.

I can create a PR for this...

tswaters:


jamuhl:
thanks for contributing...will have a look and merge this.

PR #357, opened 2015-02-11
public visibility of applyReplacement function

lcalvy:
Hi,

customLoad option is a great feature.
In angularjs context, it's necessary to use $http in place of jQuery or xmlhttprequest for unit testing.
But it's not DRY to copy applyReplacement method that handle a lot of code to prepare resource url.

Give visibility to function is really useful in conjunction with customLoad and unit testing.

Regards


jamuhl:
is it not sufficient to access that by `i18next.functions.applyReplacement`?!?

PR #369, opened 2015-03-18
Fix CommonJS with jQuery export

ArnaudRinquin:
`module.exports = i18n;` was missing. `require('i18n-client/i18next.commonjs.withJQuery.js')` would return `{}`

The rest is just simplification.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   , "main": "i18next.js"
   , "browser": "i18next.js"
   , "engines": {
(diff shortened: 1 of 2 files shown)


ArnaudRinquin:
Note that this pull request doesn't update the built files, you'd have to recompile / prepare the release.

jamuhl:
1.8.2 published

ArnaudRinquin:
Thanks :+1:

PR #371, opened 2015-03-22
.noConflict() functionality

jairovsky:
Added ```.noConflict()``` function to the public API. This works much like ```jQuery.noConflict()```.

When called, this method renames ```window.i18n``` to ```window.i18next```.

**Note:**
Even when there is no conflicting global, this method *will* rename the ```window.i18n``` global. Use carefully.


jamuhl:
@jairovsky that's great...thanks a lot...will merge this to next version...

PR #372, opened 2015-04-07
Change test command in package.json

lennym:
Running existing command of `testacular start` fails with error `testacular: not found`. Swap this for `grunt test`.

*Note: the tests still seem to be failing - I will investigate further and possbly add more commits to this PR if it's obvious*


lennym:
Updated tests so that they can be fully run in a fresh clone with `npm test`. They are now passing with the addition of phantomjs and a pre-compilation step.

lennym:
Rebased for compatibility with #374

PR #375, opened 2015-04-07
Fix fallback to default namepsace when namespace passed as an option

lennym:
Falling back to default namespace works ok with namespace as a prefix because the prefix is stripped as part of the pass through `_find` in the first instance. However if the namespace is defined as part of the options parameter then it is still present on the options for the second (fallback) call to `_find`.

Fix by setting the namespace explicitly to the default namespace when passing through to the fallback call to `_find`.


lennym:
Is there anything I can do to help get this merged and published? Thanks.

jamuhl:
sorry just not yet got to it...will do a new version in next week and publish that. sorry for the delay.

PR #376, opened 2015-04-08
Expose cache busting option from configuration options

abdavid:
Because HTML5 AppCache will obviously not fetch from cache when busting the url for the resource bundle. 

I could not see any other place that uses cache = true for ajax helper which resides in i18next.helpers.js on line 198. 

The configuration options should probably also receive a rename since it is now just called "cache" to something more explicit.


jamuhl:
will have a look into this - and will merge this for next release.

abdavid:
Thanks :+1:

PR #379, opened 2015-04-11
Allow to fallback to key when trans string is empty

limdauto:
I think it's sensible to display the key (sometimes people use English string or whichever default language they use as key) instead of an empty string.

### DIFF
--- a/src/i18next.defaults.js
+++ b/src/i18next.defaults.js
@@ -16,11 +16,12 @@ var o = {
     fallbackOnNull: true,
     fallbackOnEmpty: false,
     fallbackToDefaultNS: false,
+    showKeyIfEmpty: false,
     nsseparator: ':',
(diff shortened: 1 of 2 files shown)


captolia:
+1

timseverien:
+1

jamuhl:
if i accept this PR the other 50% will complain that they want empty string as an acceptable value (was like this before).

make that depending on a setting provide on init `showKeyIfEmpty` (like fallbackOnEmpty, fallbackOnNull)

limdauto:
@jamuhl cool. I will add that init option. Thanks for replying.

limdauto:
@jamuhl sorry for the delay. I have added the init option. Could you please take a look? Thnx

jamuhl:
perfectm, thank you...will merge this asap.

PR #384, opened 2015-04-27
Null check for detectLngFromLocalStorage

vegetableman:
When  ```useCookie``` is set ```false```, querystring is empty, ```detectLngFromLocalStorage``` is true, throws error on below line because ```localStorage``` returns ```null``` for the key ```i18next_lng``` 

https://github.com/i18next/i18next/blob/30537bee5416644c6a1dbacb19ce5857b037a9f3/src/i18next.detectLanguage.js#L57


jamuhl:
thank you for fixing this....will be merged to next version.

Issue #386, opened 2015-05-03
Config options for timeout?


abepark01:
Hello,

First, thanks for building a great library for i18n.

<strong>Question</strong>
I am loading the translation from a REST API, but I would like to be make sure I can set a timeout value.
If the service is down and X amount of milliseconds has passed, I would like to timeout the request and let i18n just display the string keys.

Is something like this possible currently?
Thank you!
Abe

jamuhl:
PR merged

PR #387, opened 2015-05-03
#386: support for adding timeout in configuration

abepark01:
Add the ability to set timeout in the ajax request.

I would like to be able to set timeout for the ajax requests when fetching the translations.
If the file path is down, then I would like the default fallback to kick in where i18next displays the string key.


jamuhl:
thanks for your contribution...could you rename `option.timeout` to `option.ajaxTimeout`?

will be merged to next version.

abepark01:
Sweet, thank you! I just pushed another commit to rename options.timeout to options.ajaxTimeout.

PR #393, opened 2015-05-24
add nodejs static server to sample folder 

thomasmodeneis:
Hi,

I've added simple nodejs server to the sample folder to make it easy to run this in the browser, so beginners can get up and running faster! :)


jamuhl:
thanks for your contribution.

PR #399, opened 2015-06-12
fix fallback lng option during translations

qraynaud:
The documentation state in features that you can give the `fallbackLng` option during translation but it does not seem to work. I noticed calls to `toLanguages` with 2 arguments (`lng` & `fallbackLng`) but this function did not seem to handle this case properly so I added some code to do it.

Careful: the option fallbackLng still does not work when the `lng` option is not provided during initialization. I added a hack in my code to provide it in all cases because I did not have the time to fix this case too.

### DIFF
--- a/src/i18next.helpers.js
+++ b/src/i18next.helpers.js
@@ -428,8 +428,12 @@ var f = {
         if (lng === 'nb-NO' || lng === 'nn-NO' || lng === 'nb-no' || lng === 'nn-no') lng_index = 1;
         return lng_index;
     },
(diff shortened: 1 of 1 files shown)


jamuhl:
will merge this...wasn't to aware that one can pass fallbackLng on calling translation....might be some addition by contribution.

qraynaud:
thanks ! any ETA?

jamuhl:
might get a new version out this or next week. stay tuned.

PR #402, opened 2015-06-15
Forward the error from sync fetch methods to the init callback function

cutesquirrel:
It works but it's far to be perfect ;)

The init callback function doesn't respect the standard : it should pass the err object as the first argument, and the 't' object as the second.

For the moment, the only way to distinguish error from translate object (t) is to test the type of the argument...

````
$.i18n.init(i18nOption, function(t){  
  // error handling...
  if (typeof(t) === 'string'){ 
    return console.error(t); 
  }

  // Continue if there isn't any error :)
});
````

### DIFF
--- a/src/i18next.init.js
+++ b/src/i18next.init.js
@@ -96,6 +96,8 @@ function init(options, cb) {
         resStore = store;
         initialized = true;
 
+        if (err) return cb(err);
+
         if (cb) cb(lngTranslate);
         if (deferred) deferred.resolve(lngTranslate);
     });
--- a/src/i18next.sync.js
+++ b/src/i18next.sync.js
@@ -13,15 +13,15 @@ sync = {
                         f.extend(store, fetched);
                         sync._storeLocal(fetched);
 
-                        cb(null, store);
+                        cb(err, store);
                     });
                 } else {
-                    cb(null, store);
+                    cb(err, store);
                 }
             });
         } else {
             sync._fetch(lngs, options, function(err, store){
-                cb(null, store);
+                cb(err, store);
             });
         }
     },
@@ -98,7 +98,7 @@ sync = {
         } else {
             // Call this once our translation has returned.
             var loadComplete = function(err, data) {
-                cb(null, data);
+                cb(err, data);
             };
 
             if(typeof options.customLoad == 'function'){

### CODE REVIEW

**Comment by jamuhl** [comment]
not sure if this adds any value...setting init option debug to true will log the error too?

**Comment by cutesquirrel** [comment]
Hi,  yes but for example,  I use i18next on a single page app with
backbone,  and with this fix,  I'm able to handle correctly the error and
react in a good way. It s better than only throw an error in debug mode
only, are you agree?
it leaves the flexibility to do what you want :-)
thanks a lot Jan

**Comment by jamuhl** [comment]
hm...give me some time to think about it...eventual best solution in this case would be to change the callback to `(err, t)`

**Comment by cutesquirrel** [comment]
I agree with you, the only issue with that is you won t conserve the
compatibily. but yes, it would respect the standard.
don t hesitate if you need help
bye
Etienne

**Comment by qraynaud** [comment]
Wouldn't it be nicer to add a second err callback like in the then spec of promises ? something like `init(options, okcb, kocb)` ?

File: src/i18next.init.js [hunk 1]
```diff
@@ -96,6 +96,8 @@ function init(options, cb) {
         resStore = store;
         initialized = true;
 
+        if (err) return cb(err);
+
         if (cb) cb(lngTranslate);
```
qraynaud: then it could become :
```js
if(cb && !err) cb(lngTranslate);
else if (errCb && err) errCb(err);
```

src/i18next.init.js
```diff
@@ -96,6 +96,8 @@ function init(options, cb) {
...
         if (deferred) deferred.resolve(lngTranslate);
```
qraynaud: and here you could change it to
```js
if (deferred) (!err ? deferred.resolve : deferred.reject)(err || lngTranslate);
```

jamuhl:
going the node.js way callback will be (err, t). deferred will resolve or reject on error.

cutesquirrel:
Great ! Thanks :+1:

archer96:
Shouldn't https://github.com/i18next/i18next/blob/master/i18next.js#L969 be e.g.

```js
if (cb) cb(null, lngTranslate);
```

Because everywhere else we have `cb(err, lngTranslate);` except when we don't have to load translation files.

This means we have to use `$.i18n.init(i18nOptions, function(t){ });` or `$.i18n.init(i18nOptions, function(err, t){ };` depending on whether we have translation files or not.

Or maybe I'm missing something? :)

Regards,
Andre

jamuhl:
you're right...won't have the time to fix this today...but will provide a patch asap.

jamuhl:
fixed in 1.10.2

PR #403, opened 2015-06-15
Adds jquery like functionality without the jquery plugin.

AtheistP3ace:
This was a quick solution, something nicer can be worked directly into i18next but I leave it to the creators to know how to best do that.

### DIFF
--- a/src/i18next.init.js
+++ b/src/i18next.init.js
@@ -61,7 +61,12 @@ function init(options, cb) {
     pluralExtensions.setCurrentLng(currentLng);
 
     // add JQuery extensions
-    if ($ && o.setJqueryExt) addJqueryFunct();
+    if ($ && o.setJqueryExt) {
+        addJqueryFunct();
+    }
+    else {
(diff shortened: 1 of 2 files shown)


jamuhl:
hey cool...think we could use this as a starting point and see where it goes from there. looks good to me...will do some testing and merge this for next release...(this or next week i would get to publish an update).

thanks again for you contribution...really happy to get this into i18next.

AtheistP3ace:
Absolutely. Happy to help!

jamuhl:
@AtheistP3ace upcoming v2 will have the selector stuff not anylonger integrated...http://i18next.github.io/i18next.com/ it will act more like a core component with an ecosystem around it. eg. see the old jquery support will be solved in own repo: https://github.com/i18next/i18next-jquery

would be great to see this outside again - not sure if you got time to extract this to own repo, like i did for jquery support.

but if you have, feel free to add it here: http://i18next.github.io/i18next.com/docs/ecosystem/#frameworks

PR #427, opened 2015-09-02
Fixed mnk plural definition

jeanregisser:
I had initially generated [i18next/i18next-ios](https://github.com/i18next/i18next-ios) plural rules from the javascript rules. And today after enabling a warning to spot dead code, I found this issue with the mnk definition. Hence I'm submitting this trivial patch ;)


jamuhl:
will be merged for next version. thanks.

PR #431, opened 2015-09-11
Add support to override Ajax HTTP headers

hippostar:
With this patch it is possible to include a "headers" option when calling i18n.init() which may be used to override the default content-type and accept headers or to provide additional custom HTTP headers.

Example usage:

i18n.init({
headers: {'content-type': 'application/json'}
});


jamuhl:
great...will be merged for next version.

PR #438, opened 2015-09-28
Runs multiple post processes for missing translations

pekala:
I found out that multiple post processes are not applied for the case of missing value. This should fix it. Let me know if I need to add anything.


jamuhl:
cool...thanks a lot...will be merged into next version.

PR #440, opened 2015-10-09
`resource` to `resources` in addResources function

max-mykhailenko:
Possible bug, please, check it


jamuhl:
yes - will be merged for next version. thanks for finding.

PR #442, opened 2015-10-16
Pluralization with a count of 1 is broken

mehcode:
This fixes it for me. I have no idea why and don't have the time to investigate. I realize this isn't mergable as is as I'm touching a built file. However this isn't for merge but more to let you know that it is broken and the solution is probably somewhere around what I did.

### DIFF
--- a/i18next.js
+++ b/i18next.js
@@ -2071,7 +2071,7 @@
             if (ext && ext.numbers.length <= 1) {
                 return false;
             } else {
-                return this.get(lng, count) !== 1;
+                return true;
             }
         },
     


jamuhl:
for which language you run into this issue?

mehcode:
>  for which language you run into this issue?

I don't think that the language matters at all. I could be wrong but.. this is i18n.

For the record my fallback language is set to `en` and I was attempting to translate it to `de` (both JSON files are present, in the right spot, etc.). If I do `t("key", {count: 2})` it works. If I do `t("key", {count: 0})` ir works. If I do `t("key", {count: 1})` it remains in the `en` language.

jamuhl:
there is a test for that, which is not failing: https://github.com/i18next/i18next/blob/master/spec/translate/translate.plurals.spec.js#L41

ApfelUser:
Same here for Korean

ApfelUser:
Actually. **Just** for Korean. Not for German. It is falling back to English in the case of Korean ( `kr` )

jamuhl:
@ApfelUser korean language code is `ko` or at least http://www.w3schools.com/tags/ref_language_codes.asp says so.

ApfelUser:
Changing it to `ko` makes it always use the **1** version of the key. Also if the count is e.g. 3.

jamuhl:
korean has no plural forms, see the rule: https://github.com/i18next/i18next/blob/master/src/i18next.plurals.js#L66

jamuhl:
closing this due upcoming v2: http://i18next.github.io/i18next.com/

PRs welcome

PR #443, opened 2015-10-18
Reimplement XSS-vulnerable sequential replacement code

KJTsanaktsidis:
The interpolation resolution code inside translate.js is looping over
each key in the interpolation dictionary and applying the replacements
one at a time. This means that if untrusted user input contains the name
of another key in the dictionary, for example `'__lastName__'`, an
unexpected interpolation can happen:

````
i18n.t('__firstName__ __lastName__', {
    firstName: '__lastName__',
    lastName: 'foo',
});
// equals "foo foo"
````

This combines with the unescaping suffix feature to cause XSS injection,
even with escapeInterpolation set to true. For example,

````
i18n.t('__firstName__ __lastName__', {
    escapeInterpolation: true,
    firstName: '__lastNameHTML__',
    lastName: '<script>',
});
// equals "<script> &lt;script&gt;"
````

This PR fixes that by performing the replacements *all at the same
time*. This means that untrusted input cannot cause these unexpected
extra interpolations.

### DIFF
--- a/spec/translate/translate.interpolation.spec.js
+++ b/spec/translate/translate.interpolation.spec.js
@@ -128,7 +128,8 @@ describe('default i18next way - with escaping interpolated arguments per default
         interpolationTest1: 'added __toAdd__',
         interpolationTest5: 'added __toAddHTML__',
         interpolationTest6: 'added __child.oneHTML__',
-        interpolationTest7: 'added __toAddHTML__ __toAdd__'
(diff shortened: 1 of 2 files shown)


jamuhl:
will be merged asap...hopefully on monday. will release updated versions. thanks again for finding and solving that one.

jamuhl:
update is out v1.10.3

jamuhl:
thanks again for finding and fast fix

Issue #331, opened 2014-12-09
Merge and overwrite resources


RaptorCZ:
Hi, is there any way how to merge resources with "overwrite" option?

Let's say I have static resource string in json files, but I have service returning overwrites and new values. So I need to init framework with static files then merge values. If string doesn't exists - add it. Otherwise replace old value with new one.

My translation.json files are for example:
```javascript
 {
    "BaseResource":
    {
        "AccessDenied": "Acces denied!",
        "Action": "Action",
        "ActionName": "Action "
    },
    
    "Storage":
    {
        "Advertising": "Advertising",
        "All": "-------All-------",
        "AllItems": "All",
        "AutoOwn": "Auto Own"
    }
}
```
Service returns the same JSON structure.

iandarlingmnetics:
I've not been able to run the tests on that commit so far (although I have added one), hence there's not a pull request yet.

jamuhl:
cool...looks like this lands soon :+1:

etreworgy:
There actually is a bug here in the existing _deepExtend (even if you don't intend to allow overwriting).  The function just iterates over 'props in source', and if you happen to reach a leaf of either target or source translation objects you'll get unintended behavior that ultimately results in an error.

If the target hits a leaf string then the '(prop in target)' test fails with an error because JS doesn't let you use 'in' with strings.  If the source happens to hit a leaf string first then the 'for(var prop in source)' starts iterating over the characters and string prototype functions such as 'hashCode'.

Both cases are not expected or documented behavior.  There should be a simple string test on source[prop] and target[prop] to prevent this from happening and then either overwrite or ignore depending on the desired behavior.

jamuhl:
landed in v1.10.3

kdydesign:
Hi, I needed the same option.
I corrected the " ResourceStore.prototype.getResource" function slightly.

```javascript
ResourceStore.prototype.getResource = function getResource(lng, ns, key) {
      var options = arguments.length > 3 && arguments[3] !== undefined ? arguments[3] : {};
      var self = this;
      var keySeparator = options.keySeparator || this.options.keySeparator;
      var originResource;
      if (keySeparator === undefined) keySeparator = '.';

      var path = [lng, ($.isArray(ns) ? ns[0] : ns)];
      if (key && typeof key !== 'string') path = path.concat(key);
      if (key && typeof key === 'string') path = path.concat(keySeparator ? key.split(keySeparator) : key);

[...]

PR #414, opened 2015-07-13
Issue 331 - Merge and overwrite resources

iandarlingmnetics:
Tests have been run in PhantomJS (test:ci/dev)

There's an issue-331-release branch I've pushed too - you can probably ignore that one - I'm only using it for deployment purposes until the code fix lands in master.

PR #444, opened 2015-10-19
Fixing _deepExtend to handle keys deep existing in source and target

etreworgy:
Fixing issue in _deepExtend as well as adding 'overwrite' flag to specify behavior (As requested in Issue #331).

Fix will still work without exposing the 'overwrite' flag, default behavior is to not overwrite when doing a deep extend.


jamuhl:
is this a duplicate of https://github.com/i18next/i18next/pull/414 ?!?

etreworgy:
It add the same overwrite functionality, but #414 doesn't fix the existing _deepExtend method (which I commented about in #331).  It simply adds a new one that is used when overwriting is desired.  

On top of that, the new _deepExtendOverwrite method has the same issue as the current _deepExtend method (though it is unlikely anyone would encounter it).  Set up a test where the target translation json has a path like 'a.b.c.d' and then try to overwrite it with a new bundle that has 'a.b.c'.  target[b] will exist and not be a string so it will recurse down to another level of _deepExtendOverwrite where target = {c: 'd'} and source = 'c'.  The for(props in source) will start traversing over the string prototype properties like 'hashCode'.

PR #446, opened 2015-10-26
Add nsseparator and keyseparator as options to translation function

pekala:
It is in some cases useful to have a different `keyseparator` for some of the translations. In our case, we moved to using English message as the key, so we had to change the `keyseparator` to something like `:::` (which will not occur in regular message).


jamuhl:
will be merged for next release

PR #453, opened 2015-11-11
Fix use of sprintf as shortcutFunction when first argument falsey

### LINKED ISSUE #451: sprintf does not work with shortcutFunction and falsey first argument
As shown in [this jsfiddle](http://jsfiddle.net/dLc7x/56/).

To reproduce:

1. Set shortcut function to `sprintf`
2. Attempt to do a translation with a falsey first sprintf argument (ex. `$.t('There are %d items', 0)`)
3. Notice the output does not  replace the token.

It's due to [this line](https://github.com/i18next/i18next/blob/master/i18next.js#L1507). You should probably check to see if the shortcutFunction is set here before assuming the second argument is supposed to be the options object.

benjyblack:
Do not overwrite first sprintf argument when it is falsey. Also,
fix bad falsey check in _translate.

Resolves #451

### DIFF
--- a/spec/translate/translate.sprintf.spec.js
+++ b/spec/translate/translate.sprintf.spec.js
@@ -7,7 +7,8 @@ describe('using sprintf', function() {
       translation: {                      
         interpolationTest1: 'The first 4 letters of the english alphabet are: %s, %s, %s and %s',
         interpolationTest2: 'Hello %(users[0].name)s, %(users[1].name)s and %(users[2].name)s',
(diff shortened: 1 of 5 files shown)


benjyblack:
Not very familiar with your build process so forgive me if there's copied code where there shouldn't be.

jamuhl:
will be merged for next version...thanks for providing the fix

jamuhl:
will be in 1.11.1

PR #515, opened 2016-01-24
Add defaultPluralValue option

pekala:
At the moment, it is not possible to provide fallback for plural translations. This change adds `defaultPluralValue` option to translation function. For example, if `someKey` is not in dictionary:
```javascript
i18n.t('someKey', { 
    count: 1, defaultValue: 'Singular', defaultPluralValue: 'Plural'
}); // "Singular"

i18n.t('someKey', {
    count: 2, defaultValue: 'Singular', defaultPluralValue: 'Plural'
}); // "Plural"
```

I know that v1 is now deprecated, but I still have to use it, thus PR for this version. But v2 could probably use the same change (I couldn't find solution for this problem in v2 docs).

### DIFF
--- a/spec/translate/translate.plurals.spec.js
+++ b/spec/translate/translate.plurals.spec.js
@@ -2,25 +2,25 @@ describe('plural usage', function() {
 
   describe('basic usage - singular and plural form', function() {
     var resStore = {
-      dev: { 'ns.2': {                      
+      dev: { 'ns.2': {
             pluralTest: 'singular from ns.2',
             pluralTest_plural: 'plural from ns.2',
             pluralTestWithCount: '__count__ item from ns.2',
             pluralTestWithCount_plural: '__count__ items from ns.2'
         }},
-      en: { },            
-      'en-US': { 
-        'ns.1': {                      
+      en: { },
+      'en-US': {
+        'ns.1': {
             pluralTest: 'singular',
             pluralTest_plural: 'plural',
             pluralTestWithCount: '__count__ item',
             pluralTestWithCount_plural: '__count__ items'
-        } 
+        }
       }
     };
-    
+
     beforeEach(function(done) {
-      i18n.init(i18n.functions.extend(opts, { 
+      i18n.init(i18n.functions.extend(opts, {
           resStore: resStore,
           ns: { namespaces: ['ns.1', 'ns.2'], defaultNs: 'ns.1'}
         }),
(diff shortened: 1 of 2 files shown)


jamuhl:
what's the usecase for this...defaultValue is just some placeholder, reminder, prefill (better then nothing) value for not yet done translations. or do i miss something.

to do this proper you would need to provide pluralValues for languages having multiple pluralforms, too.

pekala:
You are right about the `pluralValues` being the right solution, this is just to start the discussion for a more general solution (and at the same time it solves my usecase). However, this came up as a result of how I use `i18next` and might be to niche to put into the library.

I don't use `i18next.t` directly in my code, instead I use helper functions that have footprint that is expected by `xgettext` parser, i.e. `_('Always singular')` and `_n('This is singular', 'This is plural', { count: 2 })` These two functions `_` and `_n` internally call `i18next` and set `defaultValue` and `defaultPluralValue` so that for english I don't have provide any dictionary (it always uses the fallback). For other languages it just ignores these default values. Without `defaultPluralValue` (or `pluralValues` as you mentioned) this would not be possible.

jamuhl:
why not just:

```
function _n(singular, plural, options) {
  var res =  i18next.t(singular, options);
  if (options.count !== 1 && res === singular) res = plural;
  return res;
}
```

pekala:
Hmm... this would not work if you have variable interpolation inside plural, right? But sure something along these lines would work, but it's arguably more hacky as it leaks the implementation details of detecting plurals. Of course, my PR as of now is also not correct implementation, but if adjusted to account for more complex plural rules would be cleaner. 

For example, if your default language is Polish, you would use different footprint of the `_n` function and would have to reimplement the logic for plurals inside `_n`.

jamuhl:
API exports applyReplacement https://github.com/i18next/i18next/blob/1.11.x/src/i18next.api.js#L25

```
function _n(singular, plural, options) {
  options.defaultValue = singular;
  var res =  i18next.t(singular, options);
  if (options.count !== 1 && (res === singular || res === i18next.applyReplacement(singular, options)) res = plural;
  return res;
}
```

while i could add this to v1 i don't expect this to get into v2...so i still prefer to handle this edge case outside of i18next even more it only applies to languages having singular/plural only.

mhawk0:
what's the v3 API equivalent to applyReplacement?

jamuhl:
you can use: `i18next.services.interpolator.interpolate('replace {{var}}', { var: 'me' })`;

PR #540, opened 2016-02-02
Include key and value in 'added' event if adding a single resource

mogelbrod:
I've got a use case where some users are allowed to update the translations, which works fine using
`i18next.addResource(i18next.language, ns, key, translation)`. I'd like to listen to these changes, but unfortunately the `added` event emitted by `i18next.store` (docs seem to be incorrect [stating that it's emitted by i18next](http://i18next.com/docs/api/#on-added)) doesn't specify the key being added/updated.

This PR makes event listeners also receive key, along with value in case it could be useful. The latter could be removed if undesired.

### DIFF
--- a/src/ResourceStore.js
+++ b/src/ResourceStore.js
@@ -49,7 +49,7 @@ class ResourceStore extends EventEmitter {
 
     utils.setPath(this.data, path, value);
 
-    if (!options.silent) this.emit('added', lng, ns);
+    if (!options.silent) this.emit('added', lng, ns, key, value);
   }
 
   addResources(lng, ns, resources) {


jamuhl:
will be merged and published with next version...

mogelbrod:
Awesome, thanks for the quick reply!

jamuhl:
in v2.0.25

mogelbrod:
Appreciate it!

mogelbrod:
I might be missing something, but the v2.0.25 published to NPM doesn't appear to have the latest changes compiled:
https://github.com/i18next/i18next/blob/master/lib/ResourceStore.js#L93

The gulp `publish` task should maybe depend on the `build` task to avoid similar issues in the future.

jamuhl:
v2.0.26 sorry for that...yes need to rework this. more or less planned to remove gulp in favour to npm run anyway.

mogelbrod:
No worries - thanks a lot for the quick resolution :)

PR #564, opened 2016-02-16
Change point to main file

### LINKED ISSUE #477: webpack pre-built javascript warning
When I use webpack to bundle, it complaint something like this.
```
WARNING in ./~/i18next/bin/index.js
Critical dependencies:
1:409-416 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/i18next/bin/index.js 1:409-416
```
To clear this warning, I need to change the `package.json`  file.
```
"main": "./lib/index.js",
```
or `import 'i18next' from 'i18next/lib/index.js`.
I`m not sure this warning need to optimize.
Thanks for your time.

just-boris:
As I see, it can't break anything
fixes #477

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "i18next",
   "version": "2.1.0",
   "description": "i18next internationalization framework",
-  "main": "./bin/index.js",
+  "main": "./lib/index.js",
   "keywords": [
     "i18next",
     "internationalization",


jamuhl:
will change this for next minor version. might be best solution for now.

just-boris:
That will be good for me.
Thanks!

jamuhl:
v2.2.0 published

mathieumg:
@jamuhl  Are there plans to publish the same change for the official backends? Are you open to PRs for that change? Thanks!

jamuhl:
@mathieumg yes would do that when getting to it...but PRs would be great (would speed it up).

PR #581, opened 2016-02-24
Fixes detection when using both context and pluralization and context not found.

rumbcam:
Fixes detection when using both context and pluralization and context not found. It now will fallback to default plural if no context is found.

Possible key list order:
key_context_plural
key_context
key_plural
key


jamuhl:
will be merged into next version...thank you for providing the fix

jamuhl:
will be in v2.3.0 published in a few minutes

PR #597, opened 2016-03-09
Change defaultValue behavior

mogelbrod:
I figured I'd use `i18next.t()` to translate some optional strings after finding the `defaultValue` option. It didn't behave as I'd expected though, which would be that passing an empty string or null would result in nothing being returned if the translation key hadn't been defined. Instead I always got the key back, along with a `missingKey` log message.

Considering the fact that the developer has explicitly specified a default fallback value I would assume that missing translations are somewhat expected and therefore shouldn't log anything.

This PR changes the behaviour to only consider translations as missing if no defaultValue has been provided.

I also made the presence checking of the defaultValue stricter so that all values except for `undefined` is considered and then checked against with `isValidLookup()`, allowing them to be ignored through the `returnNull`and `returnEmptyString` options.

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -95,12 +95,14 @@ class Translator extends EventEmitter {
     }
     // string, empty or null
     else {
+      let usedKey = false;
       // fallback value
       if (!this.isValidLookup(res) && options.defaultValue !== undefined) {
         res = options.defaultValue;
       }
       if (!this.isValidLookup(res)) {
         res = key;
+        usedKey = true;
 
         this.logger.log('missingKey', lng, namespace, key, res);
 
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -95,23 +95,16 @@ class Translator extends EventEmitter {
     }
     // string, empty or null
     else {
-      let usedDefault = false,
-          usedKey = false;
-
       // fallback value
-      if (!this.isValidLookup(res) && options.defaultValue) {
-        usedDefault = true;
+      if (!this.isValidLookup(res) && options.defaultValue !== undefined) {
         res = options.defaultValue;
       }
       if (!this.isValidLookup(res)) {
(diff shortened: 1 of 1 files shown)


jamuhl:
the defaultValue does not define a fallbackValue if not founded.

it's the value that should be transmitted with missing key.

behaviour should be:
- send missing key
- return defaultValue

https://github.com/i18next/i18next/blob/master/src/Translator.js#L102
test: https://github.com/i18next/i18next/blob/master/test/backward/v1.11.1.compat.js#L1414

but to allow empty or null as defaultValue we would need to change https://github.com/i18next/i18next/blob/master/src/Translator.js#L102 to check for options.defaultValue !== undefined

mogelbrod:
I'm not really following, did you perhaps mistake this PR for an issue?
If `missingKey` by default always should be logged, even when looking up a key with a `defaultValue` present, would it be possible to add an option to suppress this? I don't really see any use case where the developer has specified a default value but still expect the key to exist.

jamuhl:
i don't get your usecase:

you want a key to not exist and show the defaultValue instead. why you use t function and not just type the defaultValue there hardcoded?

like said. As a developer not having control over the backend i can specifiy a new key and it's defaultValue -> those get passed via missing to backend where that key gets added with specified default. (From that point the defaultValue is only useful on backend outage)

mogelbrod:
Ah, sorry! My use case was some parts of an application where some language translations would include text while others shouldn't.

The reason for this PR was that I thought the `defaultValue` behaved differently from what I had initially expected. However, after thinking it through once more I've figured that this change is totally unnecessary since translations which should be empty can just specify `null` or an empty string,
and that the `missingKey` logging is obviously doing the right thing since the key is missing from the translations object :smile: 

Dropping this PR, thanks for the quick responses as always!

PR #645, opened 2016-04-05
Allow falsy defaultValue

jonapgar:
Need to allow falsy defaultValue such as '' or null to be used in conjunction with options.returnEmptyString or options.returnNull

### DIFF
--- a/i18next.js
+++ b/i18next.js
@@ -734,7 +734,7 @@
 	              usedKey = false;
 
 	          // fallback value
-	          if (!this.isValidLookup(res) && options.defaultValue) {
+	          if (!this.isValidLookup(res) && 'defaultValue' in options) {
 	            usedDefault = true;
 	            res = options.defaultValue;
 	          }


jonapgar:
This allows you to do something like this, which works in v1.x

var phrase = t('ns:missing',{defaultValue:null}) || somethingExpensive();

or 

var phrase = t('ns:foo') +  t('ns:missing',{defaultValue:'}) + t('ns:bar');

jamuhl:
this allows `undefined` as value too. not sure if we should limit to `''` and `null`. As undefined would need special handling in backends (as not JSON serializable).

so check should be `&& options.defaultValue !== undefined`

jonapgar:
Yeah I thought about that too. In any event the undefined would not pass
the next  *isValidLookup *condition a few lines down.

jamuhl:
can you change the PR to `&& options.defaultValue !== undefined`? can be merged after that change.

jamuhl:
i18next@2.5.0 allows all but undefined

PR #656, opened 2016-04-12
Bug in RTL lang detection

netray:
array consists of rtl langs not ltr

also indexOf returns -1 when not found, and 0+ if found, so simple if was returning true for all languages not mentioned in this array and ar (index 0).


jamuhl:
i18next@2.5.1

netray:
thanks!, justed in 2.5.1 - works fine now

PR #741, opened 2016-06-20
Allow fallback to language only when using whitelist

cbikar:


jamuhl:
great...would you mind adding some tests here: https://github.com/i18next/i18next/blob/master/test/languageUtils.spec.js

afterwards i can merge and publish a new version

jamuhl:
just published i18next@3.3.0

jamuhl:
@cbikar do you still remember what the use case of this feature was...looks somehow wrong to me nowadays: if i set nonExplicitWhitelist to true i would expect: https://github.com/i18next/i18next/pull/741/files#diff-8700466ad5cf3eb27e9887663933fc14R165 to be ['de-AT', 'de', 'en'] but basically the behaviour is the same as without nonExplicitWhitelist set to false

jamuhl:
@cbikar basically why do we use nonExplicitWhitelist only for fallbacks not for the detected language?

cbikar:
@jamuhl the use-case was to prevent requests to the server that check if the language file is available and to do correct fallbacks. In your example (user language is 'de-AT'), if `nonExplicitWhitelist` is set to `true`, it checks if 'de-AT' is in the whitelist, if so, it uses the language, if not it falls back to 'de', checks if 'de' is in the whitelist, if so, uses that language, if not falls back to 'en'. Without the `nonExplicitWhitelist` I think i18n would directly fall back to 'en' without checking if 'de' is available.

jamuhl:
In this case the implementation is wrong...nonExplicitWhitelist is triggered only for the fallbackLng case not for the language part fallback de-AT -> de

https://github.com/i18next/i18next/pull/741/files#diff-3e24e32b6f33ebb43ce089a61f08d6d8R75

would take 'de' from 'de-AT' and pass in any case

https://github.com/i18next/i18next/pull/741/files#diff-3e24e32b6f33ebb43ce089a61f08d6d8R74 will take de-AT but not pass on nonExplicitWhitelist neither as `exactMatch` override triggers.

really got some question marks on this...might need to change this so de-AT passes if nonExplicitWhitelist is enabled.

PR #746, opened 2016-06-22
fixed an issue with several unescaped key in the interpolation string

de-robat:
corresponding pull request for issue: https://github.com/i18next/i18next/issues/747 .
please review and let me know what you think. Thanks!

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -48,15 +48,13 @@ class Interpolator {
     }
 
     // unescape if has unescapePrefix/Suffix
-    var unescapedReplacements = str;
     while(match = this.regexpUnescape.exec(str)) {
       let value = utils.getPath(data, match[1].trim());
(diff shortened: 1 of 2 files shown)


jamuhl:
there is currently one test failing (compatibility) - will look into this tomorrow

de-robat:
yep i saw it and ressorted to the old way of going with "lastIndex=0" for the unescaped case as well. this leaves the backwards compatibility untouched. Hope the reassesment of travis will proof that :D unfortuneatley i just test with the "npm run tdd" test in the beginning. sry for the inconveniences!

jamuhl:
np...got no time today...so expect it to land tomorrow

de-robat:
ok, tomorrow will be sufficient for me :) Once more: ty!

jamuhl:
thanks you for the contribution: just published i18next@3.3.1 including this fix.

PR #784, opened 2016-07-13
Key based fallback plural support

7Pass:
I'd recommend to have pluralHint option so that i18next will fallback on this value as this is a very frequently used `gettext` option. With this, 
```
_n("{{count}} apple", "{{count}} apples", items);
```
becomes
```
i18n.t('{{count}} apple', {
  count: items,
  pluralHint: '{{count}} apples'
});
```

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -106,6 +106,9 @@ class Translator extends EventEmitter {
       if (!this.isValidLookup(res)) {
         usedKey = true;
         res = key;
+        
+        // Fall back on plural hint
+        if (options.pluralHint && options.count && options.count > 1) res = options.pluralHint;
       }
 
       // save missing


jamuhl:
Why not just use:

```
i18n.t('{{count}} apple', {
  count: items,
  defaultValue: '{{count}} apples'
});
```

seems to lead to same result?

7Pass:
I've tried the same, if "{{count}} apple" is not found, event with `items` > 1, "{{count}} apple" is used.

jamuhl:
ok i see...basically useful if you forgot the plural form in a language but got the singular defined. So instead of a singular a default plural will be shown.

Better would be not forgetting the plurals obviously ;)...but we can merge this if needed by you.

By the way using http://locize.com service would assert you not forget a key...you might check that out - helps us to fund the work on i18next, too ;)

jamuhl:
Would you mind adding a test case here: https://github.com/i18next/i18next/tree/master/test/translator ?

jamuhl:
closing...as no tests were provided.

PR #791, opened 2016-07-18
Fix link in readme.md

guillaumevincent:
`http://locize.com/` instead of `locize.com` because github interpret it in `https://github.com/i18next/i18next/blob/master/locize.com`


jamuhl:
thank you

PR #824, opened 2016-10-11
Fix to get the updated data values for version 1.11.x with jquery.

mrgmatthews:
So if the data label/options are changed its reflected in the i18n

### DIFF
--- a/src/i18next.jquery.js
+++ b/src/i18next.jquery.js
@@ -56,7 +56,12 @@ function addJqueryFunct() {
         }
 
         if (!options && o.useDataAttrOptions === true) {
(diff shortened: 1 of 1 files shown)


jamuhl:
don't get what this change provides...?!? why reading from attribute instead of reading data?

mrgmatthews:
So using a third party update method (Angular) in this case. It changes the data-i18n-options value. Using data doesnt get the updated value it gets from the cached value.

jamuhl:
published v1.11.5. thanks for taking the time to explain. please be aware that v1.x is outdated and won't be supported in future anymore.

PR #826, opened 2016-10-13
Fix Interpolator.escapeValue defaulting to undefined in some cases

alexmchardy:
I found that if I pass any `interpolation` options without including `escapeValue`, `this.escapeValue` will get set to `undefined` in `Interpolator`. The documentation indicates `escapeValue` defaults to `true`, so this could cause variables to not be escaped as expected in some cases.

This config, for example, will result in `i18next` not escaping variables as expected:
```
i18next.init({
  interpolation: {
    prefix: "__",
    suffix: "__"
  },
  ...
});
```
It's worth noting that it could potentially be a XSS security issue for users of `i18next` if they believe variables are getting escaped when they actually aren't.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -17,7 +17,7 @@ class Interpolator {
 
     const iOpts = options.interpolation;
 
-    this.escapeValue = iOpts.escapeValue;
+    this.escapeValue = iOpts.escapeValue !== undefined ? iOpts.escapeValue : true;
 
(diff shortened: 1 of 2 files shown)


alexmchardy:
BTW, for bonus points, I added a bunch of tests around the `Interpolator` options.

jamuhl:
thanks a lot: merged and published in i18next@3.4.4

PR #830, opened 2016-10-26
Add module entry in package.json

mull:
See https://github.com/webpack/webpack/issues/1979

This seems to be the preferred way to indicate where ES6 modules reside now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "i18next internationalization framework",
   "main": "./index.js",
   "jsnext:main": "dist/es/index.js",
+  "module": "dist/es/index.js",
   "keywords": [
     "i18next",
     "internationalization",


jamuhl:
will merge this asap - but will take me some time....will need to update all other plugins,... too

thanks for getting my attention for this change.

mull:
Thanks for merging @jamuhl!

jamuhl:
glad you added this

jamuhl:
FYI: https://github.com/i18next/i18next/issues/836#issuecomment-259517546

mull:
Sorry for causing so many problems 😁

jamuhl:
that's ok...just right now i'm unsure if there is something wrong or if it's just misconfiguration on users side...but breaking major version should give me the time to figure out....

and better solved today then later when all use webpack2 ;)

Issue #833, opened 2016-11-03
Why do not create an new Translator instance when invoke cloneInstance?


anson0370:
```js
  cloneInstance(options = {}, callback) {
    let clone = new I18n({...options, ...this.options, ...{isClone: true}}, callback);
    const membersToCopy = ['store', 'translator', 'services', 'language'];
    membersToCopy.forEach(m => {
      clone[m] = this[m];
    });

    return clone;
  }
```

Every clone of i18next have the same translator instance. If I use some code like below:
```js
const i18n = i18next.cloneInstance()
i18n.changeLanguage(someDetectedLanguage)
await someAsyncActionHere()
i18n.t('translate text')
```
The behavior is unpredictable. Because all requests change the same translator's language.

jamuhl:
Makes sense...if you like you might provide a PR? Else i will try to change this for a future release...but honestly might take some time - as i'm hard on time.

anson0370:
Of cause. I'll make a PR this days. 😄

jamuhl:
merged

PR #834, opened 2016-11-04
Create an new translator when clone i18next instance

anson0370:
I'm not sure is this the best way to fix it. And I reran all the tests, all passed.

#833


jamuhl:
looks right for me...will merge this asap...and publish a new version (not sure but might be a breaking version - but would prefer patch - as it fixes behaviour)

PR #843, opened 2016-11-19
Custom escape function, single-quotes in nested

Programmix:
* Adds the option to provide a custom escape function for interpolation
* Converts single-quotes in nesting to double-quotes for JSON parsing


jamuhl:
Would you mind adding a test: https://github.com/i18next/i18next/blob/master/test/interpolation.spec.js#L156

jamuhl:
thank you...merged and published in i18next@4.1.0

PR #863, opened 2017-01-19
Return namespace in cimode with appendNamespaceToCIMode option

aray12:


src/Translator.js
```diff
@@ -56,17 +56,25 @@ class Translator extends EventEmitter {
     if (typeof keys === 'number') keys = String(keys);
     if (typeof keys === 'string') keys = [keys];
 
-    // return key on CIMode
-    let lng = options.lng || this.language;
-    if (lng && lng.toLowerCase() === 'cimode') return keys[keys.length - 1];
```
aray12: My tests for when `appendNamespaceToCIMode: true` failed. This appears to be returning everything including the namespace if it is passed directly in the key (it is called before `this.extractFromKey`. I am not exactly sure why because this is the behavior I get in my application.

Below I started using `key` from `this.exctractFromKey`. Let me know if that was a bad idea.

jamuhl:
Thanks for the PR...your change is ok...just need to think if i need to make this a breaking version - as we get different results now. For sure like this more as no random ns:key or key only depending on the calling code.

aray12:
We use `i18next` and we make heavy use of namespaces to break files up. The weird thing is I don't get inconsistent behavior in our application even though we specify namespace in the key a lot.

e.g. in `cimode`
```
t('ns:key') => key
```


 I looked at versions as well as file history and I have no idea why. Anyway I am glad you approve of this approach. Since the behavior is different I would recommend a major version bump in case anyone depends on existing behavior. (unfortunately)

aray12:
I also created a PR for i18next.com here:

https://github.com/i18next/i18next.com/pull/17

jamuhl:
just published i18next@6.0.0

PR #870, opened 2017-02-01
Removed special cases for norwegian

henkeson:
This special case handling prevents from dealing with "nb" and "nn" as different languages, which they are.
I think it should be up to the user to define "no" as fallback language for "nb" and "nn" if they want to. It should not be enforced.

### DIFF
--- a/src/LanguageUtils.js
+++ b/src/LanguageUtils.js
@@ -15,9 +15,8 @@ class LanguageUtil {
   getLanguagePartFromCode(code) {
     if (code.indexOf('-') < 0) return code;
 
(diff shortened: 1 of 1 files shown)


jamuhl:
agree with you...this 'workaround' came from a time where you could not define fallbacks depending on language (just one fallback). So it makes sense to remove this and creating a breaking major version. Thank you for taking time to coming back to this code lines (almost forgot that extra treatment) and taking the time to create a PR. Really appreciate that.

Give me some time to merge and publish...

jamuhl:
published in i18next@7.0.0

PR #874, opened 2017-02-03
fix #873 allow desctructuring for es6 modules imports

### LINKED ISSUE #873: webpack 2.x i18next destructure
We currently destructure i18next imports, but when upgrading to webpack 2.0 (es6 modules) the functions are no longer present.  It looks like there is only `export default i18next`.

Would it be possible to export the other functions at the same level for the es6 modules?
`export const t = i18next.t` or something similar?

bookman25:
export all functions off i18next to allow desctructuring i18next imports `import { t } from 'i18next';`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,19 @@
 import i18next from './i18next';
 
 export default i18next;
+
+export const changeLanguage = i18next.changeLanguage.bind(i18next);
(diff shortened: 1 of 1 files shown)


jamuhl:
Didn't expect this so fast...cool...will merge and publish this asap.

jamuhl:
published in i18next@7.0.0

PR #896, opened 2017-03-08
remove regex escape from format separators, fixes #895

### LINKED ISSUE #895: Special characters cannot be used as interpolation format separators
Strings containing regex special characters cannot be used as interpolation format separators, as they are [regex escaped](https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L25), even though the separator is [not used as a regex](https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L60-L68).

SpeedoDevo:


jamuhl:
oh a PR...even better....thank you again.

jamuhl:
published with i18next@7.1.2 - thank you again.

SpeedoDevo:
awesome - thanks

PR #904, opened 2017-04-06
Fix: iterating over object's own properties

mpirio:
Hello,

When the result of a translation is an object (_Object, Array, etc_) and the copy is done, the translation function loops over all object's properties, even those that might be inherited and could not not have a translation.

We should use `hasOwnProperty` to not translate these properties.


jamuhl:
will be merged for next version...thank you

jamuhl:
landet in i18next@7.2.1 ... thank you

Issue #906, opened 2017-04-12
Deep/inner/nested keys do not work if key coincides with a string function.


rally25rs:
If you have a translation json file that contains:

```
{
  "action": {
    "add": "Add",
    "search": "Search",
    "length": "Length"
  }
}
```

and you try to resolve "add", it works correctly:

```
i18next.t('action.add'); // "Add"
```

However, if you try to resolve "search" it actually returns the `string.search` function:

```
i18next.t('action.search'); // function search() { [native code] }
```

Same with "length":

```
i18next.t('action.length'); // 7
```

You can test that the returned value is actually the `string.search` function:

```
"".search === i18next.t('action.search'); // true
```

jamuhl:
Thanks for reporting...rather interesting...will have a look asap and fix this.

rally25rs:
So it seems that this does not happen if you use plain i18next. I think it is actually an issue with the XHR backend. I created an issue for them here: https://github.com/i18next/i18next-xhr-backend/issues/258

PR #909, opened 2017-04-13
Issue 906

rally25rs:
# Changes:

* Added unit tests to demonstrate an issue where properties of a string could be returned when resolving a deep key.
* Logic change: When recursing into object properties to resolve "deep" keys, do not try to use properties on strings.

# Resolves Issues:

* #906 
* https://github.com/i18next/i18next-xhr-backend/issues/258

-----

I stared at this for a while and originally though I could have `Translator.isValidLookup()` just return `false` if the resolved object was not a string, but this caused a lot of issues with the `returnObjects` options, since in some cases an object tree is OK to return.

The `utils.getLastOfPath()` function ended up being the simplest and least intrusive place to change the logic that I could find.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -14,16 +14,20 @@ function getLastOfPath(object, path, Empty) {
     return (key && key.indexOf('###') > -1) ? key.replace(/###/g, '.') : key;
   }
 
+  function canNotTraverseDeeper() {
+    return !object || typeof object === 'string';
+  }
+
   let stack = (typeof path !== 'string') ? [].concat(path) : path.split('.');
   while(stack.length > 1) {
-    if (!object) return {};
+    if (canNotTraverseDeeper()) return {};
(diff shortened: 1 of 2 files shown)


jamuhl:
following up here. agree this would also have been my suggestion - delievering a PR is even more awesome ;)

jamuhl:
thanks for taking the time for locating the issue and providing a fix:

published in i18next@7.2.2

jamuhl:
Ever having a project needing a translation management on top of i18next have a look at our translation as a service layer: http://locize.com

rally25rs:
@jamuhl it looks like the "built" js files `i18next.js` and `i18next.min.js` in the root directory aren't rebuilt in v7.2.2.

jamuhl:
i18next@7.2.3 should solve this

Issue #917, opened 2017-04-28
Recursive nesting at interpolation


DarkerTV:
Hi, I'm currently developing a little Node JS application and I were using your nice library when I noticed a potential problem in nesting. Currently if you do this
```js
i18next.t('test', {
  input: '$t(test)',
});
```
with this json file
```json
{
  "test": "{{input}}"
}
```
You get this error `RangeError: Maximum call stack size exceeded`. 

Obviously this is different in my application but I wanted to simplify the problem. Is there any way to solve this problem, like avoid to process nesting from interpolation.

jamuhl:
You interpolate a nesting construct into a value and don't want it to be executed? Sorry i don't get that use case?

I know there can be circular nestings - but only thing i could do there would be to break that. So better get an error.

DarkerTV:
As I said this is a very simplified case, let me explain better. Currently my code look like this
```js
i18next.t('result', {
  userInput: args,
  result,
});
```
```json
{
    "result": "Your input is {{userInput}} and the execution result is {{result}}"
}
```
My application code will be public so anyone will be able to look how my translations are done, and I'm afraid of this problem because if the userInput contains a nesting it get processed and some malicious user can try to use this.

A way to avoid this problem could be add an option to disable the nesting at interpolations at initialize or at TranslationFunction.

jamuhl:
Wouldn't it be saver to input your userinput...?

jamuhl:
but anyway...would be rather easy to implement: https://github.com/i18next/i18next/blob/master/src/Translator.js#L176

just add a `if (options.nest !== false)` in front and call `t('key', { nest: false });`

if you like you can provide a PR or i will add this for the next version (no big deal).

DarkerTV:
Yes, I only mean that it would be and useful option in some cases, if you don't think like me I can do something like you have in your code.
```js
const userInput = args.replace(/\$/g, '$$$$');
```
Just for some reason I don't feel right to modify the original user content.

DarkerTV:
I can send the PR in some hours :)

jamuhl:
awesome...will try to publish an updated version asap.

PR #920, opened 2017-04-28
Add nest option to translation function

DarkerTV:
This PR adds a new option while running translation function.
- nest: if nest === false, nesting at interpolations will be disabled in this translation.

Example:
```js
i18next.t('result', {
  userInput: '$t(result)',
  result: 'random value',
  nest: false,
});
```
```json
{
    "result": "Your input is {{userInput}} and the execution result is {{result}}"
}
```
The result of this will be `Your input is $t(result) and the execution result is random value` instead of a `RangeError: Maximum call stack size exceeded`.

Note: This option is disabled by default, you need to call it explicitly with false as parameter to disable nesting. Any other value than false will not enable this option.

---

Closes: #917

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -173,7 +173,9 @@ class Translator extends EventEmitter {
     res = this.interpolator.interpolate(res, data, this.language);
 
     // nesting
-    res = this.interpolator.nest(res, (...args) => { return this.translate.apply(this, args); }, options);
+    if (options.nest !== false) {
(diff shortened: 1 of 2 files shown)


jamuhl:
great...will merge and publish later today..when coming home...hopefully got time.

jamuhl:
published in i18next@8.1.0

jamuhl:
thank you for adding this PR

PR #933, opened 2017-06-08
Fix languageChanged not getting emitted on initial load.

Huulivoide:
The async support broke changeLanguage() function, so that languageChanged would not get emmitted, unless language was directly given to the function as parameter. Thus language detectors would not trigger the event.

### DIFF
--- a/src/i18next.js
+++ b/src/i18next.js
@@ -205,10 +205,10 @@ class I18n extends EventEmitter {
   }
 
   changeLanguage(lng, callback) {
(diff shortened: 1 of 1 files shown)


jamuhl:
Thank you for finding and providing this PR.

Could you please fix the code climate issue...not use param `lng` as that is defined in upper scope?

Done that i will merge and publish this fix asap.

jamuhl:
in i18next@8.4.1

PR #946, opened 2017-07-02
Fix interpolation of values containing prefix/suffix

### LINKED ISSUE #945: Interpolation breaks when prefix/suffix contained in value
Prefix and suffix in interpolated values confuse the interpolator.

Given the config:
```
{ interpolation: {
    prefix: '__',
    suffix: '__'
  },
  resources: { en: { translation: {
    key1: '__username__ is __how__.',
    key2: '__username__ logged in.'
  }}},
}
```

i18next generates the following output:
```
i18next.t('key1', { username: 'luke__skywalker', how: 'great'});
'i18next::interpolator: missed to pass in variable skywalker is  for interpolating luke__skywalker is __how__.'
'lukehow__.'  // expected 'luke__skywalker is great.'

i18next.t('key1', { username: 'luke__sky__walker', how: 'great'});
'i18next::interpolator: missed to pass in variable sky for interpolating luke__sky__walker is __how__.'
'lukewalker is great.' // expected 'luke__sky__walker is great.'

i18next.t('key2', { username: 'luke__sky__walker' });
'i18next::interpolator: missed to pass in variable sky for interpolating luke__sky__walker logged in.'
'lukewalker logged in' // expected 'luke__sky__walker logged in'

// this case is fine:
i18next.t('key2', { username: 'luke__skywalker' });
'luke__skywalker logged in.'

### PR DESCRIPTION
**Pull request by stiliyan**
This ensures replaced values are not considered for interpolation by advancing the matching index past the last replaced value. Resolves #945. 

~~Would love to add tests for this; let me know where the best place would be.~~ Found it! Added coverage [here](https://github.com/i18next/i18next/pull/946/files#diff-9a26d7d449fb14ddcfae0c36a1bd0db5R230).

One of the compatibility tests started to fail below. It looks out of date and perhaps could be replaced?

There's still [an issue](https://github.com/i18next/i18next/pull/946/files#diff-9a26d7d449fb14ddcfae0c36a1bd0db5R249) with text containing an unescaped variable followed by one or more standard variables. If a value containing the standard prefix is passed to the unescaped variable, the prefix will offset interpreter variable matching for the remaining unescaped variables. But this should be a much more rare case.

### DIFF
--- a/test/backward/v1.11.1.compat.js
+++ b/test/backward/v1.11.1.compat.js
@@ -2051,7 +2051,7 @@ describe('i18next', function() {
           expect(i18n.t('interpolationTest7', {toAdd: '<html>', escapeInterpolation: true})).to.be('added <html> &lt;html&gt;');
         });
 
-        it("should not accept interpolations from inside interpolations", function() {
+        it.skip("should not accept interpolations from inside interpolations", function() {
           expect(i18n.t('interpolationTest8', { toAdd1: '__toAdd2HTML__', toAdd2: '<html>'})).to.be('added  &lt;html&gt;');
         });
 
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -55,6 +55,8 @@ class Interpolator {
   interpolate(str, data, lng) {
     let match;
     let value;
+    let pattern;
+    let patternIndex;
 
     function regexSafe(val) {
       return val.replace(/\$/g, '$$$$');
@@ -76,8 +78,11 @@ class Interpolator {
     /* eslint no-cond-assign: 0 */
     while (match = this.regexpUnescape.exec(str)) {
       value = handleFormat(match[1].trim());
-      str = str.replace(match[0], value);
-      this.regexpUnescape.lastIndex = 0;
+      pattern = match[0];
+      patternIndex = str.indexOf(pattern);
+      str = str.replace(pattern, value);
+      // progress matching just past the replacement
+      this.regexpUnescape.lastIndex = patternIndex + value.length;
     }
 
     // regular escape on demand
@@ -89,8 +94,11 @@ class Interpolator {
         value = '';
       }
       value = this.escapeValue ? regexSafe(this.escape(value)) : regexSafe(value);
-      str = str.replace(match[0], value);
-      this.regexp.lastIndex = 0;
+      pattern = match[0];
+      patternIndex = str.indexOf(pattern);
+      str = str.replace(pattern, value);
+      // progress matching just past the replacement
+      this.regexp.lastIndex = patternIndex + value.length;
     }
     return str;
   }
--- a/test/interpolation.spec.js
+++ b/test/interpolation.spec.js
@@ -227,5 +227,33 @@ describe('Interpolator', () => {
     });
   });
 
+  describe('interpolate() - interpolation of values containing prefix and suffix', () => {
+    var ip;
 
+    before(() => {
+      ip = new Interpolator({
+        interpolation: {
+          prefix: '__',
+          suffix: '__',
+          unescapePrefix: ':'
+        }
+      });
+    });
+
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: test/backward/v1.11.1.compat.js [hunk 1]
```diff
@@ -2051,7 +2051,7 @@ describe('i18next', function() {
           expect(i18n.t('interpolationTest7', {toAdd: '<html>', escapeInterpolation: true})).to.be('added <html> &lt;html&gt;');
         });
 
-        it("should not accept interpolations from inside interpolations", function() {
+        it.skip("should not accept interpolations from inside interpolations", function() {
```
stiliyan: Fails with the condition below. What would be the best way to handle this?

```
default i18next way - with escaping interpolated arguments per default
  ✗ should not accept interpolations from inside interpolations
	expected 'added __toAdd2HTML__ &lt;html&gt;' to equal 'added  &lt;html&gt;'
```

jamuhl:
Hi @stiliyan really appreciate the effort you put into this PR.

Current behaviour starting from v2 was to accept interpolating replacements inside arguments:

`t('replace {{that}}', { that: 'with {{self}}', self: 'this }`) // -> replace with this``

Having just `{{` or in your case `__` in the value - how often does this really happen? We never had that case...what about changing interpolation prefix/suffix for that case?

stiliyan:
Hi @jamuhl. Thank you for reviewing. 

This is most likely to occur with user-supplied values (usernames can be one example) and is compounded by using the same prefix and suffix. Unfortunately changing suffix and prefix for text with user-provided values will not work in all cases.

I can modify the approach to accommodate for interpolating replacements. Can we consider constructing replacement patterns from the keys in the replacements object, rather than the translation string itself? This way we'll only replace values provided in the object further reducing potential to misinterpret which patterns to replace.

jamuhl:
replacement from the object passed in can be rather difficult...

a) can be huge objects
b) we support also nested lookup like `{{user.firstname}}`

savest is to assert you escape all values from user input to not include the prefix/suffix i think

stiliyan:
Thank you for pointing those additional cases! I'd like to better understand the complexity of use cases. Can you point me to a typical with deep look up and a complex object with replacement values? How many levels does interpolated replacement support? I'm guessing something like `t('replace {{that}}', { that: 'with {{self}}', self: '{{those}}', those: 'those' }); // -> 'replace with those'` is expected, and perhaps some handling for infinite recursion: `t('replace {{that}}' { that: '{{self}}', self: '{{that}}' });`.

Escaping prefix/suffix is a viable solution to a subset of cases, but it wont address the root cause of the interpolator misinterpreting look ups in a partially interpolated translation string. I wonder if the interpolator can tackle the issue efficiently in addition to supporting nested look ups and interpolating replacements.

Perhaps interpolator can pre-process the translation string to get a full list of look ups - nested, direct, and unescaped; it can also get a list of matching replacements from the replacement object. Look ups without matching replacements can be left unchanged in the final string. Interpolating replacements could be interpolated first before interpolating the full string, so all replacements resolve to direct values before interpolating the string.

Apologies if most of this has been discussed or is handled, I haven't had a chance to review in detail yet.

jamuhl:
Currently we can say the risk of running into this issue is very low - even lower when using the default prefix/suffix `{{`/`}}`. I'm not yet convinced it's worth adding more complexity to make that little risk bulletproof. The issue could be 99,9% be avoided by eg. setting prefix/suffix to something like `{#_{`/`_}#}` -> not very practical but i think it's not worth adding more complexity for a such low risk. If we would have had a call for that issue was hit in production ok - but yet we did never hear that in over 5 years of i18next.

But if you find a solution, that is elegant, has no negativ impact on performance and is backwards compatible...go for it.

jamuhl:
closing for now...feel free to add another PR solving the backwards compatibility issue

PR #949, opened 2017-07-06
getFixedT lng parameter of array type

mxl:
As mentioned in [documentation](https://www.i18next.com/api.html#getfixedt):

> Both params could be arrays of languages or namespaces and will be treated as fallbacks in that case.

Also I'm confused about "will be treated as fallbacks in that case" part because parameters of `getFixedT` are not used to fallback language or namespace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ node_modules
 node_modules/**/*
 coverage/**/*
 dist/**/*
+
+# Ignore IntelliJ based IDEs project files
+.idea
--- a/src/i18next.js
+++ b/src/i18next.js
@@ -242,10 +242,15 @@ class I18n extends EventEmitter {
     const fixedT = (key, opts = {}) => {
       const options = { ...opts };
       options.lng = options.lng || fixedT.lng;
+      options.lngs = options.lngs || fixedT.lngs;
       options.ns = options.ns || fixedT.ns;
       return this.t(key, options);
     };
-    fixedT.lng = lng;
+    if (typeof lng === 'string') {
+      fixedT.lng = lng;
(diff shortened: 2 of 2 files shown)


jamuhl:
Sorry what does this PR fix. Please do not reformat the code...just fix relevant code.

jamuhl:
regarding the question...you can call t function with array of lngs, ns -> t function will use those beside fallbackLng, fallbackNS -> that's why it acts as fallback if passing array -> first is primar target all other array items act as additional fallbacks if not found on primary.

mxl:
@jamuhl I did not reformat. I just ran `npm run-script build` and included changes in `i18next.js` and `i18next.min.js` in commit. Should I do this or not?

mxl:
This PR fixes using `lng` as array in `getFixedT`. 
`t` has [`lngs` option](https://www.i18next.com/essentials.html#overview-options) which is used to pass array but `getFixedT` uses only `t` function `lng` option that can be only a string. Am I right?

jamuhl:
ah sorry now i see...would be easier if not pushing build in PR as i will build that on publish...

the change in languageUtils i would prefer not to have...

will copy the changes on i18next.getFixedT and publish that...or if you prefer merge a cleaned PR

mxl:
Cleaned up PR.

mxl:
Fixed quotes and semicolons.

jamuhl:
awesome...will merge this later today and publish an update to npm. thanks a lot for your contribution...

jamuhl:
published in i18next@8.4.3

thanks again for providing this. and also for improvements on documentation.

PR #953, opened 2017-07-30
Missing bindings to this when using ES6 destructuring import

ZitRos:
Hello! Thanks for the nice library. I use it in my project for a long time, but today I found one issue I think we need to fix.

Using NodeJS v8.0.0 with babel-core@6.25.0, this:

```javascript
import { changeLanguage } from "i18next";

// ...

changeLanguage("en");
```

Produces the next error:

```bash
TypeError: Cannot set property 'language' of undefined
    at setLng (Z:\ZitRos\Projects\TimingKit\node_modules\i18next\dist\commonjs\i18next.js:315:25)
    at changeLanguage (Z:\ZitRos\Projects\TimingKit\node_modules\i18next\dist\commonjs\i18next.js:333:7)
    at Client.renderDashboard (Z:\ZitRos\Projects\TimingKit\build_base\server\...)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:169:7)
```

Whether this:

```javascript
import i18next from "i18next";

// ...

i18next.changeLanguage("en");
```

Does not.

I think it is expected that the imported function must work properly, because there is always one instance of i18next exported (default export).

The reason of the error is in [i18next/index.js](https://github.com/i18next/i18next/blob/master/index.js) file. There is only `default` property exported from `dist/commonjs/index.js` instead of all the properties. Requiring `i18next/index.js` gives the **i18next plain instance** (object), hence this results in i18next instance properties cast to a set of exports. This way, `changeLanguage`, being called as a plain exported function, has **this** equals to **undefined**.

This pull request fixes the problem.

Furthermore, I think there is no need in `index.js` file, we can just specify `"main"` property of `package.json` file to be equal to `"dist/commonjs/index.js"`. What do you think?

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,1 +1,1 @@
-module.exports = require('./dist/commonjs/index.js').default;
+module.exports = require('./dist/commonjs/index.js');


jamuhl:
The point is the main export for node.js is pure commonjs support...so people can use it in node.js without problems.

If you use babel-core to support all es2015 features imports/exports you are expected to import the es build from /dist/es

both jsnext and module entry points are set in package.json https://github.com/i18next/i18next/blob/master/package.json#L6

jamuhl:
if you call changeLanguage like that it is not bound to the context. You would have to bind it to i18next.

basically those es exports are not meant to be used in the nodejs env.

ZitRos:
> if you call changeLanguage like that it is not bound to the context. You would have to bind it to i18next.

Should I? Okay. Let's see what's in [dist/commonjs/index.js file](https://unpkg.com/i18next@8.4.3/dist/commonjs/index.js), from where CommonJS `i18next/index.js` picks **default** export. You will find this:

```javascript
exports.default = _i18next2.default;
var changeLanguage = exports.changeLanguage = _i18next2.default.changeLanguage.bind(_i18next2.default);
var cloneInstance = exports.cloneInstance = _i18next2.default.cloneInstance.bind(_i18next2.default);
var createInstance = exports.createInstance = _i18next2.default.createInstance.bind(_i18next2.default);
var dir = exports.dir = _i18next2.default.dir.bind(_i18next2.default);
var exists = exports.exists = _i18next2.default.exists.bind(_i18next2.default);
var getFixedT = exports.getFixedT = _i18next2.default.getFixedT.bind(_i18next2.default);
var init = exports.init = _i18next2.default.init.bind(_i18next2.default);
var loadLanguages = exports.loadLanguages = _i18next2.default.loadLanguages.bind(_i18next2.default);
var loadNamespaces = exports.loadNamespaces = _i18next2.default.loadNamespaces.bind(_i18next2.default);
var loadResources = exports.loadResources = _i18next2.default.loadResources.bind(_i18next2.default);
var off = exports.off = _i18next2.default.off.bind(_i18next2.default);
var on = exports.on = _i18next2.default.on.bind(_i18next2.default);
var setDefaultNamespace = exports.setDefaultNamespace = _i18next2.default.setDefaultNamespace.bind(_i18next2.default);
var t = exports.t = _i18next2.default.t.bind(_i18next2.default);
var use = exports.use = _i18next2.default.use.bind(_i18next2.default);
```

All the function **are already bound to the context**, they **are just never used** because of the reason I explained in this PR: CommonJS module exports **only the _default_ property**, which is **the object**, and which properties become de-facto exported functions.

My mission here is to explain that above exports are used in the way they are not designed to be used, and it worked all the time before properly just because NodeJS modules exports and JavaScript are so flexible and no one cares. See — I have just deleted the `.default` from `i18next/index.js`, no tests are failing, ES6 module destructuring (as well as [my example with require](https://github.com/i18next/i18next/pull/953#issuecomment-319149028)) magically begins to work, no context bindings are required (as they are already done).

Do you agree?

jamuhl:
sure i agree.

but if you have an env without require of babel after your change:

```
var i18next = require('i18next');

i18next.init(); // -> ok as exported
i18next.options; // -> nok as i18next is not the instance
```

so we would have a breaking change in this case.

ZitRos:
@jamuhl, thank you very much for clarifications.

> so we would have a breaking change in this case.

I agree, then i18next is not the instance anymore but module. Right, this may break people's code...

I wish NodeJS supports `"module"` property in `package.json` one day :) Until then, to follow ES-style I need to use exports from `i18next/dist/es/index.js`.

Thank you! Feel free to close this PR.

jamuhl:
yes totally agree...hopefully proper support will land soon. Current situation is not optimal.

Issue #957, opened 2017-08-23
I18next stops working at localize


CoderStefan:
Hello,

i have the following key:
  resources: {
    en: {
      translation: {
        "key": "hello world {{name}}"
      }
    },
    de: {
      translation: {
        "key": "hello welt {{name}}"
      }
    }
  }

when i use i18next with this:
  document.getElementById('output').innerHTML = i18next.t('key', { name: 'i18next '}); 
then its works

but i use this:
document.getElementById('output').innerHTML = i18next.t('key', { name: 'i18next {{name}}'});
then i18next stops working

demo:
https://jsfiddle.net/6x1897te/

jamuhl:
this causes an endless loop

name: 'i18next {{name}}' -> interpolates name again which interpolates name again and again and again -> endless loop

CoderStefan:
Is there a setting or mechanism to interrupt the endless loop?

jamuhl:
beside fixing it in the translation?

`i18next.t('key', { name: 'i18next {{otherName}}', otherName: 'theOtherName' })`

CoderStefan:
My problem is also that in some cases unfortunately also a user input.

jamuhl:
you expect users to input a {{yourVariable}} hm...might happen....

A) replace those {{ }} in user input
B) choose another prefix/suffix users never will guess
C) provide a PR:

https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L77
https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L84

setting a counter and exit after eg. 1000 replaces ;) -> but still if handling it graceful in code the UI will run crazy ;)

jamuhl:
thank you for the awesome PR

PR #958, opened 2017-08-24
max replaces in interpolate to prevent endless loop

CoderStefan:
#957 add setting maxReplaces in interpolation options to prevent endless loop in interpolation


jamuhl:
will be merged for next version

jamuhl:
great PR (changelog, tests, all done)

published in i18next@9.0.0

PR #980, opened 2017-10-29
feat: native browser es modules via adding .js where needed

daKmoR:
hey,
currently tested only in browser.


jamuhl:
awesome...will merge and publish this later today...thanks for providing and testing on chrome

jamuhl:
merged and pubished in i18next@10.0.6

thanks again for doing the PR

PR #987, opened 2017-11-04
feat: emit fullQueueLoaded event

daKmoR:
I'm loading multiple resources in different webcomponents and my app needs to know when everything is loaded.

if you need any more details just let me know :)


daKmoR:
Also would it be possible to give it all that has been loaded?

or is there any other easy way to get all the loaded data like in a big object or so?

daKmoR:
nevermind I solved it differently :p 
sorry for the "scare" :p

PR #996, opened 2017-11-20
Return raw boolean values with returnObjects=true

MartynasSapoka:
When you have raw boolean values in locale files and call `translate('...', {returnObjects: true})`, it returns the value `{}` instead of `true`. Added a simple check to avoid processing boolean values further.


jamuhl:
will check, merge and publish this later....

thank you for taking the time doing this PR - very welcome addition

jamuhl:
thanks again

published in i18next@10.1.0

beside that added numbers beside booleans too, just in case ;)

MartynasSapoka:
My pleasure! Thanks for publishing so quickly :)

Issue #1009, opened 2017-12-28
Context and nested lost initial options


fram03:
Hi, I'm updating i18next version in my project from a prehistoric 1.7 to the last 10.2.1.
The only problem encountered is about context and nest.
If I specify a context in a resource value initial options will be lost in the nested translations.
Is there a param to force this behaviour?
```javascript
var i18n = require("i18next")
i18n.init({
    'lng': 'en',
    'compatibilityJSON': 'v1',
    'resources': {
        'en': {
            'translation': {
                "withContext_A": "hello world from __name__ - A",
                "withContext": "hello world from __name__ - B",
                "start": "Start string $t(withContext, {'context': '__test__'})",
                "startClean": "Start string $t(withContext)"
            }
        }
    },
    'interpolation': {
        'escapeValue': false,
        'prefix': '__',
        'suffix': '__'
    }
}, function() {
    // Expected "Start string hello world from John - A"  (as in 1.7)
    // but I'm receiving "Start string hello world from - A"
    console.log('TEST1', i18n.t('start', {'test': 'A', 'name': 'John'}));  
   // Expected "Start string hello world from John - B" and it's ok
    console.log('TEST2', i18n.t('startClean', {'test': 'A', 'name': 'John'}));
});
```

jamuhl:
when doing nest and you pass in options like `"Start string $t(withContext, {'context': '__test__'})"` it only takes this options in the string not the one passed in, cause of:

https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L129

a) change it to: `"Start string $t(withContext, {'context': '__test__', 'name': '__name__'})"`
b) provide a PR cloning those options back into the clonedOptions

jamuhl:
published in i18next@10.2.2

PR #1010, opened 2017-12-29
https://github.com/i18next/i18next/issues/1009

fram03:
nested translations inherit parent's options


jamuhl:
great...thanks a lot...will try to merge this later and bring out a new version...

jamuhl:
published in i18next@10.2.2

thanks for providing this

PR #1019, opened 2018-02-17
using load instead of reload to reload resources. This way we get a callback when there was nothing before

antialias:


antialias:
looking into this a little more, if `load` is used as I have it here, then fresher versions of the resources will not be loaded, since i18next prefers to cache. I believe that a good solution involves a `force fetch` flag that `queueLoad` can observe. The argument interface for backend connector does not allow for much flexibility, so I imagine some discussion will have to take place about how to do this.

jamuhl:
differences:

load checks if already loaded by https://github.com/i18next/i18next/blob/master/src/BackendConnector.js#L141

callbacks are called by https://github.com/i18next/i18next/blob/master/src/BackendConnector.js#L107

looks rather complex - but basically it asserts multiple calls to init or addNamespace,... get there callback called immediately if those resources are available (and does not load again if not needed)

out of that reload currently would get it's callback called immediately as resources are there. We do also not want to remove them as the app still needs to translate.

But i think you already got that. So yes queueLoad would need some additional flag to force fetch...

jamuhl:
do you have any plan for another approach on this...or should this PR be closed

antialias:
Your comments are good guidance and I’d like to give it another try, but
not sure when I’ll get around to it. One way or another I really think this
method deserves a callback.

jamuhl:
FYI: v11.0.0 reduced complexity in the backendConnector...removed that multiRead stuff

jamuhl:
closing for now....as outdated

Issue #1022, opened 2018-02-25
[Feature Request] pass options to addResources and addResourceBundle


olivierpascal:
Sometimes you want to load resource bundles in background without triggering events and re-rendering the whole app (ie. with react-i18next).

`addResource()` has a handy option (`options.silent`) which disable the event emitting. Is it possible to have the same with `addResources()` and `addResourceBundle`?

jamuhl:
Would you mind creating a pullrequest?

olivierpascal:
Sure!

PR #1024, opened 2018-02-26
Added options.silent to addResources and addResourceBundle

olivierpascal:
Issue #1022 

Sometimes you want to load resource bundles in background without triggering events and re-rendering the whole app (ie. with react-i18next).

Added an optional parameter to addResources() and addResourceBundle() (`option = { silent: false}`), similar to addResource(), to ignore the emitting of 'added' event.


jamuhl:
thank you for providing this...pubished in i18next@10.5.0

PR #1037, opened 2018-03-15
Avoid mutating whitelist array. Fixes #1036

### LINKED ISSUE #1036: Using the same array on `whitelist` and `preload` causes `cimode` to be preloaded
Given the following config:
```js
const locales = ["nl", "fr"]
const options = {
  whitelist: locales,
  preload: locales,
  ...
}
```

On the server (Next.js) this leads to `Error: ENOENT: no such file or directory, open '.../locales/cimode/common.json'`, and in the browser it attempts to load these `cimode` language files through AJAX leading to 404 errors.

Turns out this error is caused by the `whitelist` array being [mutated in-place](https://github.com/i18next/i18next/blob/15d7e18e44640fc0a596e11a65f7e675602d2cae/src/defaults.js#L73). The workaround is obviously to avoid passing the same array to both properties, but to me this looks like an API bug.

ghengeveld:
This replaces Array.push with Array.concat in order to avoid mutation. This prevents the issue described in #1036.

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -71,7 +71,7 @@ export function transformOptions(options) {
 
   // extend whitelist with cimode
   if (options.whitelist && options.whitelist.indexOf('cimode') < 0) {
-    options.whitelist = options.whitelist.concat(['cimode'])
+    options.whitelist = options.whitelist.concat(['cimode']);
   }
 
   return options;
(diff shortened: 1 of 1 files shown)


jamuhl:
awesome...will merge and publish this tomorrow.

ghengeveld:
Excellent 👍

jamuhl:
published in i18next@10.5.1

ghengeveld:
🚀 Thanks

PR #1039, opened 2018-03-20
Adds missingInterpolationHandler. Fixes #1033 

### LINKED ISSUE #1033: missing interpolation callback
I wonder if there is any possibility to register a callback for missing interpolation values. I'm thinking about a construct similar to the missingKeyHandler. Is there any way to do this? If not, where to add it?

I need to extract the missing placeholder (not the dynamic value) of a translation somehow.

e.g.

If i use

`i18next.t('key', { what: 'i18next' });`

with a translation like

`{
    "key": "{{what}} is {{how}}"
`}

I'd like to get information about the missing `{{how}}`

Thank you in advance!

JanTrotnow:


jamuhl:
awesome...will try to merge and publish an update asap

jamuhl:
published in i18next@10.6.0

PR #1046, opened 2018-04-19
Remove unnecessary warning when value is empty string

manixate:
This fix is same as done for i18next/react-i18next#242


jamuhl:
Could you fix the now broken test so '' will not log missing...

thanks a lot, Jan

jamuhl:
published in i18next@11.2.3

manixate:
Great. Thanks

Issue #1007, opened 2017-12-15
Disable internolation by i18n.t(key, options)


lomaster1:
I know my localication string has not dynamic values, but when I use i18n.t('somekey') interpolation check the string.

I want to disable interpolation via options for performance reason, because interpolation takes more than half the time (milliseconds but if string many?). 

For example, i18n.t('somekey', { disableInterpolation: true }).

jamuhl:
https://github.com/i18next/i18next/blob/master/src/Translator.js#L171

PR will be accepted

jamuhl:
https://github.com/i18next/i18next/pull/1050

published in i18next@11.3.0

PR #1050, opened 2018-05-08
Allow translate function to skip interpolation and return raw values instead

manixate:
This PR allows skipping interpolation but still allows post processing of the returned values. Let me know if you want post processing to be skipped as well. I allowed it because post processing already has `options.applyPostProcessor` to enable or disable it.

Fixes: #1007


jamuhl:
@manixate thanks for providing this PR. Agree with you excluding postProcessing can be done by using the other option. Will merge and publish this now.

jamuhl:
published in i18next@11.3.0

thanks again for providing this complete PR

manixate:
I agree on:

> skipInterpolation is not applied if i18nFormat is set -> different parser (eg. ICU)

For my use-case it is perfectly fine but the reason I skipped the whole i18nFormat was because the linked issue mentioned performance reasons and I thought skipping would be beneficial.

jamuhl:
the issue was created before having those additional i18nFormats...so not affected there when using those format plugins...

PR #1051, opened 2018-05-08
Allow falsy keySeparator in i18next.t

### LINKED ISSUE #1021: [Feature Request] Passing falsy `keySeparator` to i18next.t to disable the keySeparator option in specific call
reproduction code with i18next 10.4.1:
```js
i18next.init({
  lng: 'en',
  defaultNS: 'main',
  resources: {
    'en': {
      main: {
        foo: {
          foo: 'bar',
        },
        'bar.baz': 'hello',
      },
    },
  },
}, () => {
  console.log(i18next.t('foo.foo')) // bar
  console.log(i18next.t('bar.baz', { keySeparator: false })) // bar.baz
  console.log(i18next.t('bar.baz', { keySeparator: true }))  // hello
})
```

In our projects we might have dots in keys for some namespaces / components, although escaping both resources and keys is a possible solution, passing `{ keySeparator: false}` to specific i18next.t usage seems more natural

manixate:
Closes #1021

### DIFF
--- a/src/ResourceStore.js
+++ b/src/ResourceStore.js
@@ -6,6 +6,9 @@ class ResourceStore extends EventEmitter {
     super();
     this.data = data || {};
     this.options = options;
+    if (this.options.keySeparator === undefined) {
+      this.options.keySeparator = '.';
+    }
   }
 
   addNamespaces(ns) {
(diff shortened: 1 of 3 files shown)


jamuhl:
published in i18next@11.3.1

PR #1052, opened 2018-05-14
Properly handle arguments containing arrays while listening to a wildcard event

manixate:
Currently, it spread all the arguments because of how `concat` works but this is fixed in this PR.

e.g if you pass `emitter.emit('event', ['en', 'pt'], 'key')` it was resulting in args as `'en', 'pt', 'key'`. This PR fixes the arguments to result in `['en', 'pt'], 'key'` while listening to wildcard. Previous behaviour breaks listening to 'missingKey' event.


jamuhl:
published in i18next@11.3.2

PR #1066, opened 2018-06-22
Allow loading remote resources when also using local ones

dyniper:
This change allow simpler bundling of one language and getting other resources remotely. If you think this change is good, i can write unit test for it, but wanted a feel first before i invest more in this.

### DIFF
--- a/src/i18next.js
+++ b/src/i18next.js
@@ -116,7 +116,8 @@ class I18n extends EventEmitter {
 
   /* eslint consistent-return: 0 */
   loadResources(callback = noop) {
-    if (!this.options.resources) {
+    // Try to load from the backend if either no resource were provided, or if the current language resource was not
+    if (!this.options.resources || !this.options.resources[this.language]) {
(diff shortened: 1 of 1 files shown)


jamuhl:
Problem with it - it's breaking and will break consumers production by start fetching data when not expected.

setting resources on init is there and documented as not using any backend implementation, like setting lng on init is omitting detection: -> basically be like all other i18n frameworks

If you like to achieve partial bundling use either the i18next.addResourceBundle to add those initial bundled.

jamuhl:
could you make this change as an optional feature, like eg.:

`if (!this.options.resources || (this.options.partialBundledLanguages && !this.options.resources[this.languages[0]]))`

and use this.languages[0] -> not the detected but the first to load that is whitelisted

dyniper:
That's a good idea. Let me rework this in that vein. The problem with addResourceBundle is that will still try to remotely load the resource on init.

jamuhl:
Feel free to doing another pull request with the changes asked - but i guess due inactivity there is no need anylonger.

closing this for now.

Issue #1071, opened 2018-07-09
Translator: No plural when saving missing with count equal to 0


Chcap:
Hello,

when using a `BackendConnector` and having a missing translation with a `count` attribute equal to zero, the `BackendConnector.create` is called only with the `key` case, and not its plural forms.
In consequence, if the count value is random and by (bad) luck the first random value is `0`, then, the plural values are not created.

The solution is in the [Translator](https://github.com/i18next/i18next/blob/master/src/Translator.js#L158)  where the case when `count` is zero should not be counted as falsy.

jamuhl:
good catch...will be fixed

Chcap:
I opened the PR https://github.com/i18next/i18next/pull/1072 to fix it :)

PR #1072, opened 2018-07-09
Authorize zero for saveMissingPlurals

Chcap:
To solve https://github.com/i18next/i18next/issues/1071


jamuhl:
published in i18next@11.3.6

Issue #1058, opened 2018-05-28
Allow to override `interpolation.escape` function in `t` function


eugef:
Right now it is only possible to set custom `escape` function for interpolation during the init, it is not possible to override while executing `t`.

Example:
```
t('key', {interpolation: {escape: customFunc}})
```

Expected behaviour: `customFunc` is executed

Actual behaviour: `utils.escape` is still being executed

The reason is that here https://github.com/i18next/i18next/blob/9871516aa0fb0e5b12953bd6ba192a4f7cf2c980/src/Interpolator.js#L16 it needs a `reset` argument to change the `escape` function.

Is it a conscious decision not to allow override `escape` function while executing `t`?

if not - I can write a PR that will enable it.

jamuhl:
Good question...i guess we expect the escape function to be consistent over usage. What is your use case to change the escape function for certain t calls?

eugef:
I am using i18next library in Node.js. The same instance is used in 2 different contexts: 1st is directly invoking in js code, and 2nd is using via helper in from templates (I am using Handlebars).

I need a way to pass html into translations from the templates (and use Handlebars escape function instead of default one) but not from the js code.

That is why i `t` function that is invoked from the template i Want to set custom `interpolation.escape` function.

jamuhl:
ah ok...i see...guess in that case it is ok to change that so we set the escape function passed during t call and reset it afterwards like for the other values.

If you like to create a PR that would help to speed things up...else i will see if i find time next week to add this...

thanks for the additional informations to understand the need better

eugef:
There is one more thing to consider - if custom `escape` function is being used I'd add a configuration to disable conversion interpolated value into string, see https://github.com/i18next/i18next/blob/9871516aa0fb0e5b12953bd6ba192a4f7cf2c980/src/Interpolator.js#L104

The reason is that custom escape function can handle it itself, for example Handlebars escape function can handle not only simple strings, but also Handlebars.SafeString

I propose following config:
```
interpolation: {
  escape: customFunc,
  escapeRawValue: true // false by default
}
```

If `escapeRawValue` is `true`, then `value = utils.makeString(value);` is not executed

What is your opinion on this?

eugef:
Great I will try to make a PR within few weeks

jamuhl:
awesome...really appreciated

PR #1076, opened 2018-07-20
Allow to pass raw value to the custom interpolation escape function

eugef:
If custom escape function is used, then raw value (not type casted to String) can be passed to it.

This PR solves issue #1058 

New Interpolation options is added:

| option | default | description
-- | -- | --
useRawValueToEscape | false | If true, then value passed into escape function is not casted to string, use with custom escape function that does its own type check)

My example use-case: `Handlebars.Utils.escapeExpression` is used as escape function and `useRawValueToEscape` is true then escaping can be disabled for a specific value as soon as it is wrapped in Handlebars.SafeString.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -13,13 +13,14 @@ class Interpolator {
     if (reset) {
       this.options = options;
       this.format = (options.interpolation && options.interpolation.format) || (value => value);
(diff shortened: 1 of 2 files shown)


jamuhl:
published in i18next@11.4.0

thank you for setting this PR up

eugef:
Thank you for quick release!

PR #1079, opened 2018-07-23
EventEmitter.prototype.on returns this

### LINKED ISSUE #1077: Have the `EventEmitter.prototype.on` to return `this`
It would be great if the `EventEmitter.prototype.on` returned `this` so we can subscribe to events in the configuration chain.

```js
i18n.createInstance()
  .use(ChainedBackend)
  .use(LanguageDetector)
  .use(reactI18nextModule)
  .on('initialized', () => console.log('initialized'))
  .init(i18nOptions); // <- fails
```

With the current implementation you get `Uncaught TypeError: Cannot read property 'init' of undefined` because it returns `undefined`. But with an updated version, this would work.

```js
class EventEmitter {
  on(events, listener) {
    events.split(' ').forEach((event) => {
      this.observers[event] = this.observers[event] || [];
      this.observers[event].push(listener);
    });
    return this; // <- returns this
  }
}
```
https://github.com/i18next/i18next/blob/d6597d7653751cb3ea79a0f2b98fad4c23603adb/src/EventEmitter.js#L6

I’d be happy to submit a PR if you guys think it’s a good idea.

marlun78:
Having `EventEmitter.prototype.on` return`this` enables subscribing to events as part of the config chain. 
```js
i18n.createInstance()
  .use(ChainedBackend)
  .use(LanguageDetector)
  .use(reactI18nextModule)
  .on('initialized', () => console.log('initialized'))
  .init(i18nOptions);
```
This commit closes #1077.

### DIFF
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -8,6 +8,7 @@ class EventEmitter {
       this.observers[event] = this.observers[event] || [];
       this.observers[event].push(listener);
     });
+    return this;
   }
 
   off(event, listener) {
--- a/test/eventEmitter.spec.js
+++ b/test/eventEmitter.spec.js
@@ -3,7 +3,7 @@ import EventEmitter from '../src/EventEmitter';
 describe('i18next', () => {
 
   describe('published', () => {
-    
+
     let emitter;
     beforeEach(() => {
       emitter = new EventEmitter();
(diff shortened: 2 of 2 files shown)


marlun78:
Cool, thanks!

PR #1087, opened 2018-08-08
Add getDataByLanguage method to return language data currently in store #1086

### LINKED ISSUE #1086: getLanguageBundle
Hi - I'm wondering if it makes sense to add a store method similar to `getResourceBundle` that returns all namespaces loaded for that language? i.e. `getLanguageBundle(lng)`

My use case is for server rendering. I am loading only the namespaces needed into each cloned `i18n` instance. Then I serialize and send down the whole language object via HTML to hydrate `initialI18nStore` (react-i18next). Currently it's necessary to iteratively call `getResourceBundle` for each ns to build this object - it would be helpful if there was a method to return it instantly.

chacestew:
Simple method to return the whole language (i.e. namespaces loaded so far) from the store. Mostly so we don't have to access `resourceStore` object directly.  Closes #1086.

### DIFF
--- a/src/ResourceStore.js
+++ b/src/ResourceStore.js
@@ -112,6 +112,10 @@ class ResourceStore extends EventEmitter {
     return this.getResource(lng, ns);
   }
 
+  getDataByLanguage(lng) {
+    return this.data[lng]
+  }
+
   toJSON() {
     return this.data;
   }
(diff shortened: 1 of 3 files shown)


jamuhl:
thank you - was published with i18next@11.6.0

PR #1097, opened 2018-09-05
added default value for plural version (applicable in English and oth…

gunzino:
my IDE claims that defaultValue_plural is missing but I think code should be fine. Correct me if I'm wrong.

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -119,9 +119,14 @@ class Translator extends EventEmitter {
       let usedKey = false;
 
       // fallback value
-      if (!this.isValidLookup(res) && options.defaultValue !== undefined) {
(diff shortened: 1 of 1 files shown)


jamuhl:
should be ok...but sadly it does not work like this...

not every language uses the pluralform for 0....really the only way is to get the correct suffix and decide on that value if we need to take plural default...

let this PR open...i might get to it later today with catching all the pitfalls.

jamuhl:
https://github.com/i18next/i18next/commit/94173ea7000b48b9cdd145cf8685a074f98760b5

jamuhl:
the functionality was published in 18next@11.7.0

PR #1118, opened 2018-10-18
Allow missingInterpolationHandler to be provided as t() option

### LINKED ISSUE #1117: Specify missingInterpolationHandler via t(_, options)
I've got a case where I'd like to detect missing interpolations for specific calls to `t()`. The newly added `missingInterpolationHandler` option would've have been perfect for this if it had been possible to override via the `t()` `options` argument, but unfortunately this is currently not the case.

From what I can see [this block](https://github.com/i18next/i18next/blob/0de7e5683565bcda8654aed5e9a59bf1ba916eaa/src/Interpolator.js#L97-L98) would need to be modified to first check if `missingInterpolationHandler` has been provided in the data object.

Thanks to all contributors for a great library by the way!

mogelbrod:
Fixes #1117.

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -197,7 +197,7 @@ class Translator extends EventEmitter {
       // interpolate
       let data = options.replace && typeof options.replace !== 'string' ? options.replace : options;
       if (this.options.interpolation.defaultVariables) data = { ...this.options.interpolation.defaultVariables, ...data };
-      res = this.interpolator.interpolate(res, data, options.lng || this.language);
+      res = this.interpolator.interpolate(res, data, options.lng || this.language, options);
 
       // nesting
       if (options.nest !== false) res = this.interpolator.nest(res, (...args) => this.translate(...args), options);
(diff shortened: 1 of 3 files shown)


src/Translator.js
```diff
@@ -197,7 +197,7 @@ class Translator extends EventEmitter {
       // interpolate
       let data = options.replace && typeof options.replace !== 'string' ? options.replace : options;
       if (this.options.interpolation.defaultVariables) data = { ...this.options.interpolation.defaultVariables, ...data };
-      res = this.interpolator.interpolate(res, data, options.lng || this.language);
+      res = this.interpolator.interpolate(res, data, options.lng || this.language, options);
```
mogelbrod: Not entirely satisfied with the additional argument here, but I also don't wish to risk breaking backwards compatibility.

jamuhl:
already late here....will merge this tomorrow and publish an updated version to npm

jamuhl:
Thank you for taking the time to provide this PR. Was merged and published to npm in i18next@11.10.0.

And don’t forget to star this repo. Make a tweet or have a look at our https://locize.com -> there are many ways to help this project :pray:

mogelbrod:
Thanks for the quick responses, much appreciated! ⭐️

Issue #661, opened 2016-04-20
Promise based API


perrin4869:
Are there any plans to support a promise based API, such as:

``` javascript
i18next.init(...)
.then(i18next => {
    ...
});
```

I guess this would necessitate a Promise polyfill, but since many popular APIs such as `fetch` and `Service Workers` depend on promises, I think it's safe to add these days.

jamuhl:
not planned...the way i would go is to pass in a Promise Implementation on init if want support for them.

But right now there is no plan for it...

perrin4869:
Got it!

isaachinman:
@jamuhl I'll comment here first instead of opening a new issue. In my opinion, it's pretty vital that the entire API be promisified. In building the NextJs example, I actually became confused as I did not expect callback-based methods.

Are you open to this? It's been awhile since I promisified a library, but as far as I remember, there are two options:

1. Simple option - wrap everything in some browser-friendly `util.promisify` polyfill.
2. Proper option - do an actual rewrite to return promises.

In general, you probably wouldn't want to eliminate callbacks before a major version bump, so if we want to do a minor release that adds promises, we'll need to maintain callbacks as well.

Let me know your thoughts.

jamuhl:
@isaachinman we could add promises for sure...the time seems now to be ready with the async-await options.

Personally, I would prefer a solution where we return a promise from the functions accepting a callback as function argument -> what I would avoid is removing the callback option completely. (So, I wouldn't even remove them on a major version - callbacks are nice 🍝)

isaachinman:
@jamuhl What do you think about [this pattern](https://codesandbox.io/s/vyjjrw4q10)? Just my first idea. Important to note that `callback` is going to be `undefined`. You probably already protect against that in the current implementation though.

jamuhl:
@isaachinman perfect...callback can be undefined...as it's optional in all cases.

isaachinman:
So it should be a one-liner fix for all methods. Do you want to take this one, or shall I? I don't have access to this repo to be able to reopen issues, but we can either reopen here or just create a new issue to track this.

jamuhl:
Will try to look into it tomorrow...should have the time for it...

isaachinman:
Okay. If you don't have time just let me know and I'll take care of it.

jamuhl:
closing this again - will be in next major: https://github.com/i18next/i18next/issues/1129

PR #1130, opened 2018-11-09
Adds promises API

jamuhl:
solves: https://github.com/i18next/i18next/issues/661

### DIFF
--- a/src/i18next.js
+++ b/src/i18next.js
@@ -8,6 +8,7 @@ import Interpolator from './Interpolator.js';
 import BackendConnector from './BackendConnector.js';
 import { get as getDefaults, transformOptions } from './defaults.js';
 import postProcessor from './postProcessor.js';
+import { defer } from './utils';
 
 function noop() {}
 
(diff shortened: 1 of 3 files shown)


jamuhl:
@isaachinman would be great if you take a quick look

isaachinman:
I don't really understand the use of `utils.defer`. Why not just use a `Promise` wherever you need it? Also, I think you will need to explicitly return a promise from these functions to be able to successfully call `await` or `then`?

jamuhl:
A Promise needs the resolving inside the passed function:

`new Promise(function(resolve, reject) { ... });`

but i need to call `resolve` from outside -> that is what the defer() enables:

it returns a Promise and enables calling resolve on it form outside.

https://github.com/i18next/i18next/pull/1130/files#diff-c573044ba770c7e0efcf82de26be326aR121 -> is a Promise

which get resolved here: https://github.com/i18next/i18next/pull/1130/files#diff-c573044ba770c7e0efcf82de26be326aR208

based on: http://lea.verou.me/2016/12/resolve-promises-externally-with-this-one-weird-trick/

isaachinman:
Ah, I see now what you're saying. It's an OK pattern, but a little harder to follow. Haven't seen that before. Maybe we can add a link to that blog post as a comment inside the `utils.defer` function, so that people looking through the source can understand it quickly?

In general this looks good though.

As far as the test cases, it might be better/simpler to test like:

    typeof i18next.init().then === 'function'

jamuhl:
@isaachinman Promises API was published in i18next@13.0.0

isaachinman:
@jamuhl That is great news! We will start consuming it immediately at `next-i18next`.

Issue #1129, opened 2018-11-09
V13: Roadmap


jamuhl:
- [x] Typescript definitions
- [x] Promise support

Will be breaking as returning Promises eg. on i18next.init() will replace the previous return value (this) for chaining calls.

jamuhl:
Promises: https://github.com/i18next/i18next/pull/1130

jamuhl:
@schettino - do you have time to add typescript here in the next few weeks?

schettino:
I'm going to start right away

jamuhl:
@schettino awesome...but no need to rush...take the time needed.

schettino:
Checking this out. I'll review it in order to get this out asap. I'm looking forward writing the typedefs for the hooks. @jamuhl how's the roadmap for releasing them and how stable they are

jamuhl:
@schettino regarding the react-i18next hooks i would say it won't have any changes (or just smaller) as there was no feedback for some time and a few comments like this one: https://github.com/i18next/react-i18next/issues/590#issuecomment-443009608

so there is more or less just SSR (next.js), types and test coverage that are to be done before releasing (when react releases v16.x containing hooks: https://reactjs.org/blog/2018/11/27/react-16-roadmap.html (currently Q1 2019))

PR #1136, opened 2018-11-20
Partial bundled languages

dyniper:
This change add the ability to provide partial language bundle. When a language is bundled, then the backend connector is not invoked to load the resource.

I could not update the documentation as it is not part of the repo. I will also update the @types/i18next once this is merged.

Thank you for considering this change.


src/i18next.js
```diff
@@ -116,7 +116,8 @@ class I18n extends EventEmitter {
 
   /* eslint consistent-return: 0 */
   loadResources(callback = noop) {
-    if (!this.options.resources) {
+    if (!this.options.resources || (this.options.partialBundledLanguages &&
+      (!this.options.resources[this.languages[0]] || !this.options.resources[this.languages[0]][this.options.ns]))) {
```
jamuhl: Why you do this check? `this.options.resources[this.languages[0]]`

a) it's not sufficient to check only for language at index 0
c) [this.options.ns] can is an array anyway
b) if i remember right the backend won't load that lng-ns again if already in the resource store: https://github.com/i18next/i18next/blob/master/src/BackendConnector.js#L44

jamuhl:
There is one issue i currently see...beside that i would be ok to get this feature added.

src/i18next.js
dyniper: I'm checking the language at index 0 following you comments on a previous PR for that same feature: https://github.com/i18next/i18next/pull/1066

I checked for the this.options.ns to allow passing a partial language for only one of the namespace used by the application (where other namespace would need to be remotely loaded).

Do you mean that I could simply check for the flag and let the BackendConnector load or not the resources?

One tangent comment: it would be much easier to follow what is an array and such if typings were used. Did you consider moving to typescript?

jamuhl: the load function meanwhile changed to be reusable for the forced reload.

so i believe just checking for the flag would be sufficient -> but you should test that to be sure.

regarding typescript...won't use it for this module itself...but the typedefinitions will / should be added for next upcoming major version: https://github.com/i18next/i18next/issues/1129

dyniper: Just tested, and it works as expected. Submitting a change very soon.

dyniper:
Question: Should i also commit the built file? (<root>/i18next.js and <root>/i18next.min.js)

jamuhl:
> Question: Should i also commit the built file? (/i18next.js and /i18next.min.js)

nope...will do the build when releasing

src/i18next.js
dyniper: Added a commit that looks just at the flag.

jamuhl:
thank you - this was published in i18next@12.1.0

gorhawk:
Sorry, I didn't feel like opening a separate issue for this question..
Does this feature have an advantage over having an additional backend added to i18next-chained-backend?
The example case goes like this: I have some languages/namespaces that don't want to fetch ->

1. the first backend from the chain simply reads these from JS objects from the source code.
or
2. I provide these in resources in init (from 12.1.0 on)

jamuhl:
@gorhawk both will work (while the solution with the chain worked before - providing resources on init would have not allowed loading in the past - only with setting this new flag i18next will check if it needs to load additional files)

PR #1142, opened 2018-11-30
include typescript definition file

emilgpa:


emilgpa:
Yes, it is exactly the same. The only change was `export default i18next` instead of` export = i18next` in the last line.

lenovouser:
I think you guys got something wrong with the `export default`?

In https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30967 @emilgpa is saying that i18next has a default export but it actually doesn't - you actually explicitly make the default one a standard one in

https://github.com/i18next/i18next/blob/e7dd382801e88524ebb1b9285eda215e66848b9e/index.js#L1

I think the problem that was discussed in https://github.com/i18next/react-i18next/issues/627 is actually one originating from the webpack importing mechanism? Maybe it doesn't use the `./index.js` but actually something else from 

https://github.com/i18next/i18next/blob/e7dd382801e88524ebb1b9285eda215e66848b9e/package.json#L5-L7

His problem probably comes from the different exporting mechanisms in these files.

Anyways, this change does break `i18next` with TypeScript on Node because Node uses the standard `./index.js` file

jamuhl:
@lenovouser are you sure it's not right?

node.js -> use the export.default from the commonjs build
es modules -> while you can use the individual function in the index file to just call init, ... best is to import the default for usage

the main problem is people doing `import * as i18n from 'i18next'` and thinking they got a instance of i18next they can use in other modules like react-i18next (which is obviously wrong).

so an option could be to remove all those functions here: https://github.com/i18next/i18next/blob/master/src/index.js#L5

so people do no longer struggle

lenovouser:
Yeah the problem is that you can't emulate that in the types.

So for NodeJS there is no default export and if I would do `import i18next from 'i18next'` in TypeScript it would try to use `require('./index.js').default` which doesn't exist.

You could remove the `./index.js` file and point the `main` property in the `package.json` directly to the CommonJS one. Then both would use a default export if I see that correctly. But I don't know if that would break anything.

jamuhl:
but that would break node.js case for non-typescript:

```js
const i18n = require('i18next');
const realI18n = i18n.default;
```

jamuhl:
@schettino @emilgpa really would get your opinions on this

emilgpa:
I support the idea that in TypeScript should be removed the option to import i18n as `i18n as import from "i18next"` But I do not have experience in the NodeJS part nor with other module loader other than the standard one, and due to this I can not be of much help ...

jamuhl:
@schettino would be happy to get your feedback soon...so we could release v13 soon

schettino:
It's been a little tricky to see where the changes have been made. I ran a diff here with your changes and what we got in DT and there a bunch of removed code. All of this is matching the next release? Maybe you could go through the lines you're changing/removing (as it's not that much), just to point out them and thus we can make sure everything is in place. What do you think?

jamuhl:
merging this now...seems either it will break or not - guess will only find out when released - asked endlessly for verification without luck

jamuhl:
published in i18next@13.0.0

lenovouser:
@jamuhl I maintained these types at `definitelytyped/definitelytyped` and this breaks standard NodeJS + TypeScript usage like I said. This requires writing `import i18next from 'i18next'` which is a default import that doesn't exist for NodeJS.

jamuhl:
@lenovouser still doing `import * as i18n from 'i18next'` is even more wrong - causing more trouble 

point is i do not use typescript and therefore never got all this problems. But like i understand there is no real way of getting a "one solution fits all" - so either i get problems on node or in react-i18next when typescript is used. Right? Only solution would be remove default exports and only use named.

nikolay-borzov:
Looks like these definitions are different from [ones](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/i18next/index.d.ts) from DefinitelyTyped.

For example `DetectionPluginOptions`, `Services` are missing

emilgpa:
@nikolay-borzov Yes, it looks different.

Possibly it is a PR merged before I copy it.

nikolay-borzov:
Are there any plans to move types defined on DefinitelyTyped to this repository (I mean remove definitions from DefinitelyTyped )? Currently, there are two sources of "truth". I cannot use one from `@types/i18next` because TypeScript prefers ones from `i18next` package.

jamuhl:
@nikolay-borzov you might help to improve the definitions here...? Just would suggest not overwriting those here as this should fix some issues with import for esmodules (while breaking it for nodejs)

emilgpa:
Definitely it would have to eliminate the definition in DefinitelyTyped.
Right now I'm at work and I can not update the definition. I could do it at night if @nikolay-borzov is busy.

stunaz:
yeah, unfortunately this is a issue
screen shot 2018-12-17 at 9 07 40 am

nikolay-borzov:
> @nikolay-borzov you might help to improve the definitions here...? Just would suggest not overwriting those here as this should fix some issues with import for esmodules (while breaking it for nodejs)

I'd like to add definitions for Plugin/Module classes. I'm at work right now.

nikolay-borzov:
> yeah, unfortunately this is a issue

I'm not sure you can reference type like this. I might be wrong.
Have you tried using the namespace to access the type - `i18n.TranslationFunction`?

stunaz:
@nikolay-borzov I could before. your proposition worked. But I am wondering where I should import the i18n from i.e. like this `import i18n from 'i18next';` or from my instance i.e. like `import i18n from 'my-instance-of-i18n';` ?

nikolay-borzov:
I think i18n namespace is globally accessible without imports. E.g. If you define an `interface A` inside `d.ts` and put this file inside your project folder - then you can reference `A` from any place within project files.

nikolay-borzov:
FYI I'm working on updating definitions - take latest from DefinitelyTyped  and add types for plugins

sigerello:
Please, consider to revert to `export = i18next;`.
Your change breaks importing of types and interfaces.
For example now such import doesn't work as expected:
`import {i18n, InterpolationOptions} from 'i18next'`

jamuhl:
@sigerello that ever worked correctly? the export on index.js is still the same https://github.com/i18next/i18next/blob/master/src/index.js and that does not export i18n 

only the default export on index.js is a real instance so doing something like `import * as i18n from 'i18next'` will give people a bad time

sigerello:
@jamuhl, yes, I use it in typescript environment.
I have the following import statements in my project:
```
import {i18n, InterpolationOptions} from 'i18next'
```
It worked without any problems before, but now it's totally broken, so I was forced to return to `v12.1.0`.

emilgpa:
@sigerello you should to use `import i18n, {InterpolationOptions} from "i18next"`
With webpack, `import * as i18next from 'i18next'` not working properly.

lenovouser:
> @sigerello you should to use `import i18n, {InterpolationOptions} from "i18next"`
> With webpack, `import * as i18next from 'i18next'` not working properly.

@emilgpa that doesn't work because there is no default export in the node environment.

The default export gets mapped to export:

https://github.com/i18next/i18next/blob/55bcc6dc132491c5c21d98ebe338ccb24d978283/index.js#L1

PR #1143, opened 2018-12-03
refactor non valid keys handling

ddemirkhanyan-vineti:
The translate function was breaking in cases when the passed argument to the function is an object. This should handle pretty much all of the cases.


jamuhl:
published in i18next@13.0.0

Issue #1145, opened 2018-12-12
Request: add options to missingInterpolationHander


johnburkert:
It would be nice for the options passed to Interpolator.interpolate to be also passed to the missingInterpolationHandler. Then in the handler you could lookup the value like a nested value. I need this for the scenario where the variables needed for a message are not known (or could change) and we want formatting (so nesting is no adequate).

    
    const i18next = require('i18next');
    
    function formatHandler(val, fmt, lng) {
    	if (val) {
    		switch (fmt) {
    		case 'uppercase': return val.toUpperCase();
    		case 'lowercase': return val.toLowerCase();
			case 'titlecase':
				return val.toLowerCase().split(' ').map((word) => {
					return word.replace(word[0], word[0].toUpperCase());
				}).join(' ');
			case 'currency':  return '$' + val;
    		}
    		return val;
    	}
    	// returning undefined will trigger missingInterpolationHandler
    	return undefined;
    }
    
    function missingHandler(str, match, opts) {
    	if (match && match.length > 1) {
    		// match[1] is the contents of {{ }} 
    		const matches = match[1].match(/([\w_\.]+)(?:\s*,\s*(\w+))?/);
    		if (matches && matches.length > 1) {
    			// matches[1] is key, matches[2] is format
    			const val = i18next.t(matches[1], opts);
    			if (val) {
    				if (matches.length > 2) {
    					return formatHandler(val, matches[2]);
    				}
    				return val;
    			}
    		}
    	}
    	return undefined;
    }
    
    i18next.init({
    	lng: 'en',
    	debug: false,
    	resources: {
    		en: {
    			translation: {
    				"user_name": "Cookie Monster",
					"user_balance": 1.23,
    				"welcome_message_1": "HELLO {{ user_name, uppercase }}!",
    				"welcome_message_2": "{{ welcome_message_1, titlecase }} you have {{ user_balance, currency }}",
					"welcome_message_3": "Hello, how are you today?",
    			}
    		}
    	},
    	interpolation: { format: formatHandler },
    	missingInterpolationHandler: missingHandler
    }, (err, t) => {
    	console.log('i18next is ready to go');
    });
    
    console.log(i18next.t('welcome_message_1'));
    console.log(i18next.t('welcome_message_2'));
	console.log(i18next.t('welcome_message_3'));

jamuhl:
PR would be welcome.

jamuhl:
Sorry just saw you already provided a PR - will check it later today 👍

[...]

PR #1146, opened 2018-12-12
added options as arg to missingInterpolationHandler

johnburkert:
#1145


jamuhl:
published in i18next@13.0.0

Issue #1150, opened 2018-12-20
Default value as param does not support interpolation


messerm:
https://www.i18next.com/translation-function/essentials#passing-a-default-value

The example below with default value as parameter will return `default value test` instead of `default value 1 test`
`i18n.t('key1', 'default value {{value}} test', { value: 1 });`

Is this a known limitation?

Tested with v13.0.0

messerm:
My first intention would be to extend the logic for `overloadTtranslationOptionHandler()` in https://github.com/i18next/i18next/blob/master/src/defaults.js#L39 like

```
 overloadTranslationOptionHandler: function handle(args) {
      var ret = {};
      if (args[1]) ret.defaultValue = args[1];
      // pass in interpolation values
      if (typeof args[2] === 'object') {
        Object.keys(args[2]).forEach(function(key) {
          ret[key] = args[2][key];
        } )
      }
      else if (args[2]) ret.tDescription = args[2];
      return ret;
    },
```

jamuhl:
will need to check if this could introduce some issues...but guess might work.

Was on intention as you still can `i18n.t('key1',  { defaultValue: 'default value {{value}} test', value: 1 });`

messerm:
Not sure what you mean with "Description" and "in form of an object".
Have a look at the test cases of the fix, I hope this helps to clarify your expectations: https://github.com/i18next/i18next/pull/1151/files#diff-8a1b45bc53258a31adcedc23c8319059R10

jamuhl:
published in i18next@13.1.0

PR #1151, opened 2018-12-20
Support interpolation for defaultValue as parameter

messerm:
Possible fix for #1150


jamuhl:
published in i18next@13.1.0

PR #1152, opened 2018-12-20
update typescript definition file

nikolay-borzov:
- Get latest from https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/i18next
- Set return type to `Promise` for `init`, `changeLanguage`, `loadNamespaces`, `loadLanguages` and `reloadResources` methods
- Remove `DetectionPluginOptions`, `BackendPluginOptions` and `CachePluginOptions` since simple `object` work in the same way
- Specify range of values for `compatibilityJSON`
- Add interfaces for modules: `backend`, `languageDetector`, 'postProcessor' and `logger`
- Uppercase first letter in comment


jamuhl:
> But there is a callback call w/o a check

Will need to change that to: `if (callback) callback();`

jamuhl:
published with i18next@13.0.1

Issue #1157, opened 2019-01-07
init() typings


webleaf:
Such code produce error:

```
i18next
    .init({
        interpolation: {
            format: (value, format, lng) => {

            }
        }
    })
    // .then(() => console.log('i18n init done'))
    // .catch(e => console.log(e))
```
```
TS2345: Argument of type '{ interpolation: { format: (value: any, format: string, lng: string) => void; }; }' is not assignable to parameter of type 'Callback'.
  Object literal may only specify known properties, and 'interpolation' does not exist in type 'Callback'.
```

jamuhl:
@emilgpa @nikolay-borzov @schettino does this make sense? If possible i would love to get your feedback before merging?

nikolay-borzov:
TypeScript type errors are confusing sometimes. I think the problem is within
```
format: (value, format, lng) => {}
```

format function must return `string`
```
type FormatFunction = (value: any, format?: string, lng?: string) => string;
```

Try this
```
i18next
    .init({
        interpolation: {
            format: (value, format, lng) => ''
        }
    })
    // .then(() => console.log('i18n init done'))
    // .catch(e => console.log(e))
```

nikolay-borzov:
As for overloads order. TypeScript handbook states:

> In order for the compiler to pick the correct typecheck, it follows a similar process to the underlying JavaScript. It looks at the overload list, and proceeding with the first overload attempts to call the function with the provided parameters. If it finds a match, it picks this overload as the correct overload. For this reason, it’s customary to order overloads from most specific to least specific.

Looking at current definitions, the first one looks most specific to me
```
init(options: InitOptions, callback?: Callback): Promise<TranslationFunction>;
init(callback?: Callback): Promise<TranslationFunction>;
```

webleaf:
@nikolay-borzov you are right. But in practice we get what we get.
Other order for that case produce more correct error for my code:
```
Error:(105, 13) TS2322: Type '(value: any, format: string, lng: string) => void' is not assignable to type 'FormatFunction'.
  Type 'void' is not assignable to type 'string'.
```
Also it would be useful to have exported interfaces. Then, for example, we could specify type `<InitOptions>` for object and "mute" ts compiler.

webleaf:
@nikolay-borzov 

> Looking at current definitions, the first one looks most specific to me

But may be TS compiler thinks different :) More attributes may be looks like more common case (because cover more cases).

nikolay-borzov:
[...]

PR #1158, opened 2019-01-07
fix init() attributes typings

webleaf:
#1157


jamuhl:
was published in i18next@13.1.1 - thank you for taking the time to provide and explain the need of this PR 👍

PR #1163, opened 2019-01-10
change function value type to generic function interface

tkow:
- change type arrow function  to generic function interface to enable to preserve it generic parametric function type.

see https://github.com/i18next/react-i18next/pull/665 ,too.


jamuhl:
@rosskevin...might ask again for help for this one?

rosskevin:
I'm going to hold this PR until we have baseline tests in place.

tkow:
@jamuhl I copy ts tests and configure linter. About this PR, it may satisfy enough to no breaking compatibility of typing.

tkow:
I'm sorry, I have a mistake to contaminate WithT interface commit. Then I rebase it, it doesn't include in this PR. Please check it.

rosskevin:
@tkow this PR is now breaking - and looking at it now that we have tests running on `master`.  The tests are breaking even though it doesn't indicate so.  Please see https://travis-ci.org/i18next/i18next/builds/478478184 and look at fixing.  You can run `npm run test:typescript` to test.  Also there is a `npm run prettier:typescript` so we can avoid style changes in diffs.

tkow:
@rosskevin 
This may have caused to generic types parameters new orders I'm going to introduce of tests, preserving old for no BC, I'm going to preserve old one in this pull requests, but may I change the orders to be more handy?
My pr closed is preserved type-params orders for TranslateFunction and changed them for WithT but 
 I found the differences lead easy to miss orders through the case, so I'll change type orders.

PR #1166, opened 2019-01-11
Add `transEmptyNodeValue` to `ReactOptions`

rbuetzer:
This PR adds typings for option `transEmptyNodeValue` introduced in https://github.com/i18next/react-i18next/pull/462.


rosskevin:
@jamuhl this is worth publishing a patch release.

jamuhl:
published i18next@13.1.2

PR #1167, opened 2019-01-11
Run prettier on typescript files for easier diffing in PRs

rosskevin:


jamuhl:
@rosskevin could you ping me when you like this stuff to be published to npm - guess right now we could wait on other still pending stuff?

PR #1170, opened 2019-01-16
fix: tsconfig typo and npmignore

### LINKED ISSUE #1169: Exclude tsconfig.json in NPM distribution
Might want to add `tsconfig.json` file to `.npmignore` to exclude it from NPM distribution.

With `tsconfig.json` inside `node_modules/i18next`, VSCode shows error message about `No inputs were found in config file`.

Also there is typo in the `include` section:

```json
"include": ["./indext.d.ts", "./test/**/*.ts*"]
```

`./indext.d.ts` -> `./index.d.ts`

If you fix the typo, then VSCode error is gone, but I still think `tsconfig.json` should be excluded from NPM distribution.

panjiesw:
Found myself to have time for this

resolve #1169

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,3 +14,4 @@ sample.html
 .codeclimate.yml
 .coveralls.yml
 .travis.yml
+tsconfig.json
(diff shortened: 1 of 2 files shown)


rosskevin:
Thanks @panjiesw.  @jamuhl a patch release for this would be good.

PR #1168, opened 2019-01-12
Feature/add WithT interface

tkow:
- implements WithT interface for 

This PR should be merged after https://github.com/i18next/i18next/pull/1163. This includes all of them.

https://github.com/i18next/i18next/pull/1164

Please after review, squash merge or tell me renewal to make new PR  from master branch.

PR #1172, opened 2019-01-16
Pull up WithT interface allowing for overrides

tkow:
This remains https://github.com/i18next/i18next/pull/1168.

- add WithT inteface and the test

/related https://github.com/i18next/react-i18next/pull/665


rosskevin:

jamuhl:
@rosskevin published in i18next@13.1.3

PR #1175, opened 2019-01-17
Improve t function return type

VincentLanglet:
I'd like to have a more specific type for t function.
Am I wrong or it's always a "string" ?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -476,7 +476,7 @@ declare namespace i18next {
     /**
      * Please have a look at the translation functions like interpolation, formatting and plurals for more details on using it.
      */
-    t<TKeys extends string = string, TValues extends object = object, TResult = string extends string> (key: TKeys | TKeys[], options?: TranslationOptions<TValues>): TResult;
+    t<TKeys extends string = string, TValues extends object = object, TResult extends string | object | Array<string | object> = string | object | Array<string | object>> (key: TKeys | TKeys[], options?: TranslationOptions<TValues>): TResult;
   }
 
   interface Resource {
--- a/index.d.ts
+++ b/index.d.ts
@@ -469,14 +469,14 @@ declare namespace i18next {
   }
 
   /**
-     * WARNING: Order with generic type parametes are defferent from TranslationFunction for usability
+     * WARNING: Order with generic type parameters are different from TranslationFunction for usability
      * see: https://github.com/i18next/react-i18next/issues/662
    */
-  interface WithT{
+  interface WithT {
     /**
      * Please have a look at the translation functions like interpolation, formatting and plurals for more details on using it.
      */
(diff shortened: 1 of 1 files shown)


jamuhl:
return value of `t` can be string or object, array

VincentLanglet:
Thanks @jamuhl ; I changed the type

I'll be happy to have the review of the previous author : @tkow

tkow:
No problem for me. It's good improvement.

rosskevin:
@jamuhl This can be a patch release. Not urgent, can be included in any next batch release.

Jessidhia:
This turned out to have been surprisingly breaking. I'm just not sure if it's a good or a bad break.

Every single use of `t` where an `object` result is not expected is now a type error.

---

I wonder if there is a way to declare, somewhere, what kind of result a specific key should have. Because of how typescript generic arguments work, the TResult is very inconvenient to declare (you're forced to pass the key name twice and then the type of TValue). Finding a way to that could make TValue itself safer.

This could be made non-breaking by changing the _default_ to be the same as the old value. It's still not perfect, though.

---

Getting types right for i18next will probably require adding generics in a bunch of other places, including the interfaces and the components from `react-i18next` and other bindings, not just to the function inside `WithT`. As it stands, **all of those generic arguments are hidden `any`s / hidden `as` casts**. It's only the defaults and the extends that are making them not behave _directly_ like that.

Jessidhia:
Why these arguments are "hidden anys" (well, "hidden whatever the type of the `extends` are"): they are never related to anything. As in, they are never used to create a relation between two types, or between an input and an output.

Let's take this `t` function, formatted for readability:

```ts
t<
  TKeys extends string = string,
  TValues extends object = object,
  TResult extends string | object | Array<string | object> =
    | string
    | object
    | Array<string | object>
>(
  key: TKeys | TKeys[],
  options?: TranslationOptions<TValues>
): TResult
```

The type of `key` is exactly the same as `string | string[]`. The type of `options` is exactly the same as `TranslationOptions<options>`. And the type of `TResult` is exactly the same as its default.

Because `TKeys` is never used to relate the `key` to, say, the `options` (perhaps by mapping through a lookup table given as generic argument to `WithT`?), it will just be inferred to whatever it is that you pass to it.

The same goes for `TValues` itself, and `TResult`. They are only "useful" if you explicitly write the generic argument... which means they're about the same as `as` casts.

rosskevin:
See the master where the ts usage tests are broken apart. @tkow generates TKeys for usage. This is probably better filed as an open issue with tagged participants. Commenting on a merged PR is not bound to result in a response or change.

rosskevin:
BTW @Jessidhia I think I have encountered what you are experiencing while refactoring and understand the cause of your issue - a mismatch between TranslateFunction and WithT.  I'm working on it here #1180

PR #1179, opened 2019-01-22
Fix bug when import by ES Module

tmt003-ho:


jamuhl:
good catch - always miss that es module stuff :(

jamuhl:
published in i18next@13.1.5

jamuhl:

PR #1180, opened 2019-01-22
Refactor generics usage

rosskevin:
BREAKING type change for those using generics and importing/reusing types.  Typical usage in TS without importing `TranslateFunction` or generics should remain unaffected.

This fixes a regression introduced by #1163 and related PRs, but in turn disables use of generics in `i18next.t<>` pattern of usage. - see https://github.com/i18next/i18next/pull/1180#issuecomment-456859449

As I searched for a solution to this mock test, after a suggestion below by @VincentLanglet I noticed we would be moving towards duplication.  As I attempted to solve this, I think  (due to duplication), I believe I ran into https://github.com/i18next/i18next/pull/1175#issuecomment-456754506.

In order to solve this, since I needed to remove duplication and alter potential usage of parameterized arg order, the easiest way to signal this to TS users is to change the exported name so that they will intentionally break and need to review their usage (sorry everyone :wave: - seems like the best option)

- `TranslateFunction` -> `TFunction`
- `TranslateOptions` -> `TOptions`
- `WithT.t` - removed definition and extracted to `TFunction`
- `exists` - stopped (mis)using `TranslateFunction` and defined separately
- reorder the `TFunction` args to make easier for typical use (result first)
- default `TFunction` to return `string` by default.  For arrays and objects, parameterize the call e.g. `t<object>('foo')`

---
Mock test initiated by https://github.com/i18next/react-i18next/pull/684 @nikolay-borzov 

The mock should have been simple - but it wasn't.


VincentLanglet:
@rosskevin I saw that

First of all, I think as expected the error from
```
const mockWithT: i18next.WithT = {
  t: (key: string) => key,
};
```

It should be
```
const mockWithT: i18next.WithT = {
  t: (key: string | string[]) => key,
};
```

But It still does not fix the error. And I actually don't know why

rosskevin:
Agreed @VincentLanglet - I tried a variety of things - none of which worked.  The specific error from what you mentioned is:

```
test/typescript/mock.test.ts:7:3 - error TS2322: Type '(key: string | string[]) => string | string[]' is not assignable to type '<TKeys extends string = string, TValues extends object = object, TResult extends string | object | (string | object)[] = string | object | (string | object)[]>(key: TKeys | TKeys[], options?: TranslationOptions<TValues> | undefined) => TResult'.
  Type 'string | string[]' is not assignable to type 'TResult'.
    Type 'string' is not assignable to type 'TResult'.

7   t: (key: string | string[]) => key,
    ~

  index.d.ts:479:5
    479     t<
            ~~
    480       TKeys extends string = string,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ...
    488       options?: TranslationOptions<TValues>,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    489     ): TResult;
        ~~~~~~~~~~~~~~~
    The expected type comes from property 't' which is declared here on type 'WithT'
```

I normally parameterize the interface and use in the method.  I haven't tried that because I don't see how that would make a difference but perhaps I should.  @tkow - thoughts since you introduced this originally?

rosskevin:
BTW - the diff is now cleaned up and much simpler.

VincentLanglet:
@rosskevin I get no error this way

```
interface TFunction<
  TKeys extends string = string,
  TValues extends object = object,
  TResult extends string | object | Array<string | object> =
    | string
    | object
    | Array<string | object>
> {
  (key: TKeys | TKeys[], options?: TranslationOptions<TValues>): TResult;
}

interface WithT {
  t: TFunction;
}

const mockWithT: WithT = {
  t: (key: string | string[]) => key,
};
```

rosskevin:
Given that @VincentLanglet - I think we need to make a type-breaking change and just update the `TranslationFunction` because your `TFunction` duplicates it with diff parameterized arg order.  I'll make an attempt.

VincentLanglet:
@rosskevin I Agree

rosskevin:
Ok, I think I have cleaned it up, though the generics test fails here:

```
i18next.t<string, KeyList>('friend', { myVar: 'someValue' });
i18next.t<string, KeyList>(['friend', 'tree'], { myVar: 'someValue' });
i18next.t<string, KeyList, { myVar: 'someValue' }>('friend', { myVar: 'someValue' });
i18next.t<string, KeyList, { myVar: 'someValue' }>(['friend', 'tree'], { myVar: 'someValue' });

Expected 0 type arguments, but got 2.
Expected 0 type arguments, but got 2.
Expected 0 type arguments, but got 2.
Expected 0 type arguments, but got 2.
```

Thoughts?  Please check the updated description for a summary.  I am tempted to comment out this test until @tkow has time and get this merged/released as I think it is an important improvement that I think others may be running into.

VincentLanglet:
@rosskevin Indeed, this way we lose the ability to use `t<A,B,C>(...)`
@tkow or @Jessidhia will be better help than me

Since the PR of tkow introduced a regression, I understand you comment out theses tests, in order to fix it. We will search how to have a better t function after.

VincentLanglet:
There is still a debate about the choice

No breaking change, but not classic
```
interface TFunction<
  TResult extends string | object | Array<string | object> =
    | string
    | object
    | Array<string | object>,
  TValues extends object = object,
  TKeys extends string = string
> {
  (key: TKeys | TKeys[], options?: TranslationOptions<TValues>): TResult;
}
```
Versus

Breaking change, but classic
```
interface TFunction<
  TKeys extends string = string,
  TValues extends object = object,
  TResult extends string | object | Array<string | object> =
    | string
    | object
    | Array<string | object>
> {
  (key: TKeys | TKeys[], options?: TranslationOptions<TValues>): TResult;
}
```

rosskevin:
@VincentLanglet the goal has always been to support classic usage.  I just added some to t.test to demonstrate expected/typical use:

```ts
// various returns <string> is the default
i18next.t<object>('friend');
i18next.t<string[]>('friend');
i18next.t<object[]>('friend');
```

This should be the simplest use I think.

rosskevin:
The current push works for everything except the mock - back to square one but with an improved basis.  I strangely have to duplicate `TFunction` inside `WithT` to retain parameterized type usage with `i18next.t<>`.  So close...

rosskevin:
Duplication removed, change init test to rely on inference instead of explicit types for callbacks, still have the mock issue.

This works now without explicit types (unless you want the return type of object or string[]).  For parameterized `TKeys` use, you now have to use the second arg, which makes sense because most won't do that.

If I can get a good solution to the mock test, I think this is good to go.

rosskevin:
@jamuhl I need a release for this.  This is a breaking typescript change (see description) - it primarily fixes a regression introduced but to do so had to change order of parameterized args.  The best way to do not silently cause a lot of confusion for TS users is to break it in very obvious way - change the name of the interfaces that were affected.

A major bump would make sense for TS users, but not for js users - your call on versioning.  If you aren't worried, I would major bump it.

jamuhl:
@rosskevin published in i18next@14.0.0

PR #1183, opened 2019-01-23
Add lint-staged prettier

rosskevin:
For the sake of PRs and keeping files in shape for reviews, I'd like to add lint-staged and prettier.  This adds the config in the `package.json` in one commit, and performs a baseline `npm run prettier` so the repo is setup for new PRs.

If we want the prettier settings changed, let's get the `.prettierrc` corrected and I'll add this to react-i18next as well if approved.

### DIFF
--- a/src/PluralResolver.js
+++ b/src/PluralResolver.js
@@ -3,260 +3,74 @@ import baseLogger from './logger.js';
 // definition http://translate.sourceforge.net/wiki/l10n/pluralforms
 /* eslint-disable */
 let sets = [
-  {
-    lngs: [
-      'ach',
-      'ak',
-      'am',
-      'arn',
-      'br',
-      'fil',
-      'gun',
-      'ln',
-      'mfe',
-      'mg',
-      'mi',
-      'oc',
-      'pt',
-      'pt-BR',
-      'tg',
-      'ti',
-      'tr',
-      'uz',
-      'wa',
-    ],
-    nr: [1, 2],
-    fc: 1,
-  },
+  { lngs: ['ach','ak','am','arn','br','fil','gun','ln','mfe','mg','mi','oc', 'pt', 'pt-BR',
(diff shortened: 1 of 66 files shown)


src/PluralResolver.js
```diff
@@ -3,74 +3,260 @@ import baseLogger from './logger.js';
 // definition http://translate.sourceforge.net/wiki/l10n/pluralforms
 /* eslint-disable */
 let sets = [
-  { lngs: ['ach','ak','am','arn','br','fil','gun','ln','mfe','mg','mi','oc', 'pt', 'pt-BR',
```
jamuhl: can we exclude those rules stuff from getting changed by prettier (it does not really add to readablity)

src/PluralResolver.js
```diff
@@ -3,74 +3,260 @@ import baseLogger from './logger.js';
...
+  { lngs: ['mnk'], nr: [0, 1, 2], fc: 18 },
+  { lngs: ['mt'], nr: [1, 2, 11, 20], fc: 19 },
+  { lngs: ['or'], nr: [2, 1], fc: 2 },
+  { lngs: ['ro'], nr: [1, 2, 20], fc: 20 },
+  { lngs: ['sl'], nr: [5, 1, 2, 3], fc: 21 },
+  { lngs: ['he'], nr: [1, 2, 20, 21], fc: 22 },
+];
 
 let _rulesPluralsTypes = {
-  1: function(n) {return Number(n > 1);},
```
jamuhl: also this rule functions (keep them as is)

src/PluralResolver.js
rosskevin: yes, we can ignore specific files - I'll add that

rosskevin:
Ok @jamuhl - PluralResolver added.  Please check out the diff now.  We'll need to adjust codeclimate - and I can do that, just want to get our settings correct first before I make codeclimate accept it.

PR #1186, opened 2019-01-23
TFunction external use broken without using interface

rosskevin:
Unfortunately this duplicates the interface outside of `WithT`, but as the test shows (and breaks in react `NamespaceConsumer` https://github.com/i18next/react-i18next/issues/693) we need it (at least until we find a better way to reference it).


jamuhl:
@rosskevin i18next@14.0.1 is published

Issue #1189, opened 2019-01-30
allow 'TemplateStringsArray' as parameter of t function in type declaration


noru:
I originally create this issue on `react-i18next` side, but I found it makes more sense ( and easier ) to add type support here. Would you please consider it?

https://github.com/i18next/react-i18next/issues/701

> Is your feature request related to a problem? Please describe.
> I'd like to reduce a pair of parenthesis by using string literal:
> 
> t\`some_key\` // instead of t('some_key')
> which is 98% of the case in my code base. It doesn't effect anything, only need the type support.
> 
> I can raise a PR if it is acceptable.
> 
> Describe the solution you'd like
> simply add types to the declaration file

```javascript
// index.d.ts...
// NOTE: somewhat duplicate of WithT.t but cannot use interface - not sure why I can't find a workable reuse pattern
  type TFunction = <
    TResult extends string| object | Array<string | object> | undefined = string,
    TKeys extends string {| TemplateStringsArray} = string, //  simply add the quoted thingy, the same with the 't' in 'WithT'
    TValues extends object = object
  >(
    key: TKeys | TKeys[],
    options?: TOptions<TValues>,
  ) => TResult;
```

jamuhl:
Please provide a PR with this and close https://github.com/i18next/react-i18next/pull/705 if not needed any longer

jamuhl:
@noru you still got plans for a PR - or should we close this?

drewxiuxiuxiu:
Im sorry it's chinese new year recently. I'll pr it asap.

noru:
#1199

jamuhl:
pubished in i18next@15.0.3

PR #1199, opened 2019-02-10
#1189 typescript - accept templatestringsarray as TKey

noru:
I added a few cases to the tests and pass the `npm run test:typescript`. Please let me know if there is anything uncool.

PS:  
1. it may conflict with https://github.com/i18next/i18next/pull/1186, please check if my tweak is fine.
2. things like `t'key-with-${var}'` won't work with this declaration but I don't think that's a big deal

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -469,21 +469,12 @@ declare namespace i18next {
 
   interface WithT {
     // Expose parameterized t in the i18next interface hierarchy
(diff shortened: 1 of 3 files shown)


rosskevin:
@jamuhl this is a patch release

jamuhl:
published in i18next@15.0.3

PR #1204, opened 2019-02-13
Fix exports for node

lenovouser:
Both import methods (`import * as i18next from 'i18next'` and `import i18next from 'i18next'`) work this way. It is necessary since the browser exports use a default export but the node ones don't.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,1 +1,5 @@
-module.exports = require('./dist/commonjs/index.js').default;
+/* eslint no-var: 0 */
+var main = require('./dist/commonjs/index.js').default;
+
+module.exports = main;
+module.exports.default = main;


rosskevin:
@lenovouser I can't think of any reason why node would not understand `export default`, and as shown in the tests, your change does not work.  

The current tests run as node (and work), so perhaps you have an environment setup problem that is leading you to think this change would be necessary?

lenovouser:
@rosskevin well the TypeScript types were merged even though several people including me were mentioning that they'd break Node support. The file that Node requires doesn't have a default export:

https://github.com/i18next/i18next/blob/55bcc6dc132491c5c21d98ebe338ccb24d978283/index.js#L1

So thus only `const i18next = require('i18next')` or the TypeScript equivalent `import * as i18next from 'i18next'` works syntax-wise. But it doesn't work with the current types since they only `export default`.

lenovouser:
https://github.com/i18next/i18next/pull/1142#issuecomment-443525555

https://github.com/i18next/i18next/pull/1142#issuecomment-447270989

lenovouser:
Basically in the JavaScript code there is a `export default` for browser environments / other loaders (Rollup, Webpack etc.) and a `export =` for node environments.

lenovouser:
I have another idea how we could fix this in theory though. Will check locally if it works.

lenovouser:
@jamuhl I think this is something you need to review now instead of @rosskevin

jamuhl:
published in i18next@15.0.4

PR #1205, opened 2019-02-13
Code cleanup of TFunction tests, allow strong vs open generics on interpolation values

rosskevin:
This should effectively be no-change...

...unless someone was using the third generics type on `TFunction`, and if so, this corrects the problem of the `TOptions` being 100% open even if using generics.  Now when using generics, it won't allow any interpolation key.

I spotted the error when reviewing something else, did some cleanup and it was revealed.


rosskevin:
No release necessary.

PR #1206, opened 2019-02-13
Add defaultValue tests and allow second arg string as defaultValue

### LINKED ISSUE #1198: TypeScript: can't pass default value to "t" as second argument
Using `t('key', 'Default value')` (as documented in https://www.i18next.com/translation-function/essentials#passing-a-default-value) currently causes the following error: 

`Argument of type '"Default value"' is not assignable to parameter of type 'TOptions<object> | undefined'.`

I'm using version 14.1.1

rosskevin:
Closes #1198

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -485,7 +485,7 @@ declare namespace i18next {
     TInterpolationMap extends object = StringMap
   >(
     key: TKeys | TKeys[],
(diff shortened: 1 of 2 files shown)


rosskevin:
This can be a patch release @jamuhl

jamuhl:
published in i18next@15.0.4

PR #1208, opened 2019-02-20
Add `hashTransKey` to `ReactOptions`

medmunds:
Add typings for `hashTransKey` [introduced][1] in react-i18next 6.2.0.

[1]: https://github.com/i18next/react-i18next/commit/06fb359371f9c5a21eb4fd0148a7e739b5975fd1

(It might be more readable to define something like `type DefaultValue = any` and use that throughout.)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -145,7 +145,7 @@ declare namespace i18next {
      * that always throws an error.
      * @default undefined
      */
-    hashTransKey?(defaultValue: typeof TOptionsBase['defaultValue']): typeof TOptionsBase['defaultValue'];
+    hashTransKey?(defaultValue: TOptionsBase['defaultValue']): TOptionsBase['defaultValue'];
   }
 
   interface InitOptions {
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -137,6 +137,15 @@ declare namespace i18next {
...
     transEmptyNodeValue?: string;
+    /**
+     * Function to generate an i18nKey from the defaultValue (or Trans children)
+     * when no key is provided.
+     * By default, the defaultValue (Trans text) itself is used as the key.
+     * If you want to require keys for all translations, supply a function
+     * that always throws an error.
+     * @default undefined
+     */
+    hashTransKey?(defaultValue: typeof TOptionsBase['defaultValue']): typeof TOptionsBase['defaultValue'];
```
rosskevin: These can be just `TOptionsBase['defaultValue']` without the `typeof`

medmunds: Fixed.

(BTW, none of the other react options seem to have tests yet.)

rosskevin:

jamuhl:
published in i18next@15.0.5

Issue #1211, opened 2019-03-03
No indication wether reloadResources succeeded


slavab89:
Our use case is that we have our translations loaded on the server side for various usages. From time to time we need to reload the translations based on different conditions
We are using `reloadResources` API function to do that however it seems impossible to know wether that succeeded or not.
This is somewhat relates to https://github.com/i18next/i18next/issues/1053.
Even if the backend returns an error (We're using i18next-xhr-backend and calling `callback(err, { status: '500' })` in the `ajax` option) it doesnt populate to the `reload` function.

Would it be possible to populate the result from the backend somehow?

jamuhl:
If i see right reloadResources has callback which gives you any error as first argument

https://github.com/i18next/i18next/blob/master/src/i18next.js#L175

https://github.com/i18next/i18next/blob/master/src/BackendConnector.js#L170

slavab89:
It does, but based on the current code, `reloadResources` doesnt use the error from the callback at all.
The current code:
```js
this.services.backendConnector.reload(lngs, ns, () => {
  deferred.resolve();
  callback(null);
});
```
No matter what `reload` returns, it will always resolve the promise or callback with no error

jamuhl:
ah right...never looked at that...as reloading is not critical...as it just updates sources on success. Feel free to provide a PR.

slavab89:
I will.
Is there any other suggestion in working this way? We have some backend server that needs the translations for some of its operations - example is to generate email templates to be sent.
Once we change some translations we want to notify that backend to reload those.

We did not want to create a new instance of i18next and load the translations on every request. The goal was to lower the amount of requests that we do to get the list of translations - let i18next know that it should refresh some key/namespace/language.

jamuhl:
What we did on the backend using https://locize.io is doing a reload on interval: https://github.com/locize/i18next-node-locize-backend/blob/master/src/index.js#L67

PR #1212, opened 2019-03-03
Expose error on reloadResources

slavab89:
closes https://github.com/i18next/i18next/issues/1211

My only "question" is that i saw that you dont do `reject` on any promise here. Is that intentional?


jamuhl:
published in i18next@15.0.5

PR #1219, opened 2019-03-14
Add useSuspense to ReactOptions, fix error throwing on test

KlimekM:
Hello, first contribution to the project. Ran into this earlier today when integrating with i18next:

Screen Shot 2019-03-13 at 2 09 36 PM (1)

Turns out useSuspense is not on the ReactOptions interface, but is listed in the documentation: https://react.i18next.com/latest/i18next-instance

This PR should address the above. It may also be worthwhile to add the additional options for react in init options listed in the same link above. Happy to do so at a later time.

Ran tests prior to committing and ran into this:
Screen Shot 2019-03-13 at 10 52 31 PM

Which led the update to the init.test.ts file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -138,6 +138,11 @@ declare namespace i18next {
      */
     transEmptyNodeValue?: string;
     /**
+     * Set it to false if you do not want to use Suspense
+     * @default true
+     */
+    useSuspense?: boolean;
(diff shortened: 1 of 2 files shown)


jamuhl:
thanks a lot for your contribution. was published in i18next@15.0.7

PR #1223, opened 2019-03-24
adding init function to 3rdParty module typings and enforcing type property

saboya:
While the plugin interfaces can't be used directly since they're not exported, it still serves as reference / documentation.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -613,7 +613,8 @@ declare namespace i18next {
   }
 
   interface ThirdPartyModule {
-    type?: '3rdParty';
+    type: '3rdParty';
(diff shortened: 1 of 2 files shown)


rosskevin:

On a related note, I've been thinking about removing the ts `namespace` to expose our types - this will be a breaking change if we do.  I don't know why the originals were namespaced to begin with but it seems more of a hindrance than anything.

saboya:
@rosskevin Indeed, I didn't understand why the types weren't exported either, since it's not using export assignment. I'd like that very much.

jamuhl:
published in i18next@15.0.8

Issue #1226, opened 2019-04-02
IE10 - 'Unable to get property * of undefined or null reference'


ilanbm:
I am getting Unable to get property * of undefined or null reference when I have opened console in **IE10** and refresh the page.
The error is from EventEmitter.js _this.observers[event] = _this.observers[event] || [];

How to reproduce:
1/ Open in Internet Explorer 11
2/ Open debugging console (F12)
3/ Reload the page
4/ You should see Unable to get property * of undefined or null reference

jamuhl:
Accepting PR's regarding IE usage..

PR #1227, opened 2019-04-03
<=IE10 fix (unable to call parent constructor)

ilanbm:
#1226


jamuhl:
was published in i18next@15.0.9

PR #1228, opened 2019-04-03
feat: export TFunction for external use

iamandrewluca:


rosskevin:
@iamandrewluca this does not achieve what you are seeking.  In a declaration file, these are already exported but encapsulated under the namespace.

Currently, you access via 

```tsx
import * from 'i18next'

i18next.TFunction
```

I'd like to see the namespace removed and am open to a PR for that.  If done, will also need to be a PR for react-i18next as well (breaking change semver for both).

iamandrewluca:
@rosskevin you are right
```ts
import i18next from 'i18next'
const test: i18next.TFunction
```
or
```ts
const test: i18next.default.TFunction
```

Issue #1234, opened 2019-04-26
Treat objects as missing keys when returnObjects is false


acc15:
Currently (v15.1.0) if key points to object `i18next.t(key)` returns `undefined` even if `returnNull` and `returnEmptyString` is false.
Also if fallback key is provided its ignored resulting in `undefined`.

Simple reproduction test:

    const i18next = require("i18next");
    i18next.init({
        lng: "en",
        debug: true,
        resources: {
            en: {
                translation: {
                    a: {
                        b: "WoW"
                    }
                }
            }
        },
        returnEmptyString: false,
        returnNull: false,
        returnObjects: false,
        parseMissingKeyHandler: key => {
            console.log("Missing key: " + key);
            return key;
        }
    }).catch(err => console.log("i18n error", err));
    console.log("Single key: " + i18next.t("a")); // => undefined (must be 'a')
    console.log("Key with fallback: " + i18next.t(["a", "a.b"])); // => undefined (must be 'WoW')

May be i missed some point, 
but that behavior looks very strange 
especially when i18next ignores fallback keys.

jamuhl:
Simple:

`i18next.t('a');` exists and would return an object `{ b: "WoW" }` which is not allowed by setting `returnObjects: false`

`i18next.t(["a", "a.b"]);`  "a" exists (therefore no fallback) and would return an object `{ b: "WoW" }` which is not allowed by setting `returnObjects: false`

acc15:
Is there any way to return `key` instead of `undefined`? 

Without `returnObjects` its actually can be considered as a `typo` (like `a.`, `a.c`, etc), therefore it would be nice if i18next can help handle such kinds of typos.

acc15:
See #1288

PR #1239, opened 2019-05-06
Changed default returnedObjectHandler to return key instead of undefined

acc15:
Default `returnedObjectHandler` to return `key` instead of `undefined`
Added `key` to warning message when accessing an object and `returnObjects` is disabled

See #1234

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "i18next",
-  "version": "17.0.5",
+  "version": "17.0.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "i18next",
-  "version": "17.0.5",
+  "version": "17.0.6",
   "description": "i18next internationalization framework",
   "main": "./dist/cjs/i18next.js",
   "module": "./dist/esm/i18next.js",


jamuhl:
returning an empty object is more safe as you expect an object to be returned and not a string (key); actually we just should uncomment the default so the string message get triggered.

acc15:
> returning an empty object is more safe as you expect an object to be returned and not a string (key); actually we just should uncomment the default so the string message get triggered.

`returnObjectHandler` is called only when `returnObjects` is `false`. 
Why we expect an object when `returnObjects` is `false`?

jamuhl:
like said...best would be not having a default returnObjectHandler (my guess is that is only there as a boilerplate for documentation...). It returns an object as it's what people use it for.

https://github.com/i18next/i18next/pull/1239/files#diff-42176d501e1082b1431ec21640f0ff05R124 makes no sense as it's always true with having a default. So basically commenting the default out would result in a string message.

jamuhl:
How do you use that handler?

acc15:
Currently not using it, just found this feature (my bad).

But i agree, removing default will result in nice visible messages instead of `undefined`.

I'm using `react-i18next` binding and the key problem here in the fact that React renders `'undefined' => ''` so the message just dissapears

If i change default `returnObjectHandler` to `undefined` it will be ok?

jamuhl:
yes setting it to false should do the trick

PR #1244, opened 2019-05-10
Update BackendModule interface to allow null or undefined for the cal…

bopfer:
The current typings require that there is always an `Error` passed in to the `BackendModule` callbacks.  `null | undefined` needs to be added for callbacks with no error.  `null` because that is what is shown in the docs for creating a custom backend and it's commonly used.  `undefined` because many projects use linting rules that disallow `null` in favor of `undefined`.

I don't think there are any tests that need updating based on this change.  Please let me know if that is incorrect.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -544,7 +544,7 @@ declare namespace i18next {
     read(
       language: string,
       namespace: string,
-      callback: (err: Error, data: ResourceLanguage) => void,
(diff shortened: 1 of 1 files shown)


rosskevin:
Thanks @bopfer.  @jamuhl this can be a patch release.

jamuhl:
published in i18next@15.1.1

PR #1248, opened 2019-05-22
Fix type error when init with locize plugin options

lukyth:
This [example with locize](https://github.com/i18next/react-i18next/blob/master/example/locize/src/i18n.js) will cause a type error.

Fixed by adding an init options for [`locize-editor`](https://github.com/locize/locize-editor) and [`locize-lastused`](https://github.com/locize/locize-lastused) plugins.


jamuhl:
thank you - was published in i18next@15.1.3

PR #1249, opened 2019-05-22
handle both [array Array] and [array Object] in ResourceStore

i8ramin:
Fixes https://github.com/i18next/i18next/issues/1250

Not sure why this is the case, but for some reason the `addResources` method does not seem to handle message bundles defined as objects (`[object Object]`), but only `[object Array]`. This is the line of code in question:

https://github.com/i18next/i18next/blob/31201aae794b4ab19abc7a21735630c50c97606a/src/ResourceStore.js#L68-L71

For example, given the following key/value:

```json
{
  "footer": {
    "copyright": "Copyright © 2019"
  },
}
```

The value of `Object.prototype.toString.apply(resources[m])` is actually `[object Object]` and not `[object Array]`. With this change, it will check for both (although I am not sure when it would be `[object Array]`)


src/ResourceStore.js
```diff
@@ -67,7 +67,7 @@ class ResourceStore extends EventEmitter {
     for (const m in resources) {
       if (
         typeof resources[m] === 'string' ||
-        Object.prototype.toString.apply(resources[m]) === '[object Array]'
+        /[object (Object|Array)]/.test(Object.prototype.toString.apply(resources[m]))
```
i8ramin: Main change is here. Not sure why the above file was also modified when I ran `npm run build`

jamuhl:
The problem of accepting objects as values is - if you already have:

`labels.name`

and now import a

`labels: { email, street }` you will loose the labels.name as it does not a merge

So this could lead to "dangerous" situations if devs not know what they do

jamuhl:
closing (reason above)

PR #1255, opened 2019-06-02
Provide esm babel helpers for esm build

TrySound:


jamuhl:
does this not also set esModules for the commonjs build?

TrySound:
Commonjs is built without jsnext env
```js
    "build:es": "BABEL_ENV=jsnext babel src --out-dir dist/es",
    "build:cjs": "babel src --out-dir dist/commonjs",
```

jamuhl:
you're right...

jamuhl:
will merge this when I create the next version

TrySound:
I have a few more improvements. Will send after merge.

TrySound:
Why do you need iife, amd and umd at the same time? UMD already supports iife and amd out of the box. I can send PRs with removing them.

jamuhl:
iife and amd are not builded anymore: https://github.com/i18next/i18next/blob/master/package.json#L101 -> build task does not run them...so yes...can be removed

TrySound:
Could you merge this and I will submit PR with bundles

PR #1256, opened 2019-06-03
Bundle all entry points with rollup

TrySound:
Similar to https://github.com/i18next/i18next-chained-backend/pull/8

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,15 +1,3 @@
 {
-  "env": {
-    "development": {
-      "presets": ["@babel/preset-env", "@babel/react"],
-      "plugins": ["@babel/plugin-transform-runtime"]
-    },
-    "jsnext": {
-      "presets": [
-        ["@babel/preset-env", { "modules": false, "useBuiltIns": false }],
-        "@babel/react"
-      ],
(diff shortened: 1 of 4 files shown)


jamuhl:
thanks again...really appreciate your help with all this

published in i18next@16.0.0

TrySound:
My pleasure

jamuhl:
@TrySound how you handle that problem nowadays: https://github.com/i18next/i18next-express-middleware/pull/182/files#diff-1fdf421c05c1140f6d71444ea2b27638L32

in the past pkg.main was https://github.com/i18next/i18next/blob/master/index.js to avoid this problem of default export

jamuhl:
@TrySound my suggestion would be to remove those https://github.com/i18next/i18next/blob/master/src/index.js#L5 named export shortcuts to most used functions

-> they once were added on request to get things like `import { changeLanguage } from 'i18next'`

but was never a big fan of that (hides the i18next is an instance of a class where you can have multiple during runtime if needed)...

jamuhl:
@TrySound made that change in i18next@17.0.0 and also removed the build artefact from root

TrySound:
Ah, cool. Yeah, mixing default and named exports usually leads to problems like this.

PR #1263, opened 2019-06-05
Allow null as TResult

### LINKED ISSUE #1254: TResult inferred incorrectly when result of t passed into a function accepting null
If I pass the return value of calling the `t` function, to a function that accepts `string | null`, Typescript seems to be doing some kind of inference on `TResult` where it assumes `string | object | (string | object)[] | undefined` instead of the default `string`. Codesandbox: https://codesandbox.io/s/thirsty-tdd-exc2x

```
import i18n from "i18next";
import "./styles.css";

function operateOnString(x: string | null) {
  return x;
}

const definitelyAString = i18n.t("hello");

// Works
operateOnString(definitelyAString);

// Type Error Argument of type 'string | object | (string | object)[] | undefined' is not assignable to parameter of type 'string | null'.
operateOnString(i18n.t("hello"));
```

I'm lucky enough where I can just modify my functions/components to accept `string | undefined` instead (which doesn't cause this issue), but would be ideal if `null` didn't result in inferring a value other than the default `string`.

rosskevin:
Closes #1254

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -494,7 +494,7 @@ declare namespace i18next {
   }
 
   type TFunction = <
-    TResult extends string | object | Array<string | object> | undefined = string,
+    TResult extends string | object | Array<string | object> | undefined | null = string,
     TKeys extends string | TemplateStringsArray = string,
     TInterpolationMap extends object = StringMap
   >(
--- a/test/typescript/t.test.ts
+++ b/test/typescript/t.test.ts
@@ -31,8 +31,22 @@ function defautValue(t: i18next.TFunction) {
   t('translation:test', 'test_en');
 }
 
+function callsMethodWithOptionalNullArg(t: i18next.TFunction) {
+  function displayHint(hint?: string | null) {
+    return String(hint);
+  }
+  displayHint(t('friend'));
+}
+
 function callsMethodWithOptionalArg(t: i18next.TFunction) {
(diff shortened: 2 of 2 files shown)


rosskevin:

jamuhl:
published in i18next@17.0.2

tdreyno:
@rosskevin This seems to have broken usage within React with Typescript (using v3.5.2), which allows `undefined` as a valid result, but not `null`.

The error is:

```
Type 'null' is not assignable to type 'string | number | object | ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)> | (string | ... 3 more ... | undefined)[] | undefined'.
```

Which would require adding `!` to every `t` to fix:

```
{t('path')!}
```

Or 

```
{t('path') || undefined}
```

rosskevin:
@tdreyno please PR a breaking test and tag me and I'll take a look.  As-is this worked without any test changes (just additions).

tdreyno:
Ah, you were right. I had a local copy of the `TResult` type that used to be necessary since the `TResult` type isn't exported by the definition. 

Thankfully, Typescript now has the tools to pull this out:

```
type TResult = ReturnType<typeof t>;
```

Thanks!

PR #1267, opened 2019-06-16
ResourceKey type should allow top level string value

Kamahl19:
`ResourceKey` can also be string like in this case

```
{
  en: {
    "logout": "Logout"
  }
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -638,10 +638,12 @@ declare namespace i18next {
   interface ResourceLanguage {
     [namespace: string]: ResourceKey;
   }
-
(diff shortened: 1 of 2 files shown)


rosskevin:
Thanks @Kamahl19, please add a test case to https://github.com/i18next/i18next/blob/master/test/typescript/init.test.ts to prevent regressions.

Kamahl19:
@rosskevin done

rosskevin:
@jamuhl this is a patch release

jamuhl:
published in i18next@17.0.4

Issue #1271, opened 2019-06-24
Typescript import doesn't work


polRk:
```
import i18next from 'i18next'

import { enLocales } from './locales/en'
import { ruLocales } from './locales/ru'

console.log(i18next) // undefined

i18next.init({
  debug: true,
  resources: { en: enLocales, ru: ruLocales }
})

```

```
i18next_1.default.init({
                  ^

TypeError: Cannot read property 'init' of undefined
```

tsconfig
```

{
  "compilerOptions": {
    "lib": [
      "esnext",
      "dom"
    ],
    "module": "commonjs",
    "noImplicitReturns": true,
    "noUnusedLocals": true,
    "outDir": "lib",
    "sourceMap": true,
    "strict": true,
    "target": "es2017"
  },
  "compileOnSave": true,
  "include": [
    "src"
  ]
}

```
I can not use esModuleInterop, because many other libraries fall back

_Originally posted by @polRk in https://github.com/i18next/i18next/issues/1177#issuecomment-504989651_

polRk:
```
import i18next = require('i18next');

console.log(i18next.default) // undefined
```

polRk:
```
import * as i18next from 'i18next'

console.log(i18next.default) // undefined
```

rosskevin:
I believe this is because the original types imported are hidden inside a namespace. The namespace should be removed. PR is welcome.

polRk:
I think you should correct this mistake

rosskevin:
@polRk I think you should have some consideration for open source maintainers and learn that you need to contribute fixes for your own problems.

I volunteer to help with this project because I use it. It seems you are using it and have a problem that you want fixed, therefore I am willing to guide you in solving it.

TL;DR I am not responsible for fixing problems you have with this library, you are, and I am willing to help you.

polRk:
Why i can import moment, date-fns, react, angular, vue and so on... ? )

rosskevin:
The difference is a stylistic choice from the original types authors. Namespaces still have their use but fairly quickly fell out of favor. It is not "wrong" but different. Fixing this is a matter of removing the namespace, altering tests, and altering any changes in downstream projects like react-i18next.

koloboid:
```
import i18default from "i18next";
const i18next: typeof i18default = require("i18next");
```

Simple workarround for those who cannot use esModuleInterop

johannesschobel:
[...]

PR #1281, opened 2019-07-15
[TS]Allow non synthetic default imports

PierreAndreis:
Based on https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

this should work for imports such as `const i18n = require('i18next')` and `import * as i18n from "i18n"`.

Fixes https://github.com/i18next/i18next/issues/1271

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -974,4 +974,4 @@ declare namespace i18next {
 }
 
 declare const i18next: i18next.i18n;
-export default i18next;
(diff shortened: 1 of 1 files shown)


PierreAndreis:
Since tests failed (good thing), I will dive deeper into making the types compatible with all types of import.

PR #1282, opened 2019-07-15
[TS] Allow non-es6 imports

PierreAndreis:
This PR solves https://github.com/i18next/i18next/issues/1271 by defaulting to commonjs imports.

Babel generates a runtime function that essentially does something like `const m = require('i18next'); return m.default || m`. 
`esModuleInterop` basically allows typescript to interpret as that. 

`i18next` uses commonjs exports and manually creates a default to also support es6 babel environments, so the types should reflect that.
https://github.com/i18next/i18next/blob/master/index.js#L4-L5

### Breaking changes 

- Any typescript code that did not use `esModuleInterop` and relied on default export will give an error.

- Due to the nature of commonjs imports being read only, when imported as `* as i18next` it is not possible to mutate any property of the class. 
hence `i18next.modules = {}` will give an error as being a readonly property. 

I think we should enforce not allowing `i18next.modules` to be mutable and promote the use of `i18next.use` instead.


PierreAndreis:
Before merging, I would like at least one of the original authors to review it:
@mxl @lenovouser @qqilihq @butchyyyy @tkow @GiedriusGrabauskas @deerawan

PierreAndreis:
https://github.com/i18next/i18next/blob/master/index.js#L5 
This syntax is not valid as equivalent to `import default` in typescript. 

Even if the type was right (as it was the tests), a typescript compiled code would fail on run time because you are trying to import es6 module from a commonjs export, as you can see on this codesandbox: https://codesandbox.io/s/node-typescript-h3gcd
```javascript
i18next.init({});
        ^
TypeError: Cannot read property 'init' of undefined
    at Object.<anonymous> (/sandbox/src/index.ts:4:9)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Module.m._compile (/sandbox/node_modules/ts-node/src/index.ts:439:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/sandbox/node_modules/ts-node/src/index.ts:442:12)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:829:12)
    at Object.<anonymous> (/sandbox/node_modules/ts-node/src/bin.ts:157:12)
```

This PR addresses that by only allowing commonjs imports in typescript.
You can read briefly more about it here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop

rosskevin:
Let me back up and state the requirements for an import change so I make sure we have the same expectations - being that it should work with *and* without `esModuleInterop`.  

Requirements:
1. `import i18next from 'i18next'` should work with `esModuleInterop: true`
2. `esModuleInterop: false` or unspecified should also work, albeit my guess is that requires the non-preferred ts specific import `import i18n = require('i18next')`
3. All of the interfaces should be exported accessible for reuse e.g. `TResult` and `TFunction`

With this PR as-is, I _think_ you may have covered 1 and 2 above, but broke 3.  This makes me think a few things:
- we need to add a minimal test to assert that types like `TResult` are accessible from the index via `import {TResult} from 'i18next'`
- we need to remove the namespace to allow ^^^
- we need to test _at least the bare minimum_ with an `esModuleInterop:false` config so we can put this to rest once and for all.

I'm open to discussion and other approaches, but I think I've got the requirements right (and do agree we need a change).

PierreAndreis:
@rosskevin I have added tests proving that all three requirements are met. Let me know if there are any other changes needed!

rosskevin:
My goal is to get rid of the namespace to make interop vs non-interop as similar as possible in one single breaking change.  The namespace has been a point of confusion and irritation for many.

Thank you for your work here, I am going to try some experiments to further that goal.  I'll hold on this for the moment to see how my experiments work out, then circle back with you.

rosskevin:
I performed a few small changes here - leaving it mostly unchanged. I'm now going to leave this as-is and start another branch from here to attempt the namespace removal - just in case I run into issues.

PierreAndreis:
@rosskevin sounds good! let me know your progress as I would love to contribute to that as well.

rosskevin:
Closing in favor of #1286 which includes this work and enhances it with removal of namespace, standardization of imports.

@PierreAndreis please review #1286

PR #1286, opened 2019-07-18
[breaking change] Standardize imports between default and esModuleInterop

rosskevin:
## Changes
- Remove typescript namespace and use declare module
- Test with and without `esModuleInterop`
- Update test imports
- Provide named i18next export from the runtime, and standardize/limit typescript use to named export only to prevent errors in userland due to different typescript configurations

## Rationale
Using a namespace in the typescript file is more unconventional than not, and led to quite a bit of confusion.  To access types such as TFunction, one typically had to:

```tsx

i18next.init()

// type hidden inside namespace
i18next.TFunction
```

Further, esModuleInterop and default tsconfig behaviors differed.  Thanks to @PierreAndreis for starting this PR in #1282, I have followed on and removed namespaces and been able to obtain standard import behavior for both default and esModuleInterop modes.

The above example is now (regardless of tsconfig settings):

```tsx
import {i18next, TFunction} from 'i18next'

i18next.init()
```

## Breaking change
Related projects and user projects will need break - but in a worthwhile way will need to use more standardized import patterns shown above.  This includes react-i18next etc.

Solves #1271


test/typescript/exposed.test.ts
```diff
@@ -0,0 +1,14 @@
+/** Exercise exposed types/imports for different tsconfig esmoduleinterop settings */
+
+/* esModuleInterop: false (default), allowSyntheticDefaultImports: false (default) */
+import i18next, { Interpolator, Modules, WithT } from 'i18next';
+
+i18next.init();
```
PierreAndreis: In a commonjs enviroment (node.js where `esModuleInterop` is not enabled), this will give a runtime error because the entry point doesn't provide a `module.exports.default` 

As you can check by running this codesandbox: https://codesandbox.io/s/node-typescript-h3gcd
``i18next.init({});
        ^
TypeError: Cannot read property 'init' of undefined
    at Object.<anonymous> (/sandbox/src/index.ts:4:9)```

Typescript does not use the index.js in the root of the directory. it uses the published `dist/cjs/i18next.js` which has this as export
```var i18next = new I18n();

module.exports = i18next;```

Comment (rosskevin) [reply test/typescript/exposed.test.ts]: tsc will use whatever file is pointed to in the main of the package.json, which I now see is different than I expected:

https://unpkg.com/i18next@17.0.6/package.json

I'll do a local build and see if I can correct it.

Comment (PierreAndreis) [reply test/typescript/exposed.test.ts]: Makes sense. Thank you for taking a look on this.

Comment (rosskevin) [reply test/typescript/exposed.test.ts]: the esm only has

```
var i18next = new I18n();
export default i18next;
```

whereas the cjs only has

```
var i18next = new I18n();
module.exports = i18next;
```

So named export is incorrect, and makes importing it differ depending the `esModuleInterop` setting again.  The _ideal_ solution to this is to *remove* the runtime **default** export and standardize on a named import only.  This would provide identical import uses no matter it being js, ts (any config permutation)

So, with that said, I have limited the ts export to named only, and added a named export to the index.js.

With that, the cjs generated is now:

```ts
var i18next = new I18n();

exports.i18next = i18next;
exports.default = i18next;
```

and esm:

```ts
var i18next = new I18n();

export default i18next;
export { i18next };
```

rosskevin:
@jamuhl please review the comments above.  I need your approval on the additional index.js change, which is really necessary to resolve this once and for all.

If you were so inclined, I propose removing the `default` export altogether.  It is not required but I have seen plenty of blog posts about it, and since we did so in our own libraries I have seen the benefits.

test/typescript/exposed.test.ts
rosskevin: TL;DR So with AND without the esModuleInterop, the import is always the same now: `import {i18next} from 'i18next'`.

rosskevin:
There is a rollup implication in cjs to both default and named, looking now:
https://github.com/rollup/rollup/issues/1961#issuecomment-423037881

rosskevin:
@jamuhl see the end of that comment ^^^ it gives some rationale and links to a blog post about banning default exports.

It seems the only right way to get this accomplished for all permutations of users both js and ts, both cjs and esm is to remove the default and switch entirely to named export. @jamuhl what are your thoughts?

jamuhl:
@rosskevin currently I would prefer not throwing another breaking change at JS users...

jamuhl:
Also current exports are correct:

```js
// cjs
var i18next = new I18n();
module.exports = i18next; // --> var i18next = require('i18next');

// es
var i18next = new I18n();
export default i18next; // --> import i18next from 'i18next';
```

https://github.com/i18next/i18next/commit/8e1fb28e12aecf0424d8705342183d44f784a5e4#diff-1fdf421c05c1140f6d71444ea2b27638 removed the main problem of mixing default and named exports

rosskevin:
@jamuhl the problem is that we need a named export to resolve this unequivocally for all users. 

Do you have a different solution?

jamuhl:
Look as long there is no change on importing for JS, node.js users - I'm ok with any change...but if we have to enforce named export as a breaking change for everyone - just to get typescript work in client bundler and node.js I'm a little concerned.

rosskevin:
@jamuhl it is not typescript users that will have a problem if there are both named and default exports.

I mention the *named* export because it - without a doubt - will work for everyone, in every situation.

I'm ok leaving the dual exports as-is and trying it, which will work for all users with a bundler as well as typescript, but as mentioned in the linked rollup issue:
> it definitely is a problem if you export to CJS and your consumer is not using a bundler

So, we can:
1. Solve it unequivocally by changing to a named import (breaking change for all)
2. Leave the dual imports and impact (not sure how big this segment would be) any user using cjs **without** any form of bundler
3. Leave the default export, and continue to get `imports don't work` issues from typescript users because their tsconfig `esModuleInterop` dictates `import * from` vs `import from`.

I would prefer option 1, but I agree it is not my choice.

What is your decision @jamuhl?

jamuhl:
So:

1) everyone gets punished with not running code when upgrading
2) "Only" node.js users get punished
3) "Only" ts users get punished

https://github.com/rollup/rollup/issues/1961#issuecomment-422960741 states it's bad practice to mix...I don't see were we mix default and named exports? https://github.com/i18next/i18next/pull/1286#issuecomment-513143792

jamuhl:
Sorry might be to warm today...but I really don't get it...

browser bundlers: ok
browser native: ok (<script type=module .../> beside babel-helper issue)
browser oldschool umd: ok
node.js cjs: ok
node.js mjs: ok
typescript: nok (why? can't it use default export on esm? dictates import * as i18n from 'i18next' what why...?)

rosskevin:
The confusion is easy on this one.

- To make it work for all tsconfig variations I limited the definition to named only and **added** a named export.
- I then saw the roll-up warning and researched it
- This PR still has mixed exports which is bad

Your pain/punishment list is correct :)

- TS cannot use the cjs runtime file with default export and esModuleInterop: false
- With esModuleInterop ts users can, but some users have objections to turning it on

PierreAndreis:
In this scenario, the approach taken by other framework (such as React), it was to force the typescript user to deal with cjs import even when using es syntax (import * as React from "react").

In reality they are exporting from cjs files which by its specification does not contain a default export or recognize this syntax.
At this time, `import * as` is a common syntax between typescript users when using tsc compiler instead of babel (mostly, node users)

In case the user prefers to keep similar with how babel (or other bundlers) do, they can use `esModuleInterop: true` to do similar of what babel does at runtime.

I definitely agree that it would be preferable that it was a universal export, but it doesn't really make sense to punish other users with not running code.

jamuhl:
So option 5) is the one to take? Sorry, you both lost me on the way...but 5) sounds best to me...

rosskevin:
Without the breaking change to switch to named only, the only option is to make the ts users differentiate.  I'll put some work on it this week.  The next couple of days are packed for me.

PierreAndreis:
@rosskevin I can put some work but would need a bit more details in what approach are we taking.
Will ts users without esModuleInterop use import as
1) 
```javascript
import * as i18next from "i18next";
i18next.init({});
const modules: i18next.Modules = {}
```
2) 
```javascript
import { i18next, Modules } from "i18next";
i18next.init({});
const modules: Modules = {}
```

rosskevin:
I want to be rid of the namespace for sure,.  I would prefer all looked like 2, but I think we may have to fall back to 1

PierreAndreis:
@rosskevin is the namespace under the module ok? I can't think other way to export interfaces together with `export =` without it being under a namespace. Let me know if you have any ideas!

rosskevin:
Ok, I'm open.   Essentially `esModuleInterop` opt-in should make the imports super easy, and otherwise if they decide they are certain not to opt-in, they have to do what is necessary to get their imports straight.

With that said, I want to invert the tests.  Default the tests to `esModuleInterop` (our default and that of most users) and switch the one-off test for compatibility to the `esModuleInterop: false`.

PierreAndreis:
@rosskevin I have spent a few hours on this and I wasn't able to make both cases working without ending up with the same typings like on my previous PR #1282 
```typescript
declare namespace i18next {
 // ...
}
declare const i18next: i18next.i18n;
export = i18next;
```

This works for both case in my previous reply. Please advice an alternative that doesn't use namespace, as it seems a requirement to get this completed.

rosskevin:
That's what I figured based on my previous reply.  I'm sorry I was not clear.
- keep namespace
- make most tests running with esModuleInterop:true (not the other way around)
- make the one-off test esModuleInterop:false
- make the esModuleInterop:false use the `import * as` syntax
- make the esModuleInterop:true use the `import i18n, {TResult` syntax

I think this is the best result we can get for the majority of users without the named export.

PierreAndreis:
I opened a PR with the requested changes: https://github.com/i18next/i18next/pull/1291

rosskevin:
Closing in favor of #1291

PR #1288, opened 2019-07-19
Disabled default returnedObjectHandler

acc15:
To allow returning of a key when `returnObjects` is `false`. See #1234

More discussion in #1239

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -32,7 +32,7 @@ export function get() {
     returnEmptyString: true, // allows empty string value as valid translation
     returnObjects: false,
     joinArrays: false, // or string to join array
-    returnedObjectHandler: () => {}, // function(key, value, options) triggered if key returns object but returnObjects is set to false
+    returnedObjectHandler: false, // function(key, value, options) triggered if key returns object but returnObjects is set to false
     parseMissingKeyHandler: false, // function(key) parsed a key that was not found in t() before returning
     appendNamespaceToMissingKey: false,
     appendNamespaceToCIMode: false,


jamuhl:
will be published in an upcoming version

jamuhl:
published in i18next@17.0.7

src/defaults.js
```diff
@@ -32,7 +32,7 @@ export function get() {
     returnEmptyString: true, // allows empty string value as valid translation
     returnObjects: false,
     joinArrays: false, // or string to join array
-    returnedObjectHandler: () => {}, // function(key, value, options) triggered if key returns object but returnObjects is set to false
+    returnedObjectHandler: false, // function(key, value, options) triggered if key returns object but returnObjects is set to false
```
vitkarpov: @jamuhl If I get it right, somebody could expect a function here, couldn't they? If it's the case, does it break top-level API? 🤔

jamuhl: what are you talking about...

https://github.com/i18next/i18next/blob/master/src/Translator.js#L122

it just fixed the usage -> there should be a string returned to show something is wrong 

before having a noop function it just did nothing...

vitkarpov: Yeah, I see that it's kind of default options and not used from the outside of the lib. Btw, it was just a question, not a passive-aggressive one 😄

vitkarpov: Gosh, now I'm super interested in what's actually been broken since a few recent versions 😃

jamuhl: fun fact is v1 -> v17 is more or less the same exposed API 

we still run v1.x.x tests with a compatibility mode (which was part of i18next until around v10): https://github.com/i18next/i18next/tree/master/test/backward

jamuhl: Not taking it as aggressive - no worry - just as I try my best to not add to the "javascript-fatigue" by keeping the API as stable as possible by not breaking things. I got a little sad by the tweet.

vitkarpov: > I got a little sad by the tweet.

Yeah, I'm sorry about that 😞
I think it'd be cool to see your answer to the initial tweet (what's actually broken and what's not) 👍

PR #1297, opened 2019-08-01
Expose more classes in rollup bundle for downstream testing

rosskevin:
Needed for https://github.com/i18next/i18next-xhr-backend/pull/319

Updating the i18next dependency there revealed test dependence on old non-rollup dist structure.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,8 @@
 import i18next from './i18next.js';
 
 export default i18next;
+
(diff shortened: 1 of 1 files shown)


jamuhl:
I see...didn't yet update those modules to latest i18next :(

after not having the individual files exposed the way to reach those `backendConnector`, ... is to do:

```js
i18next.init();
const backendConnector = i18next.services.backendConnector;
// ...
```

Currently on my way home from holidays...will look into merging your PR's and adapting the tests to latest i18next tomorrow....

PR #1298, opened 2019-08-03
refactor: Add initializedLanguageOnce bool to types

isaachinman:
@rosskevin @jamuhl 

Closes https://github.com/i18next/react-i18next/issues/913

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -969,6 +969,11 @@ declare namespace i18next {
     isInitialized: boolean;
 
     /**
+     * Used in React SSR
+     */
+    initializedLanguageOnce: boolean;
+
+    /**
      * Emit event
      */
     emit(eventName: string): void;


rosskevin:
This _could_ be a type augmentation in react-i18next.  I checked the code here and there is no reference to it, it is a property dynamically added in react-i18next only.  

@isaachinman I don't have a firm opposition to add it here, but it might be nice to keep this just in that repo.  https://www.typescriptlang.org/docs/handbook/declaration-merging.html

Do you want to try this as a PR (with a test in `test/typescript` there?

I'm guessing the equivalent that should go there is:

```
declare module i18next {
declare namespace i18next {
interface i18n {
    /**
     * Used in React SSR
     */
    initializedLanguageOnce: boolean;
}}}
```

If this doesn't pan out there for some complication of declaration merging, we can just add it here.  I think the above is the right way to do it, and might set the right pattern to keep those things separate/additive.

isaachinman:
What does @jamuhl think? Indeed I wasn't sure where this PR should land. Mutating the i18n object inside `react-i18next` seems a bit weird to me - my initial preference would be to keep that bool at the highest level if it is going to exist in the i18n ecosystem at all.

rosskevin:
> Mutating the i18n object inside react-i18next seems a bit weird to me

But this is exactly what happens in the runtime. Indeed there are zero occurrences of such property in this project leading me to believe it should not be added here.

jamuhl:
https://github.com/i18next/react-i18next/blob/master/src/useSSR.js#L16
https://github.com/i18next/react-i18next/blob/master/src/useSSR.js#L21

Those are internal flags - but guess there is some use case next-i18next users requested that on the public types...

I would add both values if needed in typescript...

isaachinman:
@jamuhl Which repo does this PR belong in?

jamuhl:
@isaachinman @rosskevin would say into react-i18next if i18next types can be extended there...?!? but honestly i don't know...guess i18next types come from here so they need to be added here...🤷‍♂️

rosskevin:
@isaachinman I am the maintainer of the types and I have already answered your question and provided the justification. You are free to challenge my thinking.

PR #1301, opened 2019-08-05
use() should accept a class or object

### LINKED ISSUE #1300: .use() should accept class or object
Discovered in https://github.com/i18next/i18next-xhr-backend/pull/322

Myself and others thought the signature was `.use(object)` which is an incorrect representation of the runtime.

rosskevin:
Closes #1300

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -755,6 +755,9 @@ declare namespace i18next {
     external: ThirdPartyModule[];
   }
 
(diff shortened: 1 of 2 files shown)


rosskevin:

Once released, need to trigger a rebuild of https://github.com/i18next/i18next-xhr-backend/pull/322

jamuhl:
published in i18next@17.0.9

PR #1308, opened 2019-08-09
Allow augmentation of TFunction to be able to add overloads

mgroenhoff:
Required for the types for i18next-sprintf-postprocessor at DT to be updated to make use of the bundled i18next types.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -623,14 +623,16 @@ declare namespace i18next {
     t: TFunction;
   }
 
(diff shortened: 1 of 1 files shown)


adrai:
@rosskevin is this ok for you?

rosskevin:
Looks good and tests still pass!

jamuhl:
published in i18next@17.0.11

PR #1310, opened 2019-08-13
Cope with null interpolation

alexrudall:
Cope with null value as well as undefined :)

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -55,9 +55,7 @@ class Interpolator {
     const regexpStr = `${this.prefix}(.+?)${this.suffix}`;
     this.regexp = new RegExp(regexpStr, 'g');
 
-    const regexpUnescapeStr = `${this.prefix}${this.unescapePrefix}(.+?)${this.unescapeSuffix}${
-      this.suffix
-    }`;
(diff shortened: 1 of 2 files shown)


jamuhl:
if doing so, please also do it here: https://github.com/i18next/i18next/pull/1310/files#diff-c4c32a53d5f4662890bf17b0a8fd47bcL117

alexrudall:
Hmm if I add this `test/interpolation.spec.js` line 426 fails, seems like it copes with it already? Sorry if I'm being dumb!

jamuhl:
so this can be closed?

alexrudall:
Sorry, I meant I don't think it's necessary to add the null check here as you suggested:

https://github.com/i18next/i18next/pull/1310/files#diff-c4c32a53d5f4662890bf17b0a8fd47bcL117

as it seems to be already covered by existing specs, which actually fail if I add the null check you suggested.

But I could be wrong! Apologies, I'm quite new to this library.

jamuhl:
already late here - will merge it tomorrow

jamuhl:
published in i18next@17.0.12

alexrudall:
Thanks!

PR #1313, opened 2019-08-16
remove unused Interpolator#init reset param (only used internally)

xgbuils:
Hi!

I started to use this library and I would like to collaborate. I removed Interpolator#init `reset` param because it's just used internally in the constructor.

Cheers!

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -5,15 +5,13 @@ class Interpolator {
   constructor(options = {}) {
     this.logger = baseLogger.create('interpolator');
 
-    this.init(options, true);
+    this.options = options;
+    this.format = (options.interpolation && options.interpolation.format) || (value => value);
+    this.init(options);
   }
 
   /* eslint no-param-reassign: 0 */
-  init(options = {}, reset) {
-    if (reset) {
-      this.options = options;
-      this.format = (options.interpolation && options.interpolation.format) || (value => value);
-    }
+  init(options = {}) {
     if (!options.interpolation) options.interpolation = { escapeValue: true };
 
     const iOpts = options.interpolation;


jamuhl:
It's used here: https://github.com/i18next/i18next/blob/master/src/Translator.js#L284 so this change would be breaking things...

xgbuils:
I'm talking about `reset` param and not `reset` method. Could you review my PR it please?

jamuhl:
I see....might be merged for some future release - but honestly, this does not fix a thing nor save a lot in size...so...

xgbuils:
Thanks @jamuhl! I'm not trying to fix something. I'm just trying to apply some coding principles. In this case:

- Not exposing to the clients more than they need.
- Keeping code simple: I removed an unnecessary conditional. Now it's easy to understand that `this.options` is just set when an Interpolator is created.

The idea of this PR is having a more mantainable code. I would like to contribute on more PRs like this if you think is valuable.

Cheers!

jamuhl:
ok...just be aware that all plugins do not set options on construction but on init (and out of a good reason).

xgbuils:
I found some backend plugins that use interpolator object. However they are just using `services.interpolator.interpolate` method. Then, no problem.

- [i18next-xhr-backend](https://github.com/i18next/i18next-xhr-backend/blob/master/src/index.js#L33)
- [i18next-fetch-backend](https://github.com/perrin4869/i18next-fetch-backend/blob/master/src/index.js#L50)
- [i18next-fluent-backend](https://github.com/i18next/i18next-fluent-backend/blob/master/src/index.js#L34)
- [i18next-node-fs-backend](https://github.com/i18next/i18next-node-fs-backend/blob/master/src/index.js#L127)
- [i18next-sync-fs-backend](https://github.com/sallar/i18next-sync-fs-backend/blob/master/src/index.js#L120)
- [i18next-node-remote-backend](https://github.com/i18next/i18next-node-remote-backend/blob/master/src/index.js#L49)
- [i18next-firebase-backend](https://github.com/inteligator/i18next-firebase-backend/blob/master/src/index.js#L27)
- [i18next-smart-bucket-backend](https://github.com/nekuz0r/i18next-smart-bucket-backend/blob/master/src/backend.js#L25)
- [i18next-service-backend](https://github.com/timbrandin/i18next-service-backend/blob/master/src/index.js#L57)
- [i18next-node-service-backend](https://github.com/timbrandin/i18next-node-service-backend/blob/master/src/index.js#L91)

jamuhl:
I meant those plugins do a constructor call without options and then an init call with options...

xgbuils:
> I meant those plugins do a constructor call without options and then an init call with options...

Sorry, at initial point I undestood that I can break some plugin if I remove `reset` param. However, after re-reading your comments, I don't understand the analogy between plugins and Inperpolator class and what's the good reason to keep setting options on `init` (in interpolator case).

jamuhl:
Was not in relation to Interpolation it was for:

> The idea of this PR is having a more mantainable code. I would like to contribute on more PRs like this if you think is valuable.

A hint that you can't make the same changes on plugins...in cases you plan to do so

jamuhl:
published in i18next@17.0.12

PR #1317, opened 2019-09-05
Allow TFunction to receive 1-3 parameters

tobiasweibel:
Example: `i18n.t('MyKey', 'My value {{value}} with interpolation', {value: 3})`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -623,18 +623,22 @@ declare namespace i18next {
     t: TFunction;
   }
 
+  type TFunctionResult = string | object | Array<string | object> | undefined | null;
+  type TFunctionKeys = string | TemplateStringsArray;
   interface TFunction {
+    // basic usage
     <
-      TResult extends string | object | Array<string | object> | undefined | null = string,
-      TKeys extends string | TemplateStringsArray = string,
+      TResult extends TFunctionResult = string,
+      TKeys extends TFunctionKeys = string,
       TInterpolationMap extends object = StringMap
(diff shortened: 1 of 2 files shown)


tobiasweibel:
I've created the PR from the web UI. I'm going to clone this repository now and see why the linter fails.

rosskevin:
@tobiasweibel any change to the types will need to be accompanied by typescript usage tests to prevent regressions. Please see the pretest script in package.json which indicates how these tests are run.

index.d.ts
```diff
@@ -632,6 +632,15 @@ declare namespace i18next {
       key: TKeys | TKeys[],
       options?: TOptions<TInterpolationMap> | string,
     ): TResult;
+    <
+      TResult extends string | object | Array<string | object> | undefined | null = string,
+      TKeys extends string | TemplateStringsArray = string,
+      TInterpolationMap extends object = StringMap
+    >(
```
tobiasweibel: @rosskevin I'm not really experienced enough in TypeScript so that I have to define the generics in TFunction only once. Do you know how to do that?

rosskevin: I was able to get a little less cruft/more reuse, but was unable to declare types at the interface level.  It's a bit better but very much the same as you authored.

rosskevin:
Thanks @tobiasweibel - that was a bit trickier than usual.

jamuhl:
published in i18next@17.0.14

PR #1322, opened 2019-09-12
Incorrect callback type for backend module read callback

skovy:
The `ReadCallback` is the same for both `read` and `readMulti` but according to the docs they expected two different values: https://www.i18next.com/misc/creating-own-plugins#backend

Code from the docs:

```javascript
  read: function(language, namespace, callback) {
    /* return resources */
    callback(null, {
      key: 'value'
    });
  },

  // optional
  readMulti: function(languages, namespaces, callback) {
    /* return multiple resources - usefull eg. for bundling loading in one xhr request */
    callback(null, {
      en: {
        translations: {
          key: 'value'
        }
      },
      de: {
        translations: {
          key: 'value'
        }
      }
    });
  },
```

Before this change, it would error: `Type 'string' is not assignable to type 'ResourceLanguage'`.

In the following code: 

```typescript
const backendModule: BackendModule = {
  type: 'backend',
  init: () => null,
  read: (_language, _namespace, callback) => {
    callback(null, {
      key: 'value', // <-- Type 'string' is not assignable to type 'ResourceLanguage'
    });
  },
  create: () => null,
  readMulti: (_languages, _namespaces, callback) => {
    callback(null, {
      en: {
        namespace: {
          key: 'value',
        },
      },
    });
  },
  save: () => null,
};
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -684,7 +684,9 @@ declare namespace i18next {
     type: 'backend' | 'logger' | 'languageDetector' | 'postProcessor' | 'i18nFormat' | '3rdParty';
   }
 
-  type ReadCallback = (err: Error | null | undefined, data: Resource) => void;
+  type CallbackError = Error | null | undefined;
+  type ReadCallback = (err: CallbackError, data: ResourceKey) => void;
+  type MultiReadCallback = (err: CallbackError, data: Resource) => void;
 
   /**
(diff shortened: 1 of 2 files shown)


rosskevin:
Thanks @skovy 

jamuhl:
published in i18next@17.0.16

Issue #1324, opened 2019-09-13
default value not working in IE


Dgiulian:
I'm using react-i18n to translate and trying to use the default value when a key is not defined as indicated [here](https://www.i18next.com/translation-function/essentials#passing-a-default-value).

So I have:
```
const [t] = useTranslation('TranslationResources');
...
const myLabel = t('My_label_key', 'default text')
```

This is working correctly in chrome, but in IE11 it fails with the following error:
```
Uncaught (in promise) TypeError: Object.keys: argument is not an Object
  at _objectSpread (http://localhost:3000/static/js/8.chunk.js:462:5)
  at fixedT (http://localhost:3000/static/js/8.chunk.js:112273:9)
...
```
apparently the translate function expect an object as a second parameter, and if it's a string it creates one with the **defaultValue** key. 

I solved it using this: 
```
const myLabel = t('My_label_key', {defaultValue: 'default text'} )
```
However, the documentation doesn't mention anything about it.

jamuhl:
PR was merged - closing this now

PR #1325, opened 2019-09-13
Fixed default value as string in IE

Dgiulian:
This is a possible fix for #1324


jamuhl:
Will be merged and published this weekend...

jamuhl:
published in i18next@17.0.15

PR #1326, opened 2019-09-14
add getDataByLanguage into index.d.ts

### LINKED ISSUE #1273: Is getDataByLanguage a public api?
Changelog mentions it under (11.6.0) 
https://github.com/i18next/i18next/blob/master/CHANGELOG.md#1160
but I don't see it in documentation or typescript definition file.

msftjayongg:
update index.d.ts with the public getDataByLanguage API interface.  Basically this allows typescript to call the API that already exists. No functionality change.

Closes #1273

### DIFF
--- a/test/typescript/init.test.ts
+++ b/test/typescript/init.test.ts
@@ -205,6 +205,12 @@ const v: string = i18next.t('my.key');
 const de = i18next.getFixedT('de');
 const z: string = de('myKey');
 
+// verify we can get the German data
+const data = i18next.getDataByLanguage('de');
+if (data.translation.myKey !== de('myKey')) {
+  console.log('something went wrong calling getDataByLanguage');
+}
+
 // or fix the namespace to anotherNamespace
 const anotherNamespace = i18next.getFixedT(null, 'anotherNamespace');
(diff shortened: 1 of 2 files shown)


msftjayongg:
@rosskevin I was told to tag you.

I am also unsure how to actually run the typescript tests - I guess they are there just for tslint purposes? (I'd feel better if I could run it :) )

test/typescript/init.test.ts
```diff
@@ -205,6 +205,12 @@ const v: string = i18next.t('my.key');
 const de = i18next.getFixedT('de');
 const z: string = de('myKey');
 
+// verify we can get the German data
+const data = i18next.getDataByLanguage('de');
+if (data.translation.myKey !== de('myKey')) {
```
rosskevin: This is type usage test only, so we only need the line 209, please remove the `if` block to keep these tests as brief as possible.

msftjayongg:
Updated - I left a line such that we can check that the .translation field exists.

test/typescript/init.test.ts
msftjayongg: updated with a change - please look :)

rosskevin:
Thanks @msftjayongg 

jamuhl:
published in i18next@17.0.16

PR #1329, opened 2019-09-24
add i18next-ecosystem image

sinchang:


jamuhl:
thank you....wasn't aware we broke that...should have a look at the readme on github more often

PR #1333, opened 2019-10-02
Improve performance of EventEmitter.off

strassl:
The loop in `EventEmitter.off` is is unnecessarily expensive.
Looping over all observers and calling `indexOf` each time results in (roughly) quadratic time complexity, leading to slowdowns when lots of listeners are registered.

This solution might still not be ideal due to the `Array.slice` call and we might be better off just using `Array.filter` (it certainly would be cleaner).
Unfortunately I don't know enough about the backwards-compatibility requirements off this project to make this decision.

### DIFF
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -17,7 +17,7 @@ class EventEmitter {
     }
 
     const observers = this.observers[event];
-    for (const i = observers.length - 1; i >= 0; i--) {
+    for (let i = observers.length - 1; i >= 0; i--) {
       const observer = observers[i];
       if (observer === listener) {
         observers.splice(i, 1);
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -13,7 +13,10 @@ class EventEmitter {
 
   off(event, listener) {
     if (!this.observers[event]) return;
-    if (!listener) return delete this.observers[event];
+    if (!listener) {
+      delete this.observers[event];
+      return;
+    }
 
     this.observers[event] = this.observers[event].filter(l => l !== listener);
   }
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -12,17 +12,10 @@ class EventEmitter {
   }
 
   off(event, listener) {
-    if (!this.observers[event]) {
-      return;
-    }
+    if (!this.observers[event]) return;
+    if (!listener) return delete this.observers[event];
 
-    const observers = this.observers[event];
-    for (let i = observers.length - 1; i >= 0; i--) {
-      const observer = observers[i];
-      if (observer === listener) {
-        observers.splice(i, 1);
-      }
-    }
+    this.observers[event] = this.observers[event].filter(l => l !== listener);
   }
 
   emit(event, ...args) {
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -16,16 +16,13 @@ class EventEmitter {
       return;
     }
 
-    this.observers[event].forEach(() => {
-      if (!listener) {
-        delete this.observers[event];
-      } else {
-        const index = this.observers[event].indexOf(listener);
-        if (index > -1) {
-          this.observers[event].splice(index, 1);
-        }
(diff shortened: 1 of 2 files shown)


jamuhl:
Agree...guess it could be:

```js
off(event, listener) {
   if (!this.observers[event]) return;
   if (!listener) return delete this.observers[event];

   this.observers[event].filter(l => l !== listener);
}
```

jamuhl:
published in i18next@17.0.18

Saibamen:
Some benchmark before/after?

strassl:
The benchmark is a nothing fancy, but I think the results speak for themselves. Tested on my MBP A1707 in Chrome 77.0.3865.90.

Output:
```
Running benchmark for 1000 listeners with 30 samples
Old: 243.04800000015044s
New: 4.378499999802443s
```


In case it is of interest to anyone - I ran into this performance problem because I was using react-i18next for individual cells in react-table. Since react-i18next registers a listener for each component.
When unmounting the table, all cells also have to be unmounted which causes the browser to grind to a sudden halt because all the `EventEmitter.off` calls can take multiple minutes for a few hundred components.


Benchmark code:
```js
class EventEmitter1 {
  constructor() {
    this.observers = {};
  }

  on(events, listener) {
    events.split(' ').forEach(event => {
      this.observers[event] = this.observers[event] || [];
      this.observers[event].push(listener);
    });
    return this;
  }

  off(event, listener) {
    if (!this.observers[event]) {
      return;
    }

    this.observers[event].forEach(() => {
      if (!listener) {
        delete this.observers[event];
      } else {
        const index = this.observers[event].indexOf(listener);
        if (index > -1) {
          this.observers[event].splice(index, 1);
        }
      }
    });
  }

  emit(event, ...args) {
    if (this.observers[event]) {
      const cloned = [].concat(this.observers[event]);
      cloned.forEach(observer => {
        observer(...args);
      });
    }

    if (this.observers['*']) {
      const cloned = [].concat(this.observers['*']);
      cloned.forEach(observer => {
        observer.apply(observer, [event, ...args]);
      });
    }
  }
}

class EventEmitter2 {
  constructor() {
    this.observers = {};
  }

  on(events, listener) {
    events.split(' ').forEach(event => {
      this.observers[event] = this.observers[event] || [];
      this.observers[event].push(listener);
    });
    return this;
  }

  off(event, listener) {
    if (!this.observers[event]) return;
    if (!listener) {
      delete this.observers[event];
      return;
    }

    this.observers[event] = this.observers[event].filter(l => l !== listener);
  }

  emit(event, ...args) {
    if (this.observers[event]) {
      const cloned = [].concat(this.observers[event]);
      cloned.forEach(observer => {
        observer(...args);
      });
    }

    if (this.observers['*']) {
      const cloned = [].concat(this.observers['*']);
      cloned.forEach(observer => {
        observer.apply(observer, [event, ...args]);
      });
    }
  }
}

let em1 = new EventEmitter1();
let em2 = new EventEmitter2();


function bench(count, em) {
    let listeners = [...Array(count).keys()]
        .map(it => (() => {}))

    listeners.forEach(l => em.on('event', l));

    let start = performance.now();

    listeners.forEach(l => em.off('event', l));

    let end = performance.now();

    return end - start;
}

function sample(f, samples) {
    sum = 0;
    for (let i = 0; i < samples; i++) {
        sum += f();
    }

    return sum / samples;
}

let listenerCount = 1000;
let samples = 30;

console.log(`Running benchmark for ${listenerCount} listeners with ${samples} samples`);
let time1 = sample(() => bench(listenerCount, em1), samples);
let time2 = sample(() => bench(listenerCount, em2), samples);

console.log(`Old: ${time1}s`);
console.log(`New: ${time2}s`);
```

jamuhl:
@strassl was my guess the issue is related to react-i18next -> as that is currently the only place i could think of having bound / unbound listeners frequently...

But..wrapping table cells with `withTranslation` or using `useTranslation` inside is not the best idea -> personally I would suggest doing that on the outer component and just pass `t` down as props...still a gain in performance

PR #1339, opened 2019-10-11
Support using ImmutableJS Records as the data model.

stami:
Records don't support `...` spread operator, so the previous `combinedData` implementation broke usage of Records as the localization parameters.

This also changes the test case of overriding a default value with undefined as it would be difficult (and not very valuable)
to implement that with the current `getLastOfPath`.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -70,20 +70,21 @@ class Interpolator {
     const defaultData =
       (this.options && this.options.interpolation && this.options.interpolation.defaultVariables) ||
       {};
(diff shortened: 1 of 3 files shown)


jamuhl:
why not just call `immutable.toJS()` ?

stami:
That would require to go trough all of the codebase and check every translation call. We considered to wrap our usage of the module, and do the `toJS()` there if necessary, but thought that it would be nicer to have the support in this core.

jamuhl:
published in i18next@17.2.0

jamuhl:

Issue #1340, opened 2019-10-14
postprocessor: which language has been hit?


srischard:
hi all,

background:
I'm implementing a `postprocessor` to format my messages and I want to cache my compiled expressions

setup:
I got a set of languages and a fallback chain.
In the postprocessor `process: (value, key, options, translator)` I got a value.
In `translator` I can see the original requested `language`.

But where I can see from which language and namespace the value has been taken?

jamuhl:
https://github.com/i18next/i18next/blob/master/src/Translator.js#L100

resolved has https://github.com/i18next/i18next/blob/master/src/Translator.js#L383 -> usedLng

but that won't get passed down to postprocessor

srischard:
it's is possible to pass this down to the postprocessor  by merging `usedLng` & `usedNS` in options
maybe in `extendTranslation`

```
  extendTranslation(res, key, options, resolved) {
    if (this.i18nFormat && this.i18nFormat.parse) {
      res = this.i18nFormat.parse(
        res,
        options,
        resolved.usedLng,
        resolved.usedNS,
        resolved.usedKey,
        { resolved },
      );
    }
   .....
      res = postProcessor.handle(postProcessorNames, res, key, options, this);
    }

    return res;
  }
```

jamuhl:
think simplest would be changing that call to:

`res = postProcessor.handle(postProcessorNames, res, key, { ...options, resolved }, this);`

that way we do not have to change the function params...

jamuhl:
PR would be welcome

PR #1341, opened 2019-10-15
pass used language & namespace to postprocessor

srischard:
for building a cache in the postprocessor I need the used language 

the PR passthrough the used language and used namespace in `options`

fixes https://github.com/i18next/i18next/issues/1340


jamuhl:
options contain a lot of things...from passed namespace to lngs to use instead of detected...so personally i would prefer it on options -> translator is more a component with fixed api

srischard:
so I moved the parameters to `options`.
But now a collision of option fields might be possible.

srischard:
the codeclimate issue is part of the original code and port of the changed code

jamuhl:
codeclimate is not that important...more important not passing down usedLng, usedNS but the complete resolved as is

srischard:
but passing the whole resolved properties will increase naming collisions for formatting the string.
passing the object at least in a single property `resolved` in `options` I'm only restricted to not use `resolved` im my wording templates.

jamuhl:
why not just rename resolved like I suggested: `{ i18nResolved: resolved, ...options }` 

this way one would have to name a interpolation variable i18nResolved...which is more than unlikely used...and even if used it would be overwritten if options contain it...

jamuhl:
thank you - published in i18next@17.3.0

PR #1344, opened 2019-10-23
Add missing `cleanCode` option to TypeScript def

MichielDeMey:
This PR adds the missing `cleanCode` option to the `initOptions` in the TypeScript definition.


jamuhl:
published in i18next@17.3.1

PR #1352, opened 2019-10-29
Typescript use `export default` for esm-first approach

### LINKED ISSUE #1351: Reconsider typescript export to cover both esModuleInterop modes and esm by default
We originally had `export default i18next` (which was still a namespace from previous @types) before https://github.com/i18next/i18next/pull/1291, then we switched to `export = i18next` with a namespace.  @PierreAndreis and I both agreed this to be the right solution.  In order to mitigate any fallout, we added tests for both `esModuleInterop` states, of which the default is `false`.

After #1291 change, we have had a larger number of typescript import issues (#1342, #1316, #1350, not to mention others in the i18next-react project), so I took a look at what it meant to get this running, thanks in part to @sraisty1c putting together a repo in https://github.com/i18next/react-i18next-gitbook/issues/63.

What I found and documented were the gory details of what is necessary to get a webpack+jest project both type checking and running.  It took extensive knowledge of the loading semantics and was quite painful.  In large part this is because nodejs uses our `cjs` (commonjs) package, and webpack uses our `esm` package.  Node/jest only loads cjs, whereas webpack loads `esm` then falls back to `cjs` in that order by default.

---

I spent four hours on this and documented my findings yesterday in https://github.com/rosskevin/test-i18next/tree/non-esmoduleinterop-ts-loader

---

More than not, even with the default `esModuleInterop: false`, it seems like the current scheme is too painful.  It is strange that typescript has not changed the default.   As stated in the [typescript 2.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop), they recommend running __**with**__  `--esModuleInterop` even though it is still not the default:

>Note: The new behavior is added under a flag to avoid unwarranted breaks to existing code bases. We highly recommend applying it both to new and existing projects. For existing projects, namespace imports (import * as express from "express"; express();) will need to be converted to default imports (import express from "express"; express();).

## Conclusion

I have always wanted to get rid of the namespace, but I recall @PierreAndreis and I having problems when attempting to do that https://github.com/i18next/i18next/pull/1286#pullrequestreview-263857803.  It seems that we opted there for a `cjs` first approach, and this has caused more pain then it relieved.

Quite simply, the index is an `export default`.  Yes we have both cjs and esm modules, but it seems that given most of our user base is using `esm` and it is the way forward, let's examine flipping it back to `export default` without the namespace and if there is mitigation to be required, do so for older `cjs` situations, not now and future `esm` situations.  

## TL;DR

This seems more difficult than it has to be.  `export default` is a modern `esm` semantic, let's support it first out of the box for everyone and require interop for cjs if we must (as recommended by [typescript 2.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop))

I'm going to try removing the namespace again given this knowledge, and see if I can move it forward and understand the implications to both cjs and esm users.

rosskevin:
## Rationale

Long history, explanation and rationale in issue #1351

## Moving forward

Use the following import styles regardless of the `esModuleInterop` setting:
- `import i18next from 'i18next'`
- `import {TFunction} from 'i18next'`
- `import i18next, {TFunction} from 'i18next'`

## Implications for `esModuleInterop: true` users

NO CHANGE at all!  Yay.

## Implications for `esModuleInterop: false` (default) users

```diff
-`import * as i18next from 'i18next'`
- const {TFunction} = i18next
+ `import i18next, {TFunction} from 'i18next'`
```

1. jest breaks - it is run with node and there is no esm loading in ts-jest ([they recommend running with `esModuleInterop`](https://github.com/kulshekhar/ts-jest/wiki/Troubleshooting#commonjs-compatibility)) or with `@std/esm` [open issue about jest support](https://github.com/standard-things/esm/issues/706)
2. node breaks - as noted above, you must use `esModuleInterop: true` to load the old `commonjs` or use an esm loader such as @std/esm

__These breakages are expected and intended.__  We want to move forward with an `esm`-first approach, and as indicated by `ts-jest` docs, it is expected that if you load commonjs code, you need to use `esModuleInterop: true`.  [Typescript recommends the use of and the intent of `--esModuleInterop` is exactly for the purpose of loading legacy commonjs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop).

Closes #1351


rosskevin:
@jamuhl this is a typescript-only breaking change.  It is your call on what to do with semver here.  I can argue for a major or minor bump.  If I were persuaded in one direction, on this case it would be major.

I will be following on with PRs for other i18next projects to deal with the ts interop tests/imports.

jamuhl:
@rosskevin prefer major too...give me a few minutes to publish this.

jamuhl:
published in i18next@19.0.0

PR #1359, opened 2019-11-19
ignore non valid keys in saveMissing

adrai:


jamuhl:
published in i18next@19.0.1

PR #1366, opened 2019-11-28
Type ResourceStore

### LINKED ISSUE #1364: Typescript: type of "i18next.services.resourceStore" is wrong
I'm trying to access the resource store datas for Server Side Rendering.

I use: `i18next.services.resourceStore`, witch is of type 'Resource'
I expect `i18next.services.resourceStore['en']` to work, but instead I have to do:
`i18next.services.resourceStore.data['en']`

That data property does not exist for Typescript. It looks like i18next.services.resourceStore is not of ResourceType, but i18next.services.resourceStore.data is. Maybe a 'ResourceStore' is missing in the middle?

rosskevin:
This is my interpretation of the runtime code and requires review for accuracy

Closes #1364

/cc @fromi

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -672,6 +672,12 @@ export interface Interpolator {
   nest(str: string, fc: (...args: any[]) => any, options: InterpolationOptions): string;
 }
 
+export class ResourceStore {
+  constructor(data: Resource, options: InitOptions);
+  public data: Resource;
+  public options: InitOptions;
+}
+
 export interface Services {
   backendConnector: any;
   i18nFormat: any;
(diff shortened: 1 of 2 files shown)


jamuhl:
https://github.com/i18next/i18next/issues/1364#issuecomment-559687028 not against this...but it's in the gray shade of private/publich API

rosskevin:
I'm good with this as long as it is factually representative, even though it may not be preferred access.

rosskevin:
This can be a patch release

jamuhl:
was published in i18next@19.0.2

PR #1121, opened 2018-10-22
Add Hebrew plural support

LeonardDrs:
Using this [doc](http://www.unicode.org/cldr/charts/33/supplemental/language_plural_rules.html#he) as reference.

Ongoing PR for the translate documentation: https://github.com/translate/l10n-guide/pull/71/

Fixes https://github.com/i18next/i18next/issues/1120

PR #1369, opened 2019-12-09
Fix Hebrew plural resolver

dutzi:
Source: I speak Hebrew.

### DIFF
--- a/src/PluralResolver.js
+++ b/src/PluralResolver.js
@@ -35,7 +35,7 @@ let sets = [
   { lngs: ['or'], nr: [2,1], fc: 2 },
   { lngs: ['ro'], nr: [1,2,20], fc: 20 },
   { lngs: ['sl'], nr: [5,1,2,3], fc: 21 },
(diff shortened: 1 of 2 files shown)


dutzi:
Yeah... I should've actually run the test before making this PR.

jamuhl:
Based on a previous PR https://github.com/i18next/i18next/pull/1121 and unicode/cldr http://www.unicode.org/cldr/charts/36/supplemental/language_plural_rules.html#he hebrew has multiple forms (not that I can judge myself - not speaking it)

@LeonardDrs can you please confirm current rule is correct?

LeonardDrs:
I don't speak hebrew myself, I just noticed that our translate tool wasn't working for hebrew as our plural sentences didn't match the one given by our translators.
Looked up the documentation and did this fix that worked for our translators :grimacing:

felixmosh:
If my suggestion in here https://github.com/i18next/i18next/issues/1374 will be approved, you will able to disable this multiple plurals feature per usage or globally.
@jamuhl WDYT?

Issue #1384, opened 2020-01-22
Interpolation formatting does not support default format(s)


mjbcopland:
The `format` argument of the formatter function is optional but must be supplied during interpolation in order for any custom formatting to take effect.

Consider the following example

```js
i18next.init({
  lng: "en",
  interpolation: {
    escapeValue: false,
    format: function(value, format, lng) {
      return value.toLocaleString(lng);
    },
  }
});

console.log(
  i18next.t("Total {{amount}} since {{date}}", {
    amount: 123456.789,
    date: new Date()
  })
);
```

```
Total 123456.789 since Wed Jan 22 2020 17:03:40 GMT+0000 (Greenwich Mean Time)
```

I don't see any reason why the interpolated values should not be formatted, yet that is the case. We can force formatting by supplying an empty format (include the separator but no value), but this is unintuitive, easily missed, and only discoverable by [inspecting the source](https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L79).

```js
console.log(
  i18next.t("Total {{amount,}} since {{date,}}", {
    amount: 123456.789,
    date: new Date()
  })
);

```

```
Total 123,456.789 since 1/22/2020, 17:03:40
```

jamuhl:
And why should I expect any formatting happening with this:

```js
i18next.t("Total {{amount}} since {{date}}", {
    amount: 123456.789,
    date: new Date()
  })
```

This says use the raw value (toString if needed)

```js
i18next.t("Total {{amount, number}} since {{date, dateShort}}", {
    amount: 123456.789,
    date: new Date()
  })
```

says to format it and can be changed by translators as needed.

mjbcopland:
I guess my question is why are we using the raw value and/or toString when a custom formatter exists? I feel like my toLocaleString example above is a reasonable enough use-case.

jamuhl:
You do not like to call functions if not needed to call them...just out of performance reason...

jamuhl:
Can merge your PR - but really do not agree in the behaviour is / should be default anytime in future

PR #1385, opened 2020-01-22
Add option to always format interpolated values

mjbcopland:
Addresses #1384 

Although there isn't a compelling reason to be dependent on the old behaviour, this is technically a breaking change so I've added a new option `alwaysFormat` (not sure if this should also update the docs) which will, unimaginatively, always format values used for interpolation. This option defaults to `false` in order to preserve existing behaviour.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -78,17 +78,16 @@ class Interpolator {
     }
 
     const handleFormat = key => {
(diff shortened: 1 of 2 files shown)


jamuhl:
Will be published this weekend....

jamuhl:
landed in i18next@19.1.0

PR #1398, opened 2020-02-18
Add ability to use commas in nested keys

### LINKED ISSUE #1395: Keys can not contain commas if they are nested
The following translation will not work because there is a comma in the key. It is assumed that the first comma is separating the key from the options.

```js
i18next.t('some key with a nested message {{message}}', {
    message: '$t(some other key that has a literal , in it and {{num}} options, { 'num': 1 })'
});
```
This will cause a lookup for the key "some other key that has a literal " instead of "some other key that has a literal , in it and {{num}} options". It would be nice if there was a way to escape the key or an option to change the separator character for the key and the options.

lgordey:
Fix #1395 issue

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -158,8 +158,9 @@ class Interpolator {
       if (key.indexOf(',') < 0) return key;
 
       const p = key.split(',');
-      key = p.shift();
-      let optionsString = p.join(',');
+
+      let optionsString = p.pop();
+      key = p.join(',');
       optionsString = this.interpolate(optionsString, clonedOptions);
       optionsString = optionsString.replace(/'/g, '"');
 
--- a/test/interpolation.spec.js
+++ b/test/interpolation.spec.js
@@ -336,6 +336,15 @@ describe('Interpolator', () => {
         ],
         expected: 'test success',
       },
+      {
+        args: [
+          '$t(test, is, {"key": "success"})',
+          function(key, opts) {
+            return 'test, is ' + opts.key;
+          },
+        ],
+        expected: 'test, is success',
+      },
     ];
 
     tests.forEach(test => {


jamuhl:
looks good...will check and merge/publish tomorrow.

👍thank you

jamuhl:
published in i18next@19.2.0

mgroenhoff:
What if there are no options?
```js
i18next.t('some key with a nested message {{message}}', {
    message: '$t(some other key that has a literal , in it)'
});
```

jamuhl:
@mgroenhoff good catch...but I guess nothing happens in that case: `...{{message}}` has no format on time of interpolation - but not tested (does it break?!?)

lgordey:
I guess It will go to `try/catch` block, in `catch` actually here
https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L172

mgroenhoff:
@jamuhl Since the part after the last comma is always "pop"ed off, it is not part of the key anymore so in the case of the key `"a, b, c, d"` the resulting key is `"a, b, c"`. Without some sort of escaping mechanism or an override option to the separation character, the only way i guess is to just try to `JSON.parse` the part after the last comma, but don't pop is off `p` until it succeeds

jamuhl:
@mgroenhoff sorry, was a little confused...what you say has nothing todo with this PR but with the `,` inside nestings (that's a completely different story). 

https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L157

In this case, an own issue would have helped not mixing things.

Guess a fix would be returning here https://github.com/i18next/i18next/blob/master/src/Interpolator.js#L172 the original key passed in?

mgroenhoff:
@jamuhl Well this PR claims to fix #1395 which is does partly but does not really solve the underlying problem.

I'm sorry if it wasn't clear enough but I was indeed talking about nesting keys. I meant if you would use those keys withing nesting. A full example would be:
```js
i18next.t('some key with a nested messages {{message1}}, {{message2}}', {
    message1: '$t(a, b, c, d)', // here 'key' now becomes "a, b, c" instead of the whole "a, b, c, d"
    message2: '$t(d, e, f, {{last}}, { 'last': 'g' })' // this now correctly works
});
```

jamuhl:
@mgroenhoff ha...was not even aware who provided the PR...not used to different people opening a issue and providing a PR...will reopen your issue.

PR #1419, opened 2020-04-03
Could we use formatting functions in the nesting 't' function?

brett-js:
For example:

```
{
    "key1": "The following text is uppercased: $t(key2, uppercase)",
    "key2": "some text"
}
```

```
i18next.t('key1');
// -> "The following text is uppercased: SOME TEXT"
```

If uppercase is defined as:

```
i18next.init({
    interpolation: {
        format: function(value, format, lng) {
            if (format === 'uppercase') return value.toUpperCase();
            return value;
        }
    }
});
```

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -185,6 +185,21 @@ class Interpolator {
 
     // regular escape on demand
     while ((match = this.nestingRegexp.exec(str))) {
+      let formatters = [];
+
+      /**
+       * If there is more than one parameter (contains the format separator). E.g.:
+       *   - t(a, b)
+       *   - t(a, b, c)
+       *
+       * And those parameters are not dynamic values (parameters do not include curly braces). E.g.:
+       *   - Not t(a, { "key": "{{variable}}" })
+       *   - Not t(a, b, {"keyA": "valueA", "keyB": "valueB"})
+       */
+      if (match[0].includes(this.formatSeparator) && !/{.*}/.test(match[1])) {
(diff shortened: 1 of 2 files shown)


jamuhl:
If possible would you be able to add a test to avoid future regression?

brett-js:
Hey @jamuhl , sorry about that. I'll amend the pull request :)

brett-js:
@jamuhl Just finished adding in some tests. Do let me know if you'd like further changes or have other feedback. I have plenty of spare time now that my country is in lockdown.

jamuhl:
thank you...landed in i18next@19.4.0

Arkellys:
What about nested translations with params?
```
{
    "key1": "Example: $t(key2, {\"count\": 0 }, uppercase)",
    "key2": "element",
    "key2_plural": "elements"
}
```

Could this be possible?

adrai:
> What about nested translations with params?
> 
> ```
> 
> {
> 
>     "key1": "Example: $t(key2, {\"count\": 0 }, uppercase)",
> 
>     "key2": "element",
> 
>     "key2_plural": "elements"
> 
> }
> 
> ```
> 
> 
> 
> Could this be possible?

Based on the tests, it should work: https://github.com/i18next/i18next/blob/master/test/i18next.interpolation.nesting.spec.js

Arkellys:
@adrai Yes, I was wondering if there was a way to make it work with formatting.

adrai:
> @adrai Yes, I was wondering if there was a way to make it work with formatting.

no

PR #1424, opened 2020-04-10
add getLanguagePartFromCode as in language fallback resolution, should fix #1422

### LINKED ISSUE #1422: Fallback without considering country code
I have the following config for `init`:
```javascript
{
    preload: ['no', 'en'],
    load: 'languageOnly',
    fallbackLng: {
      nb: ['no'],
      nn: ['no'],
      default: ['en'],
    },
    backend: {
      loadPath: path.resolve(__dirname, 'locales/{{lng}}.json'),
    },
}
```
If I send something like `nb-NO` here, it falls back to `en` instead of `no`, having this array of languages in the request (I'm using Express middleware): `["nb", "en"]`.
How do I fallback any `nb*` or `nn*` language, with or without country code, and any `no*` language with code to just `no`? The docs didn't help with that.
<details>
  <summary>Offtopic</summary>
  I was using `i18n` previously and I find out it was immensely easier to accomplish in it, I had to pass `no` and `en` locales to it and

```
fallbacks: {
  nb: 'no',
  nn: 'no',
},
```
and it would just work, setting `en` for any non-nb*, nn*, or no* language. And now I'm messing with that for 2 hours trying to get it working...
</details>

adrai:


jamuhl:
published in i18next@19.4.1

PR #1433, opened 2020-04-22
Update typescript definition for BackendModule callbacks

berickson1:
The docs specify that if there's an error, the second argument should be null, but the typescript types do not allow it.
Update the signature to allow `null` on the second argument.

Relevant snippet from docs:
```
    /* if method fails/returns an error, call this: */
    /* callback(truthyValue, null); */
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -700,7 +700,7 @@ export interface Module {
 }
 
 export type CallbackError = Error | null | undefined;
-export type ReadCallback = (err: CallbackError, data: ResourceKey) => void;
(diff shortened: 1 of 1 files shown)


jamuhl:
Actually it's a little more complicated: https://github.com/i18next/i18next/blob/master/src/BackendConnector.js#L140

```js
callback(null, {}); // ---> success 
callback('error', true); // ---> error, but try again to load (i18next will at max retry 5 times)
callback('error', false|null|undefined); // ---> error, do not retry that call
```

berickson1:
#1439 makes this obsolete. I still think that the typings should include null and undefined as per the comment above

PR #1439, opened 2020-04-26
Allow passing retry boolean to `ReadCallback`

unindented:
For `ReadCallback`, when `err` is an `Error`, the second argument should be a boolean indicating whether to retry.


rosskevin:
@jamuhl this is a patch

jamuhl:
thanks to all. was published in i18next@19.4.4

Issue #1440, opened 2020-04-29
i18n.use() definition not compatible with source code


zhoutwo:
Hi, I'm using TypeScript 3.8.3 and i18next 19.4.4, and when I create my custom backend module, and ran into issues.

If I create a class and generate the definition, with some importing and fixing compilation errors, I get this:

```typescript
class MyBackend implements BackendModule<MyBackendOptions> {
  type: 'backend' = 'backend';
  init(services: Services, backendOptions: MyBackendOptions, i18nextOptions: InitOptions): void {}

  create(languages: string[], namespace: string, key: string, fallbackValue: string): void {}

  read(language: string, namespace: string, callback: ReadCallback): void {}
```

However, if I do
```typescript
i18n.use(MyBackend)
```
I get
```
Uncaught Error: You are passing a wrong module! Please check the object you are passing to i18next.use()
```
Which points to
```javascript
if (!module.type) throw new Error('You are passing a wrong module! Please check the object you are passing to i18next.use()');
```

However, I double-checked the type definition:
https://github.com/i18next/i18next/blob/v19.4.4/index.d.ts#L815-L817
https://github.com/i18next/i18next/blob/v19.4.4/index.d.ts#L791
```typescript
export type Newable<T> = { new (...args: any[]): T };
```

The `Newable<T>` type does not have a property called `type`. It only constructs one with the property.
This means I cannot pass in the constructor like the built-in modules. I have to instantiate it outside and call `use()` on it.

One potential solution here is to use:
```typescript
export type ModuleNewable<T extends Module> = { 
  new (...args: any[]): T;
  type: T["type"];
};

// ...

use<T extends Module>(
    module: T | ModuleNewable<T> | ThirdPartyModule[] | ModuleNewable<ThirdPartyModule>[],
  ): i18n;
```

Then, `MyBackend` will become:
```typescript
class MyBackend implements BackendModule<MyBackendOptions> {
  static type: 'backend' = 'backend';
  type: 'backend' = 'backend'; // This is here because I didn't touch "Module" definition
  // ...
```

And now I can pass the constructor into `i18n.use()` and it works.

However, I do see that this definition breaks the exported type of `i18next-browser-languagedetector`, which makes my recommendation slightly risky.

For now, I think you can tell people to add `static type: 'backend' = 'backend';`, which is defining more than the interface, but TypeScript does not error out for extra static fields.

jamuhl:
[...]

PR #1442, opened 2020-05-01
Fix i18n.use() cannot pass module constructors in TypeScript

zhoutwo:
Changes mentioned in https://github.com/i18next/i18next/issues/1440

@rosskevin

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -754,8 +754,16 @@ export interface Services {
   resourceStore: ResourceStore;
 }
 
+export type SupportedType =
+  | 'backend'
+  | 'logger'
+  | 'languageDetector'
+  | 'postProcessor'
+  | 'i18nFormat'
+  | '3rdParty';
+
 export interface Module {
-  type: 'backend' | 'logger' | 'languageDetector' | 'postProcessor' | 'i18nFormat' | '3rdParty';
+  type: SupportedType;
 }
 
 export type CallbackError = Error | null | undefined;
@@ -850,7 +858,13 @@ export interface Modules {
 }
 
 // helper to identify class https://stackoverflow.com/a/45983481/2363935
-export type Newable<T> = { new (...args: any[]): T };
+export interface Newable<T> {
+  new (...args: any[]): T;
+}
+
+export interface NewableModule<T extends Module> extends Newable<T> {
+  type: T['type'];
+}
 
 export interface i18n {
   // Expose parameterized t in the i18next interface hierarchy
@@ -872,11 +886,11 @@ export interface i18n {
    * The use function is there to load additional plugins to i18next.
    * For available module see the plugins page and don't forget to read the documentation of the plugin.
    *
-   * Accepts a class or object
+   * @param module Accepts a class or object
    */
-  use<T extends Module>(
-    module: T | Newable<T> | ThirdPartyModule[] | Newable<ThirdPartyModule>[],
-  ): i18n;
+  use<T extends Module>(module: T | NewableModule<T>): this;
+
+  use<T extends ThirdPartyModule>(module: T[] | Array<NewableModule<T>>): this;
 
   /**
    * List of modules used
--- a/index.d.ts
+++ b/index.d.ts
@@ -888,7 +888,7 @@ export interface i18n {
    *
    * @param module Accepts a class or object
    */
-  use<T extends Module>(module: T | NewableModule<T>): this;
+  use<T extends Module>(module: T | NewableModule<T> | Newable<T>): this;
 
   /**
    * List of modules used
--- a/index.d.ts
+++ b/index.d.ts
@@ -754,7 +754,7 @@ export interface Services {
   resourceStore: ResourceStore;
 }
 
-export type SupportedType =
+export type ModuleType =
   | 'backend'
   | 'logger'
   | 'languageDetector'
@@ -763,7 +763,7 @@ export type SupportedType =
   | '3rdParty';
 
 export interface Module {
-  type: SupportedType;
+  type: ModuleType;
 }
 
 export type CallbackError = Error | null | undefined;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -810,11 +824,11 @@ export interface i18n {
...
    *
-   * Accepts a class or object
+   * @param module Accepts a class or object
    */
-  use<T extends Module>(
-    module: T | Newable<T> | ThirdPartyModule[] | Newable<ThirdPartyModule>[],
-  ): i18n;
+  use<T extends Module>(module: T | NewableModule<T>): this;
+
+  use<T extends ThirdPartyModule>(module: T[] | Array<NewableModule<T>>): this;
```
zhoutwo, line 831: I split this into two signatures to improve clarity.

zhoutwo: However, both https://github.com/i18next/i18next/blob/v19.4.4/src/i18next.js#L203-L204 and https://github.com/i18next/i18next/blob/v19.4.4/src/i18next.js#L226-L228 does not seem to suggest the `.use()` method takes an array (and the doc does not mention it either). Are we sure this is valid?

test/typescript/modules.test.ts
```diff
@@ -72,6 +72,7 @@ i18next.use(externalModules);
 
 // exercise class usage
 class MyLoggerModule implements LoggerModule {
+  static type: 'logger' = 'logger';
   type: 'logger' = 'logger';
```
zhoutwo, line 76: Unfortunate there is a duplication. However, they mean different things. The `static` one ensures the constructor has a property called `type`. And the original one ensures the instances have the said property. I fixed the overall issue in the easy way. The hard way is to acknowledge that `type` is only needed for that `.use()` method invocation and should separate this field from the `Module` interface for good. However, the problem is how can we make sure someone who makes a `LanguageDetector` will end up assigning the `type` field to be `languageDetector` when we separate them. That's the hard part and I don't have an easy solution for it.

zhoutwo: Nonetheless, if this change is approved. All other repos need also to duplicate the `type` field in their TypeScript definition. It seems it doesn't hurt break functionalities though, since all of them are implemented in JavaScript. But it will break those who use TypeScript.

index.d.ts
```diff
@@ -695,8 +695,16 @@ export interface Services {
   resourceStore: ResourceStore;
 }
 
+export type SupportedType =
```
rosskevin, line 698: Rename to `ModuleType`

index.d.ts
rosskevin: Based on the runtime, the signature with `[]` no longer appears correct, seems we should remove it.

test/typescript/modules.test.ts
rosskevin: This is the 100% correct interpretation of the existing runtime.  It seems rather inconvenient but there are tradeoffs made for flexibility and usage.  Given the frequency of usage is several magnitudes the number of module coded, it makes sense to keep the runtime signatures as-is and use this duplicative typescript use in subordinate modules.

With that said, add a comment to this particular file that simply states:
> Need both static and member definitions of `type` to satisfy `use()` signature, see #1442

zhoutwo:
Thanks @rosskevin , can you please take another look?

rosskevin:
@jamuhl this is going to be a typescript-breaking change for _every_ i18next module.  The effective example diff will be:

```diff
class MyLoggerModule implements LoggerModule {
+  static type: 'logger' = 'logger';
  type: 'logger' = 'logger';
```

So far it has only affected #1440, it can effectively be worked around using `as any`.  This change is technically correct but effects a small set of users.  Do you have a major release planned?  I think this is pretty signal via major release.  It's not a high priority to release.

Do you want to merge or hold?

jamuhl:
@rosskevin currently no major release is planned...let's keep this in on hold (At least if not that important for now)

rosskevin:
Thanks @zhoutwo - this will go into the next major release sometime in the future to minimize disruption to downstream i18next modules.

zhoutwo:
No problem. Thanks for the review!

adrai:
@pedrodurek is this still relevant with the new types?

pedrodurek:
👀

pedrodurek:
Hey @adrai and @zhoutwo, the solution proposed here, will break the use for most modules (unless you instantiate them), for example:
```tsx
// breaks
use(I18NextHttpBackend)

// doesn't break
use(new I18NextHttpBackend())
```

That happens because `typeof I18NextHttpBackend` won't have the property `type`. In order to work it, we must accept plain `Newable` as well.
```tsx
use<T extends Module>(module: T | NewableModule<T> | Newable<T>): this;
```
https://tsplay.dev/NBzddN

zhoutwo:
Looks like I have to give it both static and non static ones
```
static type: 'backend';
type: 'backend';
```

adrai:
Is there any disadvantage of accepting plain Newable as well?
As I've understood, this will not be breaking then, right?

zhoutwo:
Actually there is a problem: I can't have a member `type` and a static member `type` in the same interface at the same time

pedrodurek:
classes can though, but we'd have to update all modules (types).
Screenshot 2021-04-23 at 12 48 37 AM

zhoutwo:
I can certainly go in and update each concrete class, but do we want to accommodate custom backends? I'm trying to find a way to do this on the interface level

pedrodurek:
Does this approach work for you? If so, we wouldn't need to touch any of the modules.
`use<T extends Module>(module: T | NewableModule<T> | Newable<T>): this;`

zhoutwo:
Oh, I see. Sorry I misunderstood your point. Yes let me update the PR.

pedrodurek (approved):
LGTM! 🎉

pedrodurek:
Hey @adrai, can we merge this one? Considering the new approach, it shouldn't have breaking changes

PR #1458, opened 2020-05-20
fix: type declaration export

### LINKED ISSUE #1447: Mismatch between TypeScript types and CommonJS export
After [disabling an ESLint rule that was causing me trouble](https://github.com/i18next/i18next/issues/1446), I ran into a new issue trying to incorporate i18next into my project.

I'm writing code for a back-end NodeJS server that is compiled with TypeScript. Therefore the `import` syntax is handled not by Babel, but by TypeScript - and gets translated as such:

```
import i18next from "i18next";

// Becomes:

const i18next_1 = require("i18next").default;
```

This output JavaScript is then run in NodeJS like normal, relying on the `main` key in the `package.json` to locate the module. Thus it loads `./dist/cjs/i18next.js`, which exports like so:

```
//...

var i18next = new I18n();

module.exports = i18next;
```

As you can see, there's no export called "default". Therefore what I really need is to import the `module.exports` under the name i18next. This can be done in TypeScript:

```
import * as i18next from "i18next";
```

However when I do this, I start getting errors whenever I try to use it:

Screen Shot 2020-05-07 at 4 24 06 PM

Worse yet, since these are **TypeScript errors** and not ESLint errors, I can't just hand-wave them or disable them. Attempting to run `tsc` against my code stops compilation early with the error:

```
stevenbarnett@MacBook-Pro bf-hapijs-plugin % npx tsc
src/i18n/index.ts:33:35 - error TS2339: Property 'init' does not exist on type 'typeof import("/Users/stevenbarnett/Repos/pangolin-jerky/bf-hapijs-plugin/node_modules/i18next/index")'.

33      const translate = await i18next.init({
                                        ~~~~


Found 1 error.
```

For the record, if I **don't** try to import i18next this way (using `import * as ...`) then the following happens:

```
import i18next from "i18next";

console.log(i18next); // -> undefined

i18next.init(); // -> crashes with error "Cannot read property 'init' of undefined"
```

mjbcopland:
Fixes #1447.

Effectively reverts https://github.com/i18next/i18next/pull/1352/commits/13b10da7e0f368cf1ed4f1c62d5735c972d9e5d7. Hiding whitespace changes is recommended due to a significant change in indentation.

> [For an NPM package, `export =` is accurate if `node -p 'require("foo")'` works to import a module, and `export default` is accurate if `node -p 'require("foo").default'` works to import a module.](https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default)

```
$ node -p 'require("i18next")'        
I18n {
  observers: {},
  options: {},
  services: {},
  logger: Logger {
    prefix: 'i18next:',
    logger: {
      type: 'logger',
      log: [Function: log],
      warn: [Function: warn],
      error: [Function: error],
      output: [Function: output]
    },
    options: {},
    debug: undefined
  },
  modules: { external: [] }
}

$ node -p 'require("i18next").default'
undefined
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -679,10 +679,9 @@ declare namespace i18next {
     nest(str: string, fc: (...args: any[]) => any, options: InterpolationOptions): string;
   }
 
-  export class ResourceStore {
-    constructor(data: Resource, options: InitOptions);
-    public data: Resource;
-    public options: InitOptions;
+  export interface ResourceStore {
+    data: Resource;
+    options: InitOptions;
   }
 
   export interface Services {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,1022 +1,1024 @@
-export interface FallbackLngObjList {
-  [language: string]: string[];
-}
+declare namespace i18next {
+  export interface FallbackLngObjList {
+    [language: string]: string[];
+  }
 
-export type FallbackLng = string | string[] | FallbackLngObjList;
+  export type FallbackLng = string | string[] | FallbackLngObjList;
 
-export type FormatFunction = (value: any, format?: string, lng?: string) => string;
+  export type FormatFunction = (value: any, format?: string, lng?: string) => string;
 
-export interface InterpolationOptions {
-  /**
-   * Format function see formatting for details
-   * @default noop
-   */
-  format?: FormatFunction;
-  /**
-   * Used to separate format from interpolation value
-   * @default ','
-   */
-  formatSeparator?: string;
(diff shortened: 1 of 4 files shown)


mjbcopland:
This fixes consumers but breaks types in the library itself; will investigate.

rosskevin:
#1352 shows extensive research into this subject and is our chosen path.

PR #1459, opened 2020-05-27
shrink size another 2KB by using shorter variable names

adrai:
#### Checklist

- [ ] tests are included


adrai:
in .min.js decreases only 28bytes.... not that much

jamuhl:
build step of consumers should include minification

PR #1460, opened 2020-05-31
fix type declarion of exposed EventEmitter#off methods

jokester:
Hello there, I found that `off` method in ResourceStore and i18n had some incomplete TypeScript types (the method is inherited from parent EventEmitter)

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/exposed.test.ts
+++ b/test/typescript/exposed.test.ts
@@ -16,4 +16,5 @@ const modules: Modules = { external: [] };
 const resourceStore: ResourceStore = i18next.services.resourceStore;
 resourceStore.on('added', console.log);
 resourceStore.off('added', console.log);
+resourceStore.off('added');
 resourceStore.data['en'];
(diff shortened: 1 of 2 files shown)


test/typescript/exposed.test.ts
```diff
@@ -14,4 +14,6 @@ const mockWithT: WithT = {
 const modules: Modules = { external: [] };
 
 const resourceStore: ResourceStore = i18next.services.resourceStore;
+resourceStore.on('added', console.log);
+resourceStore.off('added', console.log);
```
rosskevin, line 18: Please add another usage test for the optional listerner, e.g. `resourceStore.off('added')`

adrai: @jokester as soon as you've completed the suggestions of @rosskevin we will merge this

test/typescript/exposed.test.ts
```diff
@@ -14,4 +14,7 @@ const mockWithT: WithT = {
...
+resourceStore.off('added');
```
adrai, line 19: @rosskevin ok?

test/typescript/exposed.test.ts
jokester: @rosskevin @adrai  Sorry I didn't see the notifications. Thanks for adding the missing code. Are there more to add?

adrai:
released with v19.5.4

PR #1461, opened 2020-06-09
use == everywhere in the rules

adrai:
currently it's mixed == and ===

#### Checklist

- [ ] tests are included


adrai:
decreases also the size by 5 bytes ;-)

PR #1471, opened 2020-06-21
getBestMatchFromCodes: use fallbackLng if nothing found, fixes #1470

### LINKED ISSUE #1470: fallbackLng is not respected with version 19.5.0 and languageDetector v5
## 🐛 Bug Report
When I upgraded i18next from 19.4.0 to 19.5.0 and browser languageDetector from 4.3.1 to 5, fallbackLng doesn't work anymore, reverting versions back make the issue disappear.

I'm configuring languageDetector to detect language from a cookie named "language", I later changeLanguage using clickable buttons.

Previously when the cookie isn't present(the user hasn't chosen a language yet), the App's language was set to the fallback and the cookie was added and set to the fallback too, now the cookie isn't set and this warning appears:

> i18n.languages were undefined or empty undefined

I tried awaiting the promise returned by `init` and `changeLanguage` before using the translation function with no use.

## To Reproduce
this is how I configure i18n:

```js

import i18n from "i18next";

import LanguageDetector from "i18next-browser-languagedetector";

// import arabic translations
import arTranslations from "./ar.json";
// import english translations
import enTranslations from "./en.json";

// resources object is passed to i18n
// for i18next to know our translations
const resources = {
    en: {
        translation: enTranslations,
    },
    ar: {
        translation: arTranslations,
    },
};

i18n
    // use the language detector library, its configuration is passed in init below
    .use(LanguageDetector)
    // init i18next
    .init({
        resources,
        // configuration for language detector library
        detection: {
            order: ["cookie"],
            caches: ["cookie"],
            lookupCookie: "language",
        },
        // currently we only support arabic and english
        supportedLngs: ["ar", "en"],
        // when no language is detected, fallback to arabic
        fallbackLng: "ar",
        debug: process.env.NODE_ENV === "development",
        interpolation: {            
            escapeValue: false,
        },
    });
// export the configured instance
export default i18n;

```

## Expected behavior

`fallbackLng` should be used when cookie "language" isn't set and the cookie should be added and set to `fallbackLng` just like it used to work.

## Your Environment

- *runtime version*: edge chromium 83.0.478.54
- *i18next version*: 19.5.0
- *os*: Windows 10 x64
- *i18next-browser-languagedetector*: 5.0.0

adrai:
#### Checklist

- [x] tests are included

### DIFF
--- a/src/LanguageUtils.js
+++ b/src/LanguageUtils.js
@@ -111,6 +111,9 @@ class LanguageUtil {
       });
     }
 
+    // if nothing found, use fallbackLng
+    if (!found) found = this.getFallbackCodes(this.options.fallbackLng)[0];
+
     return found;
   }
 
--- a/test/languageUtils.spec.js
+++ b/test/languageUtils.spec.js
@@ -228,6 +228,8 @@ describe('LanguageUtils', () => {
       { args: [['en-GB']], expected: 'en' },
       { args: [['ru', 'en-GB']], expected: 'en' },
       { args: [['de-CH']], expected: 'de-DE' },
+      { args: [['ru']], expected: 'en' },
+      { args: [[]], expected: 'en' },
     ];
 
     tests.forEach(test => {


jamuhl:
published in i18next@19.5.1

PR #1472, opened 2020-06-21
fix: getDataByLanguage typings & test

ljosberinn:
Fixes an issue I had with checking whether a certain language is already loaded without passing a namespace through `instance.hasResourceBundle(lng, ns)`.

Currently, the typings act as if you'd always receive an object with key `translation` (which might have to be fixed too, but I guess its fine for now). But if that language is not loaded yet, as [the underlying implementation just returns whatever it finds on `this.data`](https://github.com/i18next/i18next/blob/master/src/ResourceStore.js#L126), it will be undefined.

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -857,7 +857,7 @@ export interface i18n {
   /**
    * Returns a resource data by language.
    */
-  getDataByLanguage(lng: string): { translation: { [key: string]: string } };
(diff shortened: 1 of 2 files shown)


adrai:
released with v19.5.4

PR #1475, opened 2020-06-30
fix nesting interpolation with prepended namespace, fixes #1474

### LINKED ISSUE #1474: Nesting with namespace does not work when passed to i18n.t()
## 🐛 Bug Report

Nesting that includes a namespace no longer works when passed to `t()`.

## To Reproduce

Sample: https://codesandbox.io/s/mystifying-nightingale-hc8o2?file=/src/index.js

## Expected behavior

Given the translations
```js
        translation: {
          key: "hello world",
          hasNesting: "$t(translation:key)"
        }
```

All of the following should produce the same result `hello world`
```js
i18n.t('hasNesting'); // 'hello world' (correct)
i18n.t('key'); // 'hello world' (correct)
i18n.t('$t(key)'); // 'hello world' (correct)
i18n.t('translation:key'); // 'hello world' (correct)

i18n.t('$t(translation:key)'); // 'key)' (incorrect)
```

## Your Environment

- *runtime version*: firefox
- *i18next version*: 19.5.1
- *os*: Mac

adrai:
Fixes: https://github.com/i18next/i18next/issues/1474
as documented here: https://www.i18next.com/translation-function/nesting#basic
image

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -52,6 +52,13 @@ class Translator extends EventEmitter {
 
     let namespaces = options.ns || this.options.defaultNS;
     if (nsSeparator && key.indexOf(nsSeparator) > -1) {
+      const m = key.match(this.interpolator.nestingRegexp);
+      if (m && m.length > 0) {
+        return {
+          key,
+          namespaces,
+        };
+      }
       const parts = key.split(nsSeparator);
       if (
         nsSeparator !== keySeparator ||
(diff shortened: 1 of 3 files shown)


src/Translator.js
```diff
@@ -52,6 +52,13 @@ class Translator extends EventEmitter {
 
     let namespaces = options.ns || this.options.defaultNS;
     if (nsSeparator && key.indexOf(nsSeparator) > -1) {
+      const m = key.match(this.interpolator.nestingRegexp);
+      if (m && m.length > 0) {
+        return {
+          key,
+          namespaces,
```
edimitchel, line 59: Hi @adrai,
I located here an issue !

https://github.com/i18next/i18next/pull/1475/files#diff-d7b95552757249c43b739889204bb12002da36af74ac74506364a713ceecf86eR70

If the namespaces is a string, you made this, but not here.

PR #1488, opened 2020-07-15
Extend skipOnVariables

### LINKED ISSUE #1487: Unexpected interpolation within variables
# 🐛 Bug Report

During interpolation of variables, i18next also performs interpolation **within** the contents of the variables. I would not expect the variables to also be affected by interpolation, as the contents of them might not be from a trusted source (user input). At least there should be an option to disable this (which in my opinion should be the default in a future version).

I don't really understand why it would be desirable to use variables within variables, as developers can just inline the variables themselves instead of relying on i18next to derefence them.

An additional side effect of the interpolation within variables (and this side effect actually led me to find out about this issue) is that it makes it impossible to render double curly braces in a translated text. I have my own template engine in my application that also uses double braces for placeholders, and I wanted an introductory paragraph to explain how the user can use the placeholders, which required me to inline double braces within the text. But I did not find any way documented on how the double braces could be escaped, so I tried adding them through a variable, only to realize that the contents of the variable will also have interpolation applied (which leads to the removal of the "placeholder").

## To Reproduce

See [this codesandbox](https://codesandbox.io/s/i18next-bugreport-variable-resolving-8i1eq?file=/src/index.js).

Or to reproduce:

1. Add a translation like this: `example: "Example Placeholder: {{examplePlaceholder}}"`
2. Invoke the translation like so: `const result = t("example", { examplePlaceholder: "{{ThisShouldNotGetParsed}}" })`

## Expected behavior

I would expect the result to contain the string: `"Example Placeholder: {{ThisShouldNotGetParsed}}"`. However, the current i18next implementation would result in a string like this: `"Example Placeholder: "` (Because it tries to resolve `{{ThisShouldNotGetParsed}}` but is unable to, as there is no such variable. It even prints a warning in debug mode saying that the variable ThisShouldNotGetParsed is not set)

I also enabled the new `skipOnVariables` option as well as disabling escaping of values (with `escapeValue: false` in the interpolation options). Both did not change the result.

## Your Environment

- *runtime version*: node v12
- *i18next version*: 19.6.0
- *os*: Windows

adrai:
Should fix #1487

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -77,6 +77,10 @@ class Interpolator {
       (this.options && this.options.interpolation && this.options.interpolation.defaultVariables) ||
       {};
 
+    const skipOnVariables =
+      (options && options.interpolation && options.interpolation.skipOnVariables) ||
+      this.options.interpolation.skipOnVariables;
+
     function regexSafe(val) {
       return val.replace(/\$/g, '$$$$');
     }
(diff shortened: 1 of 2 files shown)


adrai:
This is the relevant line (condition) => https://github.com/i18next/i18next/pull/1488/files#diff-c4c32a53d5f4662890bf17b0a8fd47bcR128

Issue #1486, opened 2020-07-14
bindStore should be bindI18nStore in ReactOptions d.ts


JaniL:
## 🐛 Bug Report

`bindStore` has been [misnamed in the type file](https://github.com/i18next/i18next/blob/9be2ed6a9d0775b4a07a23f0cd85ae6d5a7f04cf/index.d.ts#L129) of i18next `ReactOptions`. To my understanding it should refer to `bindI18nStore` that is used in [React specific configuration attributes in i18next-react](https://github.com/i18next/react-i18next/blob/c84d98a2c62b926fee44fc99f12ab8993f8b3e78/src/context.js#L5).

## To Reproduce

```js
void i18n
  .use(initReactI18next)
  .init({
    debug: false,
    resources,
    lng: 'en',
    react: {
      bindI18n: 'loaded',
      bindI18nStore: 'added'
    }
  })
```

This results to following TS error:

```
Type '{ bindI18n: string; bindI18nStore: string; }' is not assignable to type 'ReactOptions'.
      Object literal may only specify known properties, and 'bindI18nStore' does not exist in type 'ReactOptions'.
```

## Expected behavior

TypeScript should not throw error from using `bindI18nStore`

## Your Environment

- *runtime version*: inode v12.16.1, ts 3.8.3
- *i18next version*: 19.6.0
- *react-i18next version*: 11.7.0
- *os*: macos
- *any other relevant information*

jamuhl:
PR welcome

PR #1489, opened 2020-07-15
Change bindStore to bindI18nStore in ReactOptions type and update default values

JaniL:
This resolves https://github.com/i18next/i18next/issues/1486. It also updates the default values of `bindI18n` and `bindI18nStore` attributes in `ReactOptions` type as they are described [here](https://github.com/i18next/react-i18next/blob/master/src/context.js#L3).
#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -126,7 +126,7 @@ export interface ReactOptions {
    * Set which events on store trigger a re-render, can be set to false or string of events
    * @default 'added removed'
    */
-  bindStore?: string | false;
+  bindI18nStore?: string | false;
   /**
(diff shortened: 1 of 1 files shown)


adrai:
thank you, landed in v19.6.1

JaniL:
Great, no need to keep `// @ts-expect-error` in my code anymore then. 😄  Thanks!

adrai:
If you like this module don’t forget to star this repo. Make a tweet, share the word or have a look at our https://locize.com to support the devs of this project.

There are many ways to help this project 🙏

PR #1492, opened 2020-07-18
Add a plural resolver for locale "ht"

### LINKED ISSUE #1491: Locale "ht" is missing from PluralResolver.js
## 🐛 Bug Report

Locale "ht" is missing from [PluralResolver.js](https://github.com/i18next/i18next/blob/master/src/PluralResolver.js). Tools such as [i18next-parser](https://github.com/i18next/i18next-parser) expect there to be a valid PluralResolver for each desired locale. Since "ht" has no PluralResolver, i18next-parser fails when passed this locale.

## To Reproduce

Configure i18next-parser.config.js with:

```js
module.exports = {
  locales: ["ht"]
};
```

and run it against a codebase. Fails with a null-reference exception at https://github.com/i18next/i18next-parser/blob/master/src/transform.js#L130. The root cause is that there is no setting for "ht" at https://github.com/i18next/i18next/blob/master/src/PluralResolver.js#L15. 

## Expected behavior

Given that "ht" [doesn't support plurals](https://www.howtocreole.com/2018/03/plural-words-haitian-creole.html), "ht" should be added at https://github.com/i18next/i18next/blob/master/src/PluralResolver.js#L15. For example:

```js
  { lngs: ['ay','bo','cgg','fa','ht','id','ja','jbo','ka','kk','km','ko','ky','lo',
    'ms','sah','su','th','tt','ug','vi','wo','zh'], nr: [1], fc: 3 },
```

## Your Environment

- *runtime version*: node v14
- *i18next version*: 19.6.1
- *os*: Mac

marcja:
Fix #1491

#### Checklist

- [ ] run tests `npm run test`
- [ ] tests are included

### DIFF
--- a/src/PluralResolver.js
+++ b/src/PluralResolver.js
@@ -12,7 +12,7 @@ let sets = [
     'ne','nl','nn','no','nso','pa','pap','pms','ps','pt-PT','rm','sco',
     'se','si','so','son','sq','sv','sw','ta','te','tk','ur','yo'], nr: [1,2], fc: 2 },
 
-  { lngs: ['ay','bo','cgg','fa','id','ja','jbo','ka','kk','km','ko','ky','lo',
+  { lngs: ['ay','bo','cgg','fa','ht','id','ja','jbo','ka','kk','km','ko','ky','lo',
     'ms','sah','su','th','tt','ug','vi','wo','zh'], nr: [1], fc: 3 },
 
   { lngs: ['be','bs', 'cnr', 'dz','hr','ru','sr','uk'], nr: [1,2,5], fc: 4 },


jamuhl:
`ht` can't be found in the official https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html#ht what is the language called (guess it's Haiti)?

marcja:
Yes, Haitian Creole. It's a required locale for a project I'm working on.

jamuhl:
Will be deployed with the next version...hope that is ok for you?

If needing it now, you can add rules using `i18next.services.pluralResolver.addRule` meanwhile...https://github.com/i18next/i18next/blob/master/src/PluralResolver.js#L90

marcja:
Awesome. Thanks!

adrai:
released with v19.6.3

PR #1518, opened 2020-11-11
Update URLs

KristjanESPERANTO:


jamuhl:
thank you for taking the time and providing the PR

Issue #992, opened 2017-11-12
Howto Use interpolation format function with parameters?


daKmoR:
hey,

so we have a ton of already translated elements in the Intl Api but I really would like to move to i18next.
Problem is some of those translation messages are using pretty specific Intl Apis. (e.g. intl-messageformat)

I though yeah no problem just provide interpolate format function... but I don't have the parameters in the interpolation format function...

If I had I could just add a format for Intl like so right?
```
en:
  translations:
    key: "Hey {{there}}",
    keyWithIntl: "{{Hey {there}, Intl}}"

interpolation:
  format: function(value, format, language, parameters) { //parameters would be new
    if (format === 'Intl') {
      let intlObject= new IntlMessageFormat(value, language);
      return intlObject.format(parameters);
    }
  }
```

jamuhl:
i personally already thought about adding a translator supporting icu format. But ended in not doing it...as there are a lot of options out there to support icu. But did not have the migration case in mind.

But for a migration use case i would not do it by interpolation - but adding a post processor: https://www.i18next.com/misc/creating-own-plugins.html#post-processor

using that like eg. the interval plural post processor: https://github.com/i18next/i18next-intervalPlural-postProcessor

so instead of using format, which only passes an interpolate value you get the complete found value - translation in message format and from there you can parse and format like in your code snipplet.

daKmoR:
oh I see - it's a pity it does not work as I hoped... would have been so easy :p but makes total sense :)
never mind as long as there is a way I guess I can make it work - so I'm off trying my luck on creating an intl postProcessor :p
thx for the hint and the example :)

daKmoR:
Simple Test
```
let IntlPostProcessor = {
  type: 'postProcessor',
  name: 'intl',
  process: function(value, key, options, translator) {
    console.log(value, key, options, translator);
    return value;
  }
};

i18next
  .use(i18nextXHRBackend)
  .use(i18nextBrowserLanguageDetector)
  .use(IntlPostProcessor)
  .init(settings);
````

console.log never get's called? what am I missing?

**Comment by daKmoR** [comment]
ok the PostProcessor was actually easy :p

```
let IntlPostProcessor = {
  type: 'postProcessor',
  name: 'intl',
  process: function(value, key, options, translator) {
    if (value.indexOf('{') !== -1 && value.indexOf('{{') === -1) {
      let parameters = options[1];
      let intlObject = new IntlMessageFormat(value, translator.language);
[...]

PR #1520, opened 2020-11-18
Updated FormatFunction signature to match code

klevytskyi:
According to this issue: https://github.com/i18next/i18next/issues/992 and code that was added, made as tweak to get 4th param in format function to be able to use it with TS.


adrai:
lgtm

adrai:
Released with v19.8.4
Thank you for your contribution.

Issue #1504, opened 2020-09-09
Make the translation function fully type-safe


karol-majewski:
## 🚀 Feature Proposal

With the latest advancements in TypeScript, it's now possible to infer the correct paths from an object type. TypeScript 4.1 introduces [recursive conditional types](https://github.com/microsoft/TypeScript/pull/40002), and there is a [PR open](https://github.com/microsoft/TypeScript/pull/40336) that will enable string type concatenation on the type level.

## Motivation

`18next.t` accepts arbitrary strings. This allows silly human errors, since it's easy to make a typo.

## Example

[...]

PR #1523, opened 2020-12-08
Add ability to type translation keys

pierpo:
This is a first step towards solving https://github.com/i18next/i18next/issues/1504
It allows you to type your translation keys freely. Which allows you to type them using the template literals of Typescript 4.1 for nested keys 😉  (see #1504).

# Usage

In my project, I augment the module definition in a type definition file`typings/i18next.d.ts`.

```typescript
import 'i18next';

declare module 'i18next' {
  export interface TranslationKeys {
    keys: 'a' | 'b' | 'c';
  }
}
```

Then in the whole project, I may use the `t` function like this:

```typescript
t('a'); // OK
t('wrong'); // Fails!
```

Same if I use `react-i18next`.

```typescript
const Component = () => {
  t('a'); // OK
  t('wrong'); // Fails!
  // ...
}
```

## Advanced usage

<details>
  <summary>Deeply typed keys using Typescript 4.1</summary>

```typescript
import 'i18next';

import fr from 'src/i18n/translations/fr.json'

type PathOf<T extends object> = {
  [K in keyof T]: T[K] extends object ? [K, ...PathOf<T[K]>] : [K]
}[keyof T];

type SerializedPathOf<T extends object> = Join<Extract<PathOf<T>, string[]>, '.'>;

type Join<T extends string[], D extends string> =
    T extends [] ? '' :
    T extends [unknown] ? `${T[0]}` :
    T extends [unknown, ...infer U] ? `${T[0]}${D}${Join<U, D>}` :
    string;


declare module 'i18next' {
  export interface I18nTranslationKeys {
    keys: SerializedPathOf<typeof fr>;
  }
}
```

</details>

# What's left to do

Well, it's just a first step. I'm open to suggestions: overriding the interface this way feels hacky, but I found no other way.

It also needs documentation somewhere.

- [ ] tests are included
  - I would love to write a test, but adding an augmentation breaks the other tests since it's only done globally :cry: 

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,6 +2,11 @@ export interface FallbackLngObjList {
   [language: string]: string[];
 }
 
+/**
+ * Augment this with a 'keys' key where you can add all the available translation keys
+ */
+export interface I18nTranslationKeys {}
+
 export type FallbackLng = string | string[] | FallbackLngObjList | ((code:string) => string | string[] | FallbackLngObjList);
 
 export type FormatFunction = (value: any, format?: string, lng?: string, options?: InterpolationOptions & { [key: string]: any }) => string;
@@ -639,7 +644,7 @@ export type Callback = (error: any, t: TFunction) => void;
  * Uses similar args as the t function and returns true if a key exists.
  */
 export interface ExistsFunction<
-  TKeys extends string = string,
+  TKeys extends TranslationKeys = TranslationKeys,
   TInterpolationMap extends object = StringMap
 > {
   (key: TKeys | TKeys[], options?: TOptions<TInterpolationMap>): boolean;
@@ -652,11 +657,16 @@ export interface WithT {
 
 export type TFunctionResult = string | object | Array<string | object> | undefined | null;
 export type TFunctionKeys = string | TemplateStringsArray;
+
+/**
+ * Gets the defined translation keys if they have been overridden, 'string' type otherwise
+ */
+type TranslationKeys = I18nTranslationKeys['keys'] extends string ? I18nTranslationKeys['keys'] : TFunctionKeys;
 export interface TFunction {
   // basic usage
   <
     TResult extends TFunctionResult = string,
-    TKeys extends TFunctionKeys = string,
+    TKeys extends TranslationKeys = TranslationKeys,
     TInterpolationMap extends object = StringMap
   >(
     key: TKeys | TKeys[],
@@ -665,7 +675,7 @@ export interface TFunction {
   // overloaded usage
   <
     TResult extends TFunctionResult = string,
-    TKeys extends TFunctionKeys = string,
+    TKeys extends TranslationKeys = TranslationKeys,
     TInterpolationMap extends object = StringMap
   >(
     key: TKeys | TKeys[],


index.d.ts
```diff
@@ -650,13 +654,17 @@ export interface WithT {
...
+ */
+type TKeys = TranslationKeys['keys'] extends string ? TranslationKeys['keys'] : string;
+                     
 export type TFunctionResult = string | object | Array<string | object> | undefined | null;
 export type TFunctionKeys = string | TemplateStringsArray;
 export interface TFunction {
   // basic usage
   <
     TResult extends TFunctionResult = string,
-    TKeys extends TFunctionKeys = string,
```
pierpo, line 659: `TFunctionKeys` is `string | TemplateStringsArray`. What was the point of `TemplateStringsArray`?

I guess I introduced a regression by removing this but I need to understand the type to fix it 😉

index.d.ts
```diff
@@ -650,13 +655,18 @@ export interface WithT {
   t: TFunction;
 }
 
+/**
+ * Gets the defined translation keys if they have been overridden, 'string' type otherwise
+ */
+type TranslationKeys = I18nTranslationKeys['keys'] extends string ? I18nTranslationKeys['keys'] : string;
```
pierpo, line 661: This raises a typescript error, so it should be done another way, but I have no idea how.

(btw, `npm run tsc` should be run in the CI as well, CI checks are green but they shouldn't)

pierpo: (btw, `npm run tsc` should be run in the CI as well, CI checks are green but they shouldn't)

pierpo: I had a long thought about this, and it might have no solution as it is. I can't think of any way to add a strict augmented type that would prevail over the default type.

It works here as it is, but it needs a `@ts-ignore`, and I doubt that it's acceptable.

jamuhl:
I'm not 100% sure why users insist on having the keys typed -> personally I add keys on the fly and let them added to the JSON by the saveMissing feature of i18next -> typed you have to manually add keys upfront to not break (or do I misunderstand how this works?)

adrai:
//cc: @pedrodurek

pierpo:
> I'm not 100% sure why users insist on having the keys typed -> personally I add keys on the fly and let them added to the JSON by the saveMissing feature of i18next -> typed you have to manually add keys upfront to not break (or do I misunderstand how this works?)

Indeed, such a feature makes it fine to work with the keys (I didn't know it to be honest, thanks!). But typing provides the following advantages:

- I can't make a mistake. The second I make a mistake, my editor tells me, and the CI checks will be red.
- I can safely remove outdated keys. If they're still being used, I'll get an error.
- The auto-completion is great for devX.

Also, it feels more natural to add my keys in the JSON then using them with the autocompletion :blush:

pierpo:
Oh...  I just noticed that this work has already been done on `react-i18next` directly!

https://github.com/i18next/react-i18next/pull/1193

But I think it should be in here as well. It should be available to all frameworks!

pedrodurek:
Hey @pierpo thanks flagging that, as soon as we fixed all issues pointed out for the new types (almost there), we can certainly bring the types to here, if it's interesting for everyone.

jamuhl:
I like the comment here: https://github.com/i18next/i18next/issues/1504#issuecomment-741653842 would be great to keep things opt-in and not again break existing builds.

PR #1532, opened 2020-12-31
chore: update rollup and plugins

pravi:
Updated rollup to a recent version and replace deprecated rollup-plugins with new versions

#### Checklist

- [ ] tests are included


jamuhl:
thanks again

PR #1551, opened 2021-02-13
Updated changeLanguage signature to match codebase

### LINKED ISSUE #1550: changeLanguage() type definition is incorrect
## 🐛 Bug Report

The documentation says [here](https://www.i18next.com/overview/api#changelanguage):
> Calling changeLanguage without `lng` uses the language detector to choose the language to set.

But the TS definition file has this definition ([here](https://github.com/i18next/i18next/blob/master/index.d.ts#L885)):

```ts
changeLanguage(lng: string, callback?: Callback): Promise<TFunction>;
```

So the `lng` is required when we use TypeScript.

## Expected behavior

In the codebase, the following done when the call is parameter-less:

https://github.com/i18next/i18next/blob/ad88092292cc32fe250c6352f796dd710705d0da/src/i18next.js#L296-L299

So the TS definition should be:
```ts
changeLanguage(lng?: string, callback?: Callback): Promise<TFunction>;
```

## Your Environment

- *i18next version*: 19.8.7

Drarig29:
Fix #1550 

The `lng` parameter must be optional, to be able to run the function without any parameter to use the language detector.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -882,7 +882,7 @@ export interface i18n {
    * Changes the language. The callback will be called as soon translations were loaded or an error occurs while loading.
    * HINT: For easy testing - setting lng to 'cimode' will set t function to always return the key.
    */
-  changeLanguage(lng: string, callback?: Callback): Promise<TFunction>;
+  changeLanguage(lng?: string, callback?: Callback): Promise<TFunction>;
 
   /**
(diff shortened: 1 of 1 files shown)


jamuhl:
@adrai merged but will wait for release till your fix lands...

PR #1553, opened 2021-02-14
fix for cases when calling changeLanguage before finished to initialized

### LINKED ISSUE #1552: fallbackLng as default lng
Why did you change the behavior in the patch version?

19.8.5 - "use fallbackLng as default lng"

After update, language is always taken from "fallbackLng"
image

The behavior can be viewed in this repository
https://github.com/ogarich89/iso/tree/feature/i18next

adrai:
should fix #1552

#### Checklist

- [ ] tests are included

### DIFF
--- a/src/i18next.js
+++ b/src/i18next.js
@@ -154,14 +154,17 @@ class I18n extends EventEmitter {
     const deferred = defer();
 
     const load = () => {
-      this.changeLanguage(this.options.lng, (err, t) => {
+      const finish = (err, t) => {
         this.isInitialized = true;
         if (!this.options.isClone) this.logger.log('initialized', this.options);
         this.emit('initialized', this.options);
 
         deferred.resolve(t); // not rejecting on err (as err is only a loading translation failed warning)
         callback(err, t);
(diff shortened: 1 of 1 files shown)


src/i18next.js
```diff
@@ -154,14 +154,17 @@ class I18n extends EventEmitter {
...
         this.isInitialized = true;
         if (!this.options.isClone) this.logger.log('initialized', this.options);
         this.emit('initialized', this.options);
 
         deferred.resolve(t); // not rejecting on err (as err is only a loading translation failed warning)
         callback(err, t);
-      });
+      };
+      // fix for use cases when calling changeLanguage before finished to initialized (i.e. https://github.com/i18next/i18next/issues/1552)
+      if (this.languages && this.options.compatibilityAPI !== 'v1') return finish(null, this.t.bind(this));
```
jamuhl, line 166: is it API v1? v1 is more the v1 of i18next...this change was more v?!

adrai: this is just because of the v1 tests

PR #1558, opened 2021-02-19
Improve updateMissing default plural behavior

### LINKED ISSUE #1549: Missing key default pluralization issues
## 🐛 Bug Report

There seems to be some issues with default value pluralization when [`missingKeys`](https://www.i18next.com/overview/configuration-options#missing-keys) is enabled:

1. The result of interpolation is incorrect and does not seem to respect pluralization
2. The singular and plural values passed to the backend, when configured, are the same

## To Reproduce

Here's a [CodeSandbox](https://codesandbox.io/s/i18next-missing-key-pluralization-40ktz?file=/src/App.js) that illustrates the issue when evaluating:

```js
t("myMissingKey", {
    count,
    defaultValue: "{{ count }} word",
    defaultValue_plural: "{{ count }} words"
})
```

This is _not_ an issue when `saveMissing=false`.

## Expected behavior

Per the [`defaultValue` description](https://www.i18next.com/translation-function/essentials#overview-options):

> you also can define defaults for plurals by adding defaultValue_plural / defaultValue_2 -> _suffix depends on same pluralrules

I would expect the `myMissingKey` string to be pluralized correctly as `count` changes, and the missing key handler configured via a backend, should receive:

```json
{
    "myMissingKey": "{{ count }} word",
    "myMissingKey_plural": "{{ count }} words"
}
```

## Your Environment

- *runtime version*: Chrome v88
- *i18next version*: v19.8.7
- *react-i18next version*: v11.8.5
- *i18next-locize-backend version*: v4.1.9
- *os*: macOS

namoscato:
Fixes #1549

* This should at least fix default plural values sent to the backend when `updateMissing` is enabled. I'm not sure if this will _also_ fix the interpolation result described in #1549.
* Remove some test duplication

#### Checklist

- [x] tests are included

### DIFF
--- a/src/PluralResolver.js
+++ b/src/PluralResolver.js
@@ -102,18 +102,17 @@ class PluralResolver {
   }
 
   getPluralFormsOfKey(code, key) {
-    const ret = [];
+    return this.getSuffixes(code).map((suffix) => key + suffix)
+  }
 
+  getSuffixes(code) {
     const rule = this.getRule(code);
 
-    if (!rule) return ret;
+    if (!rule) {
+      return [];
+    }
 
-    rule.numbers.forEach((n) => {
-      const suffix = this.getSuffix(code, n);
-      ret.push(`${key}${suffix}`);
-    });
-
-    return ret;
+    return rule.numbers.map((number) => this.getSuffix(code, number))
   }
 
   getSuffix(code, count) {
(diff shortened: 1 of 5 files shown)


namoscato:
I can look into the CI failures this weekend...

jamuhl:
Thank you for contributing this improvement. Changes were published in i18next@19.9.0

PR #1560, opened 2021-02-22
fix: type definitions

javiertury:
- Allow nullable "data" in BackendModule.read callback. According to the docs, callback(err, null) is valid.
- BackendModule.create is optional.
- BackendModule.readMulti should use MultiReadCallback instead of ReadCallback.

#### Checklist

- [ ] tests are included        <- Not needed (in my opinion)
- [ ] documentation is changed or added  <- Not needed (in my opinion)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -729,8 +729,8 @@ export interface Module {
 }
 
 export type CallbackError = Error | null | undefined;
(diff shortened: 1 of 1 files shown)


jamuhl:
thank you for contributing - was published in i18next@19.9.0

PR #1563, opened 2021-02-26
Fallback on singular defaultValue

namoscato:
Relates to https://github.com/i18next/i18next/issues/1549

Minor followup to https://github.com/i18next/i18next/pull/1558 that fallbacks to the singular `defaultValue` if the plural variant is undefined, which resembles the previous behavior. This at least supports scenarios in which the singular and plural strings are the same.

#### Checklist

- [x] tests are included


jamuhl:
published in i18next@19.9.1

PR #1571, opened 2021-03-09
[F-1569] Add key name to format function as optional parameter

plindse:
#### Checklist

[x] only relevant code is changed (make a diff before you submit the PR)
[x] run tests `npm run test`
[x] tests are included
[x] documentation is changed or added


adrai:
not only the relevant code is changes

please also add at least 1 test

btw: this is a breaking change and probably won't be merged like this

jamuhl:
Might be I missed it - but what is the use case of doing a format based on the key?

plindse:
I've added test cases for both the format function and Interpolator.js.  There were a couple of breaking issues that the tests uncovered.  Let me know what else is needed and I'll resolve.

adrai:
I still don't get why this should not be possible with a second format.

localCurrency and convertedCurrency

"Amount 1 {{localValue, localCurrency}} and Amount 2 {{convertedValue, convertedCurrency}}"

This way it's also more clear for the translators.

adrai:
btw: https://stackoverflow.com/questions/63586596/how-to-format-currency-using-react-i18next-library

plindse:
We have dozens of options and would have to add one for every possible combination.  For retail, there are at least "storeLocale" "customerLocale" "operatorLocale" in combination with multiple currencies.  Some stores support several currencies at the same time.  We would need to basically provide the name of the variable with every option and those would be the only ones allowed.  Every change in what the data represents would need to have a new entry in the format function.  However, if we provide what the source of the data was, then the value can be indexed without requiring coding changes.  Even if we had an index into which time the format function was called within the translate function, we'd be able to determine what the source of the data was.

adrai:
maybe having the key in the options is more non-breakable and will not change the function signature?
something like:
image

@jamuhl ?

jamuhl:
@plindse why not just call `i18next.t(key, { key })` -> options already get passed to the format function

jamuhl:
or why not just `i18next.t(key, { lng })` -> where the lng is set to target lng you need?

adrai:
@jamuhl because there are multiple format interpolated variables in the same translation string :-(

adrai:
If i've understood correctly, something like this is desired:

```js
const i18next = require('i18next')
i18next.init({
  lng: 'en',
  resources: {
    en: {
      translation: {
        twoInterpolationsWithUniqueFormatOptions:
          'The value is {{localValue, currency}} or {{altValue, currency}}',
      },
    },
  },
  interpolation: {
    format: function(value, format, lng, options) {
      // desired is options.key to be first localValue and on second call altValue
      if (format === 'currency')
        return Intl.NumberFormat(options.parmOptions[options.key].locale, {
          style: 'currency',
          currency: options.parmOptions[options.key].currency,
        }).format(value);
      return value;
    }
  }
})

i18next.t('twoInterpolationsWithUniqueFormatOptions', {
  localValue: 12345.67,
  altValue: 16543.21,
  parmOptions: {
    localValue: { currency: 'USD', locale: 'en-US' },
    altValue: { currency: 'CAD', locale: 'fr-CA' },
  }
})
// The value is $12,345.67 or 16 543,21 $
```

plindse:
@adrai  If you are having trouble with passing the key name as a parameter, I did think about another option like what you described above.  The current interpolation key name could be added by the translation function to the options (ie. options.key) that are passed into the format function. It has to be added by the translate function itself because the format function is called multiple times.  I didn't care for that option as much, because options.key would have to be a unique name that wouldn't collide with user's person choice of variable names.  Maybe options.interpolationkeyname woudl be unique enough.

adrai:
@plindse can you refactor the code, so the format function takes the additional interpolationkeyname via options?

plindse:
@adrai Yes.  I'll work on that.

plindse:
@adrai Done.

src/Interpolator.js
```diff
@@ -196,6 +200,7 @@ class Interpolator {
       if (match[0].includes(this.formatSeparator) && !/{.*}/.test(match[1])) {
         const r = match[1].split(this.formatSeparator).map(elem => elem.trim());
         match[1] = r.shift();
+        r.push(match[1]);
```
adrai, line 203: Can you add a specific test that uses this code part?

src/Interpolator.js
```diff
@@ -213,7 +218,10 @@ class Interpolator {
       }
 
       if (doReduce) {
-        value = formatters.reduce((v, f) => this.format(v, f, options.lng, options), value.trim());
+        value = formatters.reduce(
+          (v, f, k) => this.format(v, f, options.lng, { ...options, interpolationkey: k }),
```
adrai, line 222: Isn't k not just the index in the array? Dies this work?

adrai:
@plindse did you have a chance to check my comments?

plindse:
Yes, working on it, but on vacation. Will try to get it tonight.

Sent from ProtonMail mobile

-------- Original Message --------

adrai:
no rush

src/Interpolator.js
plindse: Removed line.  It wasn't necessary and did case an issue.

src/Interpolator.js
plindse: No, test case revealed it. It didn't cause an issue, but it didn't do what was expected.  This has been changed to properly pass the key name to the format function.

plindse:
@adrai If you have any suggestions on how to resolve the codeclimate issue, I'd appreciate it.   I didn't change the use of a function in the loop, but it now is getting flagged.  I've tried several options and none have worked.

adrai:
lgtm

adrai:
thank you, will be included in the next version

plindse:
Thank you.

plindse:
@adrai What is typical publish schedule?

adrai:
@plindse just want to wait if there is an additional TypeScript fix that should be done... and waiting for @pedrodurek opinion, if we should release v20.1.0 now or wait for a potential TypeScript fix...
https://github.com/i18next/i18next/pull/1583#issuecomment-805588638
https://github.com/i18next/i18next-http-backend/issues/58#issuecomment-805758116

pedrodurek:
Hey @adrai, you can merge it, the fixes will be on the backend plugins.

adrai:
released in v20.1.0

PR #1586, opened 2021-03-23
Add alwaysFormat into the types

bsalesc:
There is the config alwaysFormat that works fine, but is missing from the types.

#### Checklist

- [ ] tests are included


adrai:
thank you

bsalesc:
@adrai where would be the best place to add in the docs? maybe here https://www.i18next.com/translation-function/interpolation?

PR #1617, opened 2021-05-27
1614 - Disable warn 'accessing an object' when the  'returnedObjectHandler' is defined

### LINKED ISSUE #1614: There's always a warning when using "returnedObjectHandler" + "returnObjects = false" + "debug= true " options
## 🐛 Bug Report

When you use i18next in `debug mode` ,  set `returnObjects = false` and define a `returnedObjectHandler`, there's a warning in console: `accessing an object - but returnObjects options is not enabled!` when translating a key that has an object/array inside.

## To Reproduce

https://codesandbox.io/s/i18next-forked-ui2p0?file=/src/i18n.js:233-377

With the following configuration: 

```tsx
  debug: true,
  returnObjects: false,
  // This is a dummy implementation
  returnedObjectHandler: (key) => `"${key}" with returnedObjectHandler`
``` 

And a resource like this: 
```json
"myObject": { "foo": " "  }
``` 

You will see the warning in console: `accessing an object - but returnObjects options is not enabled! `

## Expected behavior

I expected that, if I have defined a `returnedObjectHandler` it would not appear a warning, because the `returnObjects` is disabled on purpose so that the `returnedObjectHandler` can be executed.

In my opinion this warning should only be logged when the `returnObjects = false` but no `returnedObjectHandler` is defined.

The source code would be [this line](https://github.com/i18next/i18next/blob/b3f9d4aaa1e0d1e4ab2006402b37c37406bda53a/src/Translator.js#L134) and I suggest a change like this:

image

What do you think? I can **create a PR** if you think this change is valid

## Your Environment

- *runtime version*: Node v14.15.4, Chrome 90.0.4430.93
- *i18next version*: 20.3.0
- *os*: Linux

miguelCT:
Disable warn 'accessing an object' when the  'returnedObjectHandler' is defined

I have prepared the tests for the `returnObject = false` so that more cases can be added. I'll provide more in another MR

closes #1614 

#### Checklist

- [x] tests are included


jamuhl:
lgtm

adrai (approved):

adrai:
@miguelCT #1619 got already merged, can you update this PR with the master?

adrai:
ok, I've done this locally...

adrai:
released with v20.3.1

PR #1619, opened 2021-05-27
1618 - Add 'ns' to the 'returnedObjectHandler' options

### LINKED ISSUE #1618: The namespace from the translated key is lost when using "returnedObjectHandler"
## 🐛 Bug Report

In the **returnedObjectHandler** function, the _namespace_ that comes in the `options` object has not the correct value.

## To Reproduce

[codesandbox](https://codesandbox.io/s/i18next-forked-ibp1m?file=/src/app.js:234-297) example

## Expected behavior

With this configuration
```tsx
init({
...
  defaultNS: 'common',
  returnObjects: false,
  returnedObjectHandler: (key, value, options) =>
    `ns: "${options.ns}" - key: "${key}"`
})
```

And this use of `t`: 

```tsx
t('ns1:foo') {/* <-- ❌ It should have options.ns: "ns1" */}
``` 
`options.ns` should be `["ns1"]`, but it has `["common"]` instead ( "defaultNS" value)

**I'll provide a PR with a possible fix**

## Your Environment

- *runtime version*: Node v14.15.4, Chrome 90.0.4430.93
- *i18next version*: 20.3.0
- *os*: Linux

miguelCT:
Add 'ns' to the 'returnedObjectHandler' options

I've used the same base code from https://github.com/i18next/i18next/pull/1617 to add tests for the `returnObject = false` case

closes #1618

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -133,7 +133,7 @@ class Translator extends EventEmitter {
       if (!options.returnObjects && !this.options.returnObjects) {
         this.logger.warn('accessing an object - but returnObjects options is not enabled!');
         return this.options.returnedObjectHandler
-          ? this.options.returnedObjectHandler(resUsedKey, res, options)
+          ? this.options.returnedObjectHandler(resUsedKey, res, { ...options, ns: namespaces })
           : `key '${key} (${this.language})' returned an object instead of string.`;
       }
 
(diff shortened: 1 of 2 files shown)


jamuhl:
lgtm

adrai (approved):

adrai:
released with v20.3.1

PR #1626, opened 2021-07-08
Add transWrapTextNodes to types

Kamahl19:


jamuhl:
lgtm

adrai:
@pedrodurek shouldn't this be added in react-i18next?

pedrodurek:
Hey @adrai. Actually, no, it shouldn't. `react-18next` relies on ReactOptions coming from `i18next`.
https://github.com/i18next/react-i18next/blob/master/ts4.1/index.d.ts#L66-L67
https://github.com/i18next/react-i18next/blob/85889f21141950bf9c52e57326fac12231380f86/src/Trans.js#L287

adrai:
The question is, if this could be changed, like the other i18next plugins, like i18next-http-backend?

pedrodurek:
I'll take a look

adrai:
no rush, was just an idea 😉

adrai:
@Kamahl19 released with v20.3.3

PR #1629, opened 2021-07-20
fix(types): only use readonly array of strings in API

pcorpet:
In the API, i18next uses many `string[]` but has no intention to modify them (for inputs or properties), nor that user changes them (for outputs). Making them all readonly arrays make it closer to what's expected without breaking any valid usage.

#### Checklist

- [X] tests are included


adrai:
lgtm... @pedrodurek ?

pedrodurek (approved):

adrai:
thank you, included in v20.3.4

PR #1635, opened 2021-08-04
change to return false when a nullable argument is passed

raccoonback:
I have used i18next, react-i18next.
However, when using the i18n object to determine the existence of the key, I found that passing one `nullish factor(undefined or null, or '')` results in the error below:

```js
TypeError
Cannot read property 'forEach' of undefined
```

I think `i18n.exist(undefined or null, or '')` should return` false`.

Similar to mine, there will be many cases that deliver nullish factors, and I think it is obvious that nullish factors do not exist as keys in the resource.

#### Checklist

- [ ] only relevant code is changed (make a diff before you submit the PR)
- [x] tests are included

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -2,6 +2,7 @@ import baseLogger from './logger.js';
 import EventEmitter from './EventEmitter.js';
 import postProcessor from './postProcessor.js';
 import * as utils from './utils.js';
+import { isOptional } from './utils.js';
 
 const checkedLoadedFor = {};
 
@@ -39,6 +40,10 @@ class Translator extends EventEmitter {
   }
 
   exists(key, options = { interpolation: {} }) {
+    if (isOptional(key)) {
+      return false;
+    }
+
     const resolved = this.resolve(key, options);
     return resolved && resolved.res !== undefined;
   }
@@ -84,7 +89,7 @@ class Translator extends EventEmitter {
     if (!options) options = {};
 
     // non valid keys handling
-    if (keys === undefined || keys === null /* || keys === ''*/) return '';
+    if (isOptional(keys) /* || keys === ''*/) return '';
     if (!Array.isArray(keys)) keys = [String(keys)];
 
     // separators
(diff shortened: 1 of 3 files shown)


jamuhl:
wouldn't it be cleaner to add a check in the exists function itself than in the reused resolve? Like it's already done for the translate here: https://github.com/i18next/i18next/blob/master/src/Translator.js#L87

raccoonback:
@jamuhl 

I like your opinion, 
but wouldn't it be better to deal with common within the `resolve() function` because this error is from the `resolve() function`?

because `nullable values(undefined or null, or '')` can be passed to `keys parameter` elsewhere using the resolve() function`.

jamuhl:
resolve is internal...currently translate, exist are "public" functions - as translate handles the return value for nullish already it seems logical for me to let exists handle it itself too

raccoonback:
@jamuhl 
Yes, I understand. I'll change it.

raccoonback:
@jamuhl 
I've reflected all the reviews. Please check it again.

src/Translator.js
```diff
@@ -2,6 +2,7 @@ import baseLogger from './logger.js';
 import EventEmitter from './EventEmitter.js';
 import postProcessor from './postProcessor.js';
 import * as utils from './utils.js';
+import { isOptional } from './utils.js';
```
jamuhl, line 5: does not need an additional import -> can be accessed as utils.isOptional

personally I wouldn't add a helper in utils for that but have those written inline:

translate -> `if (keys === undefined || keys === null) return '';`
exists -> `if (keys === undefined || keys === null) return false;`


having a reusable helper function makes no sense -> it's not usable in other places, isOptional is non-descriptive if not knowing where it is used...

jamuhl (changes requested):

src/Translator.js
raccoonback: Yes, I'll change it to inline.

raccoonback:
@jamuhl 
I've reflected all the your review. Please check it one more. 😃

adrai:
lgtm

jamuhl (approved):
lgtm

PR #1650, opened 2021-09-15
Updated all packages related to tests

Elindorath:
I made this PR only to update all packages related to tests. This should allow the use of new features (like sinon's `spy.calledOnceWith(arg)`) to simplify and clarify the tests reading and writing. Only 2 changes were required in test files to adapt to the new APIs.

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [ ] ~~tests are included~~

#### Checklist (for documentation change)

- [ ] ~~only relevant documentation part is changed (make a diff before you submit the PR)~~
- [x] motivation/reason is provided

PR #1651, opened 2021-09-16
Updated some dev dependencies

Elindorath:
Following #1650, I made this PR to update all packages related to karma and pre-commit hooks (except prettier which I will update in separated PR because it will change a lot of files).
I had to update `i18next-xhr-backend` because the compiled source in v1.3.0 included a bug preventing `sinon` to correctly mock `XHR`. Since I don't really know what this `v1.11.1.compat.js` file is all about, I'm not sure if the version fixed to v1.3.0 was intentional.
The same goes for `i18next-browser-languagedetector` and `i18next-localstorage-cache`, I left their version untouched while waiting to know if it's okay to do so.
I also updated coverage related packages. It looks good for me but I might be missing something here. The same goes for `husky` and `lint-staged`.

#### Checklist

- [ ] ~~tests are included~~

- [ ] ~~only relevant documentation part is changed (make a diff before you submit the PR)~~


adrai:
imo i18next-xhr-backend can also be replaced by i18next-http-backend...
but we need to wait for @jamuhl's advice

jamuhl:
as the xhr-backend was deprecated it could be replaced for sure (it that helps)

adrai:
@Elindorath do you not like to replace i18next-xhr-backend with i18next-http-backend?

Elindorath:
Could you go a little bit more in depth of the `v1.11.1.compat.js` file? Based on its name (but I'm worried about potential tests added afterwards its creation, top comment says v1.10.2), is it still relevant to keep backward compatibility tests for a version 6 years old?

If not, removing it would enable the removal of those 4 test dependencies: 
- `i18next-xhr-backend`
- `i18next-localstorage-cache`
- `i18next-sprintf-postprocessor`
- `i18next-browser-languagedetector`

Otherwise, this PR seems good to be merged

adrai:
> Believe it or not but by the time of writing, the v11.x.x can be dropped in as a replacement for v1 by just adding a minimal compatibility layer. It's important to us not breaking things just for the sake of using new fancy stuff (which does not mean we do not keep up with latest possibilities of development).

https://www.i18next.com/overview/comparison-to-others#sustainability

@jamuhl needs to decide if these tests can be deeted or not

jamuhl:
if possible I would prefer to keep them...just one more safety net

Elindorath:
You merged it too quickly, I just updated other i18next dependencies 😉

adrai:
sorry, you cann add it here if you want to: #1652

Elindorath:
I made it in a separate PR #1655

PR #1652, opened 2021-09-16
Update prettier

Elindorath:
Following #1651, this is "just" the update of `prettier`. This PR should be merged after #1651

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [ ] ~~tests are included~~

#### Checklist (for documentation change)

- [ ] ~~only relevant documentation part is changed (make a diff before you submit the PR)~~
- [x] motivation/reason is provided

PR #1648, opened 2021-09-13
[WIP] Use `Intl` API in PluralResolver

Elindorath:
I open this PR in a very early stage to have a place where we can discuss if/how i18next should rely on the `Intl` API to handle plural forms. The goal is to solve #1647.

#### Checklist

- [ ] only relevant code is changed (make a diff before you submit the PR)
- [ ] run tests `npm run test`
- [ ] tests are included

#### Checklist (for documentation change)

- [ ] only relevant documentation part is changed (make a diff before you submit the PR)
- [ ] motivation/reason is provided

PR #1653, opened 2021-09-16
Use `Intl` API in PluralResolver

### LINKED ISSUE #1647: Plural forms inaccuracy
## 🐛 Bug Report

During the process of translating my application into Polish, I noticed that the Polish plural forms seem incomplete: "1.5 months" should be translated "1.5 miesiąca", not "1.5 miesiące" as it is currently.
Looking into the code, based on [this comment](https://github.com/i18next/i18next/blob/master/src/PluralResolver.js#L3), I saw that i18next use Translate House's solution. And indeed, they only support basic plural forms, considering only the integer part of the evaluated quantity.

## To Reproduce

Here is a basic [JSFiddle](https://jsfiddle.net/2s7abn0y/20/) to illustrate

## Expected behavior

i18next should be able to handle more complex scenarios regarding plural forms. The [CLDR (Common Locale Data Repository)](https://github.com/unicode-org/cldr), edited by the Unicode Consortium, seems to me to be a better approach, based on a real standard used by the majority of tools in the ecosystem, and with the assurance of its good maintenance over time (at the time of writing, [Translate House's plural forms](https://github.com/translate/l10n-guide/blob/master/docs/l10n/pluralforms.rst) are 5 years old, [Unicode's one](https://github.com/unicode-org/cldr/blob/master/common/supplemental/plurals.xml) are only 27 days old).
See how they handle [Croatian](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#hr) or [Filipino](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#fil) plural forms.
They maintain [this repo](https://github.com/unicode-org/cldr-json) that contains all data in JSON format, which could simplify the automatic generation of the js file.
I came across [this repo](https://github.com/nodeca/plurals-cldr) that does exactly that. I don't know if it is usable as it is, but it could serve as an example.

## Your Environment

- *runtime version*: node v14
- *i18next version*: 20.6.1
- *os*: Mac

Elindorath:
This fixes #1647.
This is the following of #1648 which was closed by inadvertance. I'm unable to re-open it so sorry for the inconvenience.
This PR needs #1651 and #1652 to be merged before.
For futur readers, the goal here is to use the standard `Intl` API to handle plural forms. This leads to a new JSON format (v4) closer than what Unicode uses to classify plural forms. This also means that there is no more such a concept of "simple plural" (the `simplifyPluralSuffix` option is deprecated)

#### Checklist

- [x] tests are included


Elindorath:
I noticed coveralls reported a test coverage decrease. Yet the coveralls status check is now green, although it indicates a lower percentage than the master branch.
Not sure if I'm right but I think it is due to the fact of having put coveralls in the list of reporters. Then it messes up coveralls when executing test suites in local. I think reporting to coveralls should better be performed only on CI.

Elindorath:
I think this might be considered done.

Regarding documentation, do you want me to make a PR on [i18next-gitbook](https://github.com/i18next/i18next-gitbook/) to edit [the configurations options page](https://github.com/i18next/i18next-gitbook/blob/master/overview/configuration-options.md) and [the json format page](https://github.com/i18next/i18next-gitbook/blob/master/misc/json-format.md)?

adrai:
> Regarding documentation, do you want me to make a PR on i18next-gitbook to edit the configurations options page and the json format page?
If you like, yes.

adrai:
@Elindorath Just added also ordinal support ;-) https://github.com/i18next/i18next/commit/a8f335a75979a4d00c793dd5ff610f42d812cf71

PR #1654, opened 2021-09-19
introduce new i18next.resolvedLanguage property

adrai:
It can be used as primary used language, for example in a language switcher.

#### Checklist

- [x] tests are included


jamuhl:
lgtm

PR #1681, opened 2021-10-24
Add named exports for createInstance

perrin4869:
Updating [`i18next-fetch-backend`](https://github.com/dotcore64/i18next-fetch-backend) to be an esm-first module, I found that I could not do `import { createInstance } from "i18next"` like I [expected](https://github.com/dotcore64/i18next-fetch-backend/blob/24ea9e653e218090210f73bce3bfce7455b7c560/test/index.js#L9).
This PR adds support for this.


perrin4869:
Solved the merge conflicts and updated to reflect #1682. I actually agree with @jamuhl and don't particularly like #1682, originally all I wanted was to be able to `import { createInstance } from 'i18next'` XD
but since that one was merged, I went ahead and made the ESM module behave like the `cjs`. I also moved the binding outside of the constructor, so that only the members of the exported `i18next` instance are bound... having all functions bound to a member is just not intuitive...

adrai:
> outside of the constructor

This will probably not automatically work for cloneInstance and createInstance now: https://github.com/i18next/i18next/blob/master/src/i18next.js#L569

perrin4869:
Hm... I think there would be no problem calling those 2...
Nevertheless, it doesn't make much sense to bind cloneInstance to begin with - it's right there in the name :P
And createInstance i changed into a static method and it is bound separately.

adrai:
I mean the bindMemberFunctions call is probably also needed to be called here: https://github.com/i18next/i18next/pull/1681/files#diff-6859f2a371e0f563ec9c27c5cf64cb82f8d91750467509caaa8f6f0e6d06dec5R565

perrin4869:
Ah, i see what you mean! I did that on purpose, i think binding on the default export instance is enough...
Do you think it's better to bind all instances?
I don't like the idea cause it could lead to unexpected or unintuitive behavior...

adrai:
what do you think @joshkel ?

package.json
```diff
@@ -5,6 +5,13 @@
   "main": "./dist/cjs/i18next.js",
   "module": "./dist/esm/i18next.js",
   "types": "./index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/esm/i18next.mjs",
```
adrai, line 10: this is probably not .mjs

rollup.config.js
```diff
@@ -18,14 +18,17 @@ const external = (id) => !id.startsWith('.') && !id.startsWith('/') && !id.inclu
...
     plugins: [babel(getBabelOptions({ useESModules: false }))],
   },
 
   {
-    input,
-    output: { format: 'esm', file: pkg.module },
+    input: './src/index.esm.js',
+    output: [
+      { format: 'esm', file: pkg.module },
+      { format: 'esm', file: pkg.exports['.'].import },
```
adrai, line 30: why is this additional output needed?

src/index.esm.js
```diff
@@ -0,0 +1,18 @@
...
+export const reloadResources = i18next.reloadResources;
+export const use = i18next.use;
+export const changeLanguage = i18next.changeLanguage;
+export const getFixedT = i18next.getFixedT;
+export const t = i18next.t;
+export const exists = i18next.exists;
+export const setDefaultNamespace = i18next.setDefaultNamespace;
+export const hasLoadedNamespace = i18next.hasLoadedNamespace;
+export const loadNamespaces = i18next.loadNamespaces;
+export const loadLanguages = i18next.loadLanguages;
```
adrai, line 18: and clone Instance? it may be there are also other functions?

rollup.config.js
perrin4869: This is the `mjs` output above :)
At first I wanted to add a package.json file, but since `dist` is purged before every build, the package.json was being nuked along with it, so it was tricky to handle... using `mjs` was easier

package.json
perrin4869: I added it to rollup

src/index.esm.js
perrin4869: I checked all the functions that are bound, these were the ones that made the most sense to export...
I did stop and consider `cloneInstance` as well, but I didn't add it because it is really weird to call `cloneInstance` without the actual instance...

rollup.config.js
adrai: But why is the mjs file needed at all?

perrin4869: Well, we either need `mjs` or `package.json` so `node` knows to load it as an es module... is `package.json` better?

adrai: isn't this sufficient? https://github.com/i18next/i18next/pull/1681/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R6

adrai: ahh you mean the package.json in the dist/esm folder

adrai: Personally, I would do something like this while build: `echo '{\"type\":\"module\"}' > dist/esm/package.json`

perrin4869: sure! let me update

perrin4869: done

perrin4869: ah wait

package.json
adrai: so this is can be removed now, right?

rollup.config.js
perrin4869: now it looks good :)

package.json
```diff
@@ -5,6 +5,13 @@
   "main": "./dist/cjs/i18next.js",
   "module": "./dist/esm/i18next.js",
   "types": "./index.d.ts",
+  "exports": {
```
adrai, line 8: is this whole section still required?

perrin4869: I think so... I guess I could remove the `"."` section, and the `types` are for future-proof `typescript` which will support the types key in there: https://www.typescriptlang.org/docs/handbook/esm-node.html

rollup.config.js
```diff
@@ -18,13 +18,13 @@ const external = (id) => !id.startsWith('.') && !id.startsWith('/') && !id.inclu
 export default [
   {
     input,
```
adrai, line 20: I have an idea to optimize the setup also for Deno...

let's create an index.cjs.js file instead of an index.esm.js file... what do you think? This way the src/index.js would be the correct esm export...

btw: I think the exports: 'defaults' option is not necessary

perrin4869: Sure, I don't know anything about deno, but I can try that no problem!

adrai: please also remove the 'defaults' option here: https://github.com/i18next/i18next/pull/1681/files#diff-6814bf77564b4f1c92f5861e184e28fe217c080a047fefa8b73a728f755ec45cR22

perrin4869: done!

adrai: image

adrai:
lgtm now...
just waiting for @joshkel's feedback for the bind question: https://github.com/i18next/i18next/pull/1681#issuecomment-968376401

joshkel:
For my particular use case, I believe that binding on the default instance would be enough.

I would lean toward binding all instances; that feels more consistent and removes any possibility of error. But I'll defer to whatever others decide. Thank you.

adrai:
> I would lean toward binding all instances; that feels more consistent and removes any possibility of error. But I'll defer to whatever others decide. Thank you.

So let's do it for all instance creations ok @perrin4869 ?

perrin4869:
Done!
I think that some people might not expect the instance functions to be bound, in general people don't bind their instances... don't know of other libs that do this... but let's go with this!

adrai:
thank you for everything you've done @perrin4869 v21.5.0 has just been released

stryaponoff:
I've got the warning after upgrading to v21.5.0 with Yarn 3.1.0:
```
warn Package i18next has been ignored because it contains invalid configuration. Reason: Package subpath './package.json' is not defined by "exports"
```

But it seems to work anyway without any problems.

adrai:
> I've got the warning after upgrading to v21.5.0 with Yarn 3.1.0:
> 
> ```
> warn Package i18next has been ignored because it contains invalid configuration. Reason: Package subpath './package.json' is not defined by "exports"
> ```
> 
> But it seems to work anyway without any problems.

@stryaponoff it should be better with v21.5.1

Issue #1287, opened 2019-07-18
Cannot read property 'translator' of undefined when using t()


dinosaurjoe:
Hello,

Having a bit of an issue with my translations for a typescript/react project that my team is working on. We are using the t function to translate components, but are having issues with importing it properly and are consistently getting errors when compiling. 

At the top of our files we are importing via 

```import i18next from "i18next";```

then deconstructing the t within the render method of the component like so:

```const { t } = i18next```

From there, we are calling a specific translation like so:

```{t('common:hello'}```

Any clue why we would be getting 

```Uncaught TypeError: Cannot read property 'translator' of undefined```

When the application compiles?

When I look into it I see this bit of code that is seemingly causing the issue:

```
I18n.prototype.t = function t() {
    var _translator;

    return this.translator && (_translator = this.translator).translate.apply(_translator, arguments);
  };
```

But do not know how to fix it. Thank you!

**Comment by jamuhl** [comment]
you can't deconstruct a function like this...`t` is not an export on a module...it's a function inside a class expecting being called from the instance like `i18next.t` you would have to bind that function...

**Comment by minodisk** [comment]
This worked well for me:

```
const t = i18next.t.bind(i18next)
```

**Comment by nielsboecker** [comment]
Building on top of @minodisk's answer, I have added this in the same module where I am initializing i18next (I am using `react-i18next`, initializing in a top-level `i18n.ts`):
```typescript
...
export default i18n;

const t = i18n.t.bind(i18n);
export { t };
```

so that from other modules, it can be imported as expected:
```typescript
import { t } from "./i18n";
```

90dy:
```
i18next.t = i18next.t.bind(i18next);
```

PR #1682, opened 2021-10-26
Bind i18n.t

joshkel:
I recently ran into #1287, and based on comments there, I saw that I'm not the only one. (In my case, I was trying to write a lower-level function that could take a `t` function either from `useTranslation` or from a I18n instance.) Since this does seem to be a somewhat common error, and since destructuring `t` or trying to use it as a callback can be useful, I believe it would make sense to bind the `t` function within the I18n constructor (similar to how React class components can bind their callbacks) so that `this` is always defined, regardless of how the `t` function is used.

While working on this PR, I fixed some typos, and I noticed that tests were failing, so I set a fixed timezone to help address that. (I can split these changes into separate PRs if you prefer.)

Thank you.

#### Checklist

- [X] tests are included

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,6 +1,10 @@
 //const istanbul = require( 'browserify-istanbul' );
 
 module.exports = function (karma) {
+  // Ensure consistent conversion of timestamps to date strings (has no effect
+  // on Windows)
+  process.env.TZ = 'UTC';
+
   karma.set({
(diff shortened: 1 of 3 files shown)


joshkel:
@adrai Done. I apologize for the delayed response.

adrai:
Thank you, it's included in v21.4.0

Issue #1703, opened 2021-12-07
`defaultValue`s of pluralized messages are not being sent to `missingKeyHandler` properly


hon2a:
_Please note that there's a similar report in https://github.com/i18next/i18next/issues/1549, but a different aspect is observed there and eventually fixed._

## 🐛 Bug Report

The "save missing" feature doesn't work properly for pluralized messages - it calls the `missingKeyHandler` (or `backendConnector.saveMissing`) once for every plural form, but passes the same default value to all even if there are proper default values per plural form defined in translation call options.

#### Steps

Using the following `init` options

```javascript
{
    fallbackLng: "en",
    saveMissing: true,
    saveMissingPlurals: true,
    missingKeyHandler: (_1, _2, key, defaultValue) => console.log({ key, defaultValue })
  }
```

call

```javascript
t("firstMissing", {
  count: 1,
  defaultValue_one: "first singular",
  defaultValue_other: "first plural"
})
t("secondMissing", {
  count: 10,
  defaultValue_one: "second singular",
  defaultValue_other: "second plural"
})
```

#### Actual

The following is logged:

    {key: "firstMissing_one", defaultValue: "first singular"}
    {key: "firstMissing_other", defaultValue: "first singular"}
    {key: "secondMissing_one", defaultValue: "second plural"}
    {key: "secondMissing_other", defaultValue: "second plural"}

#### Expected

The following should be logged instead:

    {key: "firstMissing_one", defaultValue: "first singular"}
    {key: "firstMissing_other", defaultValue: "first plural"}
    {key: "secondMissing_one", defaultValue: "second singular"}
    {key: "secondMissing_other", defaultValue: "second plural"}

#### Notes

The above seems to be expected in the source code (note the "ignored" comment in [test](https://github.com/i18next/i18next/blob/69aedb61dc09a9bde69986d6e8b91377c039b0ee/test/translator/translator.translate.missing.spec.js#L118)), but without any documentation or explanation.

## To Reproduce

[CodeSandbox demo](https://codesandbox.io/s/i18next-missing-key-pluralization-forked-j0rpu) (open console and observe the logged output)

## Expected behavior

The pluralized default values, when provided (e.g. `{ default_one: 'singular', default_other: 'plural' }`), should each be supplied to the missing key handling logic with the respective keys (see _Expected_ section above). There's no sensible reason to use the "currently used" default (based on `count`) for _all_ the plural forms.

## Your Environment

[...]

PR #1704, opened 2021-12-07
fix-1703

hon2a:
Fixed `Translator` to supply correct defaults for plural forms (if supplied) to `saveMissing`. (Fixes https://github.com/i18next/i18next/issues/1703 .)

#### Checklist

- [x] tests are included (or rather the existing test is modified)


hon2a:
The failed CodeClimate check has nothing to do with this patch (it's on unchanged code).

Issue #1220, opened 2019-03-17
How to add 0 plural option for English


brightpixels:
Hello,
This works as expected for the English language:

````
{
  "photos": "one photo",
  "photos_plural": "{{count}} photos"
}
````

But I would like to also add a phrase when count is 0 like this

````
{
"photos_0": "No photos",
  "photos": "one photo",
  "photos_plural": "{{count}} photos"
}
````

Above doesn't work because I read somewhere else you said it depends on the language rules. So for Russian above will work but this isn't possible for English. Is there a way to over ride this with something custom? At the moment to deal with this I am using if-else block to check if count is 0 so I can use the translation when count is 0.

**Comment by jamuhl** [comment]
Yes. English has no extra form for zero - that's why you can't define photos_0 for english.

But there is a plugin to help with this: https://github.com/i18next/i18next-intervalPlural-postProcessor

Just define in the interval the extra form for zero - all others not getting handled by a interval will automatically fallback to regular plural forms.

**Comment by brightpixels** [comment]
Thank you

**Comment by jamuhl** [comment]
If you like this module don’t forget to star this repo. Make a tweet, share the word or have a look at our https://locize.com to support the devs of this project -> there are many ways to help this project :pray:

**Comment by AleksandrChernyavenko** [comment]
We can use [Context](https://www.i18next.com/translation-function/context) for this purpose. Example: 


```json
{
  "photos_0": "one photo",
  "photos": "one photo",
  "photos_plural": "{{count}} photos"
}
```

```js
const count = 0;
i18next.t('photos', {count: count, context: `${count}`})
```

kg-currenxie:
We can't we add `_zero`? It should be a super common case (in most languages?), and we can skip adding `, context: `${count}``

```
{
  "photos_zero": "no photos",
  "photos": "one photo",
  "photos_plural": "{{count}} photos"
}

const count = 0;
i18next.t('photos', { count: count })
```

Straight from RubyOnRails ;) https://guides.rubyonrails.org/i18n.html#pluralization

jamuhl:
@kg-currenxie `photos_zero` is something you like to express -> but by definition it is not a pluralform

so there is the `context` option or https://github.com/i18next/i18next-intervalPlural-postProcessor/ to express more custom ranges

kg-currenxie:
yes, i didn't say its a plural form
by definition, its "nothing", and i see it independent from language grammar
its a logical one imo :)

jamuhl:
[...]

PR #1705, opened 2021-12-07
Add _zero suffix lookup if count === 0

kgominyuk:
Localizing in a specific way 0 count is a very common case.
There was a related [issue](https://github.com/i18next/i18next/issues/1220).
Also unicode documentation says: _"implementations are encouraged to provide the ability to have special plural messages for 0 in particular, so that more natural language can be used"_ https://cldr.unicode.org/index/cldr-spec/plural-rules

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -394,6 +394,11 @@ class Translator extends EventEmitter {
       if (this.options.fallbackNS) namespaces = namespaces.concat(this.options.fallbackNS);
 
       const needsPluralHandling = options.count !== undefined && typeof options.count !== 'string';
+      const needsZeroSuffixLookup =
+        needsPluralHandling &&
+        !options.ordinal &&
+        options.count === 0 &&
+        this.pluralResolver.shouldUseIntlApi();
       const needsContextHandling =
         options.context !== undefined &&
         (typeof options.context === 'string' || typeof options.context === 'number') &&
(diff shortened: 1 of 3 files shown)


adrai:
Currently it's what the Intl plural resolves to...

I don't know if we should introduce a special zero handling...

What do you think @jamuhl?

kgominyuk:
This approach is very common:
RubyOnRails - https://guides.rubyonrails.org/i18n.html#pluralization
Apple - https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals

jamuhl:
For me ok...as it's a rather common use case. But I think the check should add `&& (pluralSuffix === 'one' || pluralSuffix === 'other')` to avoid using it with v3 and using it on languages like chinese where the only used form is other

kgominyuk:
I added `this.pluralResolver.shouldUseIntlApi()` into condition to avoid using it with v3.
Adding `&& (pluralSuffix === 'one' || pluralSuffix === 'other')` won't prevent zero lookup for chinese. 

Why you think that this is irrelevant for chinese? 

The goal of this change is to provide the ability to use more natural language: 
`None of your friends are online.` instead of `You have 0 friends online.`

This is also actual for chinese.

jamuhl:
It's a guess that chinese as it has only one plural form does not have a thing like zero, one, other at all...but idk...I'm not capable of speaking any asian language

kgominyuk:
Also want to mention that this pr isn't introduce requirement to add `_zero` entries for all localized keys.
If the count is zero, and a `_zero` entry is present, then it will be used. 
If `_zero` entry isn't present, original plural suffix will be used.

adrai:
needs to be documented somewhere

adrai:
Thank you for your contribution, it's included in v21.6.0

adrai:
> Is there a reason for only supporting the default plural separator?

v21.8.12 should respect the plural separator

Issue #1763, opened 2022-05-05
Get language of translation for a certain key


c7hm4r:
## 🚀 Feature Proposal

Expose a public API which does the same like the `t` function but also returns the language of the returned translation text.

## Motivation

To recap: For a key, `i18next` dynamically chooses the translation language depending on the configuration of the language resources priority (i.e. `i18n.languages`) and the existence of the key in the specified resources. Thus, when one does `translation = t("some_key")` (usually) a string is returned whose language depends on different factors.

Depending on the language of the returned `translation` string, the code should behave differently. E.g. it should set for each paragraph the right `lang` HTML attribute. However, `t("some_key")` can currently not be configured with options to also return the language code. There are two different approaches how to get the language of a translation by a key:

[Credits – a StackOverflow question](https://stackoverflow.com/questions/72120443/how-to-obtain-the-language-of-the-string-returned-by-i18next-t/)

### Current approach 1: Reimplement the wheel

```typescript
const language = i18n.languages.find((language) => i18n.exists("some_key", {lng: language}));
```

This approach is bad because it reimplements something which happens already inside the `t` function call. So it consumes unnecessary computation time. Furthermore, it could be in the future that the internal implementation of the `t` function changes such that the behavior is different. This approach is thus neither robust nor maintainable. It is incompatible with the `i18n.getFixedT` API.

### Current approach 2: Exploit internal API

```typescript
const { res,  usedLng } = (i18n as any).translator.resolve("some_key") as { res: unknown, usedLng: string };
```

Not maintainable because it uses an internal API. No TypeScript support (`any` and `as` turn the TypeScript compiler off and on). Does not provide the text in the same way `t` does (it seems to not respect `returnObjects` and does not automatically switch to the key as last fallback).

### Current approach 3: Bloat all resources

1. Configure `returnObjects: true`.
2. Replace each translation entry `"the text"` in all resources (`"en"`) by `{ language: "en", text: "the text" }` programmatically.
3. Use 

    ```typescript
    const tResult = t("some_key") as string | { language: string, text: string };
    const { language, text } = typeof tResult === "string" ? { language: "unknown", text: tResult } : tResult;
    ```
Bloat; several tedious corner cases; no TypeScript support.

## Example

```typescript
const { text, usedLanguage } = i18n.resolve("some_key");
return `<p lang="${usedLanguage}">${text}</p>`;
```

jamuhl:
I would add a `verbose` flag to t-options like:

[...]

PR #1764, opened 2022-05-05
introduce returnResolved t option that pratically exposes the result of the resolve function

adrai:
usage: 

```js
i18next.t('key', { returnResolved: true })
// returns {
//   usedKey: 'key',
//   res: 'translation value,
//   exactUsedKey: 'key',
//   usedLng: 'en',
//   usedNS: 'translation',
// }
```

While this PR technically works to help in use cases like #1763 I'm not sure if we should offer this kind of interface...

Normally the lang HTML attribute is set to the [i18next.resolvedLanguage](https://www.i18next.com/overview/api#resolvedlanguage) value.

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -690,10 +690,25 @@ export interface WithT {
  * Object returned from t() function when passed returnDetails: true option.
  */
 export type TFunctionDetailedResult<T = string> = {
+  /**
+   * The plain used key
+   */
   usedKey: string;
+  /**
+   * The translation result.
+   */
   res: T;
+  /**
+   * The key with context / plural
+   */
   exactUsedKey: string;
+  /**
+   * The used language for this translation.
+   */
   usedLng: string;
+  /**
+   * The used namespace for this translation.
+   */
   usedNS: string;
 };
 export type TFunctionResult =
--- a/index.d.ts
+++ b/index.d.ts
@@ -689,7 +689,7 @@ export interface WithT {
 /**
  * Object returned from t() function when passed returnDetails: true option.
  */
-export type TFunctionResolvedResult<T = string> = {
+export type TFunctionDetailedResult<T = string> = {
   usedKey: string;
   res: T;
   exactUsedKey: string;
@@ -699,7 +699,7 @@ export type TFunctionResolvedResult<T = string> = {
 export type TFunctionResult =
   | string
   | object
-  | TFunctionResolvedResult
+  | TFunctionDetailedResult
   | Array<string | object>
   | undefined
   | null;
@@ -714,7 +714,7 @@ export interface TFunction {
     key: TKeys | TKeys[],
   ): TResult;
   <
-    TResult extends TFunctionResult = TFunctionResolvedResult<object>,
+    TResult extends TFunctionResult = TFunctionDetailedResult<object>,
     TKeys extends TFunctionKeys = string,
     TInterpolationMap extends object = StringMap,
   >(
@@ -722,7 +722,7 @@ export interface TFunction {
     options?: TOptions<TInterpolationMap> & { returnDetails: true; returnObjects: true },
   ): TResult;
   <
-    TResult extends TFunctionResult = TFunctionResolvedResult,
+    TResult extends TFunctionResult = TFunctionDetailedResult,
     TKeys extends TFunctionKeys = string,
     TInterpolationMap extends object = StringMap,
   >(
--- a/index.d.ts
+++ b/index.d.ts
@@ -686,7 +686,23 @@ export interface WithT {
   t: TFunction;
 }
 
-export type TFunctionResult = string | object | Array<string | object> | undefined | null;
+/**
+ * Object returned from t() function when passed returnResolved: true option.
+ */
+export type TFunctionResolvedResult = {
+  usedKey: string;
+  res: TFunctionResult;
+  exactUsedKey: string;
+  usedLng: string;
+  usedNS: string;
+};
+export type TFunctionResult =
+  | string
+  | object
+  | TFunctionResolvedResult
+  | Array<string | object>
+  | undefined
+  | null;
 export type TFunctionKeys = string | TemplateStringsArray;
 export interface TFunction {
   // basic usage
--- a/index.d.ts
+++ b/index.d.ts
@@ -655,7 +655,7 @@ export interface TOptionsBase {
   /**
    * Returning an object that includes information about the used language, namespace, key and value
    */
-  // returnResolved?: boolean;
+  returnResolved?: boolean;
 }
 
 /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -689,9 +689,9 @@ export interface WithT {
 /**
  * Object returned from t() function when passed returnResolved: true option.
  */
-export type TFunctionResolvedResult = {
+export type TFunctionResolvedResult<T = string> = {
   usedKey: string;
-  res: TFunctionResult;
+  res: T;
   exactUsedKey: string;
   usedLng: string;
   usedNS: string;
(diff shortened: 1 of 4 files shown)


adrai:
@pedrodurek ok for the types? Or is there a clash when using also `returnObjects: true` ?

pedrodurek:
hey @adrai, it's good enough 😄 , nice job!

adrai:
@pedrodurek can't we change the return type based on the returnResolved option?

Something like:
```ts
/**
 * Object returned from t() function when passed returnResolved: true option.
 */
export type ResolvedResult = {
  usedKey: string;
  res: TFunctionResult;
  exactUsedKey: string,
  usedLng: string,
  usedNS: string,
};
export type TFunctionResult = string | object | ResolvedResult | Array<string | object> | undefined | null;
export type TFunctionKeys = string | TemplateStringsArray;
export interface TFunction {
  // basic usage
  <
    TReturnResolved extends boolean | undefined,
    TResult extends TFunctionResult = TReturnResolved extends true ? ResolvedResult : string,
    TKeys extends TFunctionKeys = string,
    TInterpolationMap extends object = StringMap & { returnResolved?: TReturnResolved },
  >(
    key: TKeys | TKeys[],
    options?: TOptions<TInterpolationMap> | string,
  ): TResult;
```

This will for sure not work, because I've no clue of TypeScript, but just to give you the idea of what I'm asking for...

adrai:
I think I got it: https://github.com/i18next/i18next/pull/1764/commits/cf5bc34e1e8d3e6c94588ef93619a53f9e665595

adrai:
@pedrodurek I've also tried with the returnObjects case: https://github.com/i18next/i18next/pull/1764/commits/5eaec77452d0c336759df3b7e2a56b0e17d11f7e

Can you check if this is ok or if I messed up everything?

src/Translator.js
```diff
@@ -113,9 +113,17 @@ class Translator extends EventEmitter {
     if (lng && lng.toLowerCase() === 'cimode') {
       if (appendNamespaceToCIMode) {
         const nsSeparator = options.nsSeparator || this.options.nsSeparator;
+        if (options.returnResolved) {
+          resolved.res = namespace + nsSeparator + key;
```
c7hm4r, line 117: Maybe use a template string here. That could be a little bit faster.

src/Translator.js
```diff
@@ -113,9 +113,17 @@ class Translator extends EventEmitter {
...
+          return resolved;
+        }
         return namespace + nsSeparator + key;
```
c7hm4r, line 120: … and here

c7hm4r (approved):

c7hm4r:
Thank you for your proposal! Looks great!

1. I wonder whether it could be of interest for someone to configure `returnResolved` in the `i18next` options.
2. `returnResolved` – it is not clear what "resolved" means. It would be the first occurrence of this term in the API with that meaning. Maybe name it `returnDetails`?

index.d.ts
```diff
@@ -713,6 +713,14 @@ export interface TFunction {
   >(
     key: TKeys | TKeys[],
   ): TResult;
+  <
+    TResult extends TFunctionResult = TFunctionResolvedResult<object>,
+    TKeys extends TFunctionKeys = string,
+    TInterpolationMap extends object = StringMap,
+  >(
+    key: TKeys | TKeys[],
+    options?: TOptions<TInterpolationMap> & { returnResolved: true; returnObjects: true },
```
c7hm4r, line 722: I think that will break the typing, when someone configures `returnObjects` in the `i18next` global options.

c7hm4r:
Is it already possible to configure `returnResolved: true` in the `i18next` global options? That would probably break the current TypeScript typing.

index.d.ts
adrai: I'm not a TypeScript user... any advice on how to change this, if this is a problem?

adrai:
> Is it already possible to configure `returnResolved: true` in the `i18next` global options? That would probably break the current TypeScript typing.

no, this is not a global option yet... I don't think this makes sense to make this a global option, right?

c7hm4r:
> > Is it already possible to configure `returnResolved: true` in the `i18next` global options? That would probably break the current TypeScript typing.
> 
> no, this is not a global option yet... I don't think this makes sense to make this a global option, right?

I'm a bit inclined to say that it could make sense, but I'm probably biased.

index.d.ts
c7hm4r: Sorry, I think I was wrong with that. The current typing seems good to me.

adrai:
> > > Is it already possible to configure `returnResolved: true` in the `i18next` global options? That would probably break the current TypeScript typing.
> > 
> > 
> > no, this is not a global option yet... I don't think this makes sense to make this a global option, right?
> 
> I'm a bit inclined to say that it could make sense, but I'm probably biased.

ok, now it is... for JS there is for sure no problem... but I have no idea about TS

c7hm4r:
> returnResolved – it is not clear what "resolved" means. It would be the first occurrence of this term in the API with that meaning. Maybe name it returnDetails?

`verbose` is also not bad for someone who is familar with UNIX CLI tools.

c7hm4r:
> ok, now it is... for JS there is for sure no problem... but I have no idea about TS

thank you

adrai:
regarding the name: I'm not bound to returnResolved, I just took that, because internally the translation key needs to be resolved and the resulting information is exactly what we return...
But I can rename this to verbose or returnDetails or returnMeta or returnInfo or whatever you like....

c7hm4r:
> `verbose` is also not bad for someone who is familar with UNIX CLI tools.

Ok, after thinking a bit about  "verbose", that could be a bit confusing considering that this is a localization/translation library. :thinking:

c7hm4r:
I think my favorite name is `returnDetails`, which is exactly what this option does. `returnResolved` does not imply that `t` would return more details; when `t` returns only the translation, the translation could also be seen as something which was "resolved", so for someone else it could be unclear what `returnResolved: false` does.

adrai:
renamed

index.d.ts
```diff
@@ -682,10 +686,57 @@ export interface WithT {
   t: TFunction;
 }
 
-export type TFunctionResult = string | object | Array<string | object> | undefined | null;
+/**
+ * Object returned from t() function when passed returnDetails: true option.
+ */
+export type TFunctionResolvedResult<T = string> = {
```
c7hm4r, line 692: Rename to `TFunctionDetailsResult`?

c7hm4r (changes requested):

index.d.ts
c7hm4r: `TFunctionDetailedResult` would be better…

c7hm4r:
> renamed

:heart_eyes:

index.d.ts
adrai: done

c7hm4r:
thank you, I like it

adrai:
Do you think it is ready to be merged?

c7hm4r:
> Do you think it is ready to be merged?

Yes. `TFunctionDetailedResult` is undocumented, but the documentation resides in another repository, right?

adrai:
> doc for TFunctionDetailedResult

Do you mean this? https://github.com/i18next/i18next/pull/1764/commits/c9754be864970616c249405b66faecee0afdd2cc

c7hm4r:
> > Yes. TFunctionDetailedResult is undocumented, but the documentation resides in another repository, right?
> 
> Do you mean this? [c9754be](https://github.com/i18next/i18next/commit/c9754be864970616c249405b66faecee0afdd2cc)

Perfect

c7hm4r:
Thank you very much again!

adrai:
released with v21.7.0

adrai:
If you like this module don’t forget to star this repo. Make a tweet, share the word or have a look at our https://locize.com to support the devs of this project.

There are many ways to help this project 🙏

PR #1771, opened 2022-05-08
Add maxParallelReads option, default 10

### LINKED ISSUE #1769: Unbounded parallelism on BackendConnector.load
## 🐛 Bug Report

If, say, 20 languages and 400 namespaces are passed into `BackendConnector.load` then all 8,000 of those load calls will be dispatched to the backend without waiting for a single one of them to complete.  If 100,000 items are loaded then all of those would be called, and so on.

The effect of this is that backends such as `i18next-fs-backend` will run out of file descriptors when using the default async loader and other backends may try to open tens of thousands of sockets to a host.

This results in very real problems on systems with low `ulimit`'s on file descriptors, such as AWS Lambda which has a 1,024 limit.  The large number of files being opened results in `EMFILE` errors that are not raised to the caller or printed in any way unless `debug` is enabled.

The problem starts here:

https://github.com/i18next/i18next/blob/a0daee7d64b19190d9dc335a764a6e1735dfc7aa/src/BackendConnector.js#L168-L170

`this.loadOne` is called on each item in `toLoad.toLoad`.  `this.loadOne` calls `this.read` which calls `this.backend`.  The problem is that `this.read` is an "async" function with a callback but `this.loadOne` does not "await" that function, it just keeps looping.

## To Reproduce

A [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example).
A [codesandbox](https://codesandbox.io/examples/package/i18next) example or similar
or at least steps to reproduce the behavior:

[test code](https://github.com/i18next/i18next/compare/master...huntharo:unconstrained-parallelism-demo?expand=1)

image

## Expected behavior

Options to address the issue:

1. Document for backend implementors that they must be aware that backends may be called tens of thousands of times and that backends must implement a parallelization limit if that is necessary to avoid reaching socket, file descriptor, etc. limits / document for users that `initImmediate` should be set to false to address resource constraint issues with large numbers of languages and namespaces
2. Implement parallelization constraint within this library - Unfortunately, this is not easy to do without adding a dependency and making other breaking changes

Once the decision is made I can potentially create a PR to update the docs.

## Your Environment

- *runtime version*: any
- *i18next version*: master
- *os*: any

huntharo:
Fixes #1769

#### Checklist

- [x] tests are included
- [x] PR for https://github.com/i18next/i18next-gitbook documenting new option and default value
  - https://github.com/i18next/i18next-gitbook/pull/196

### DIFF
--- a/src/BackendConnector.js
+++ b/src/BackendConnector.js
@@ -21,6 +21,10 @@ class Connector extends EventEmitter {
     this.options = options;
     this.logger = baseLogger.create('backendConnector');
 
+    this.waitingReads = [];
+    this.maxParallelReads = options.maxParallelReads || 10;
+    this.readingCalls = 0;
+
     this.state = {};
     this.queue = [];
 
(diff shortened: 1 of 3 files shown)


adrai:
released with v21.8.0

Issue #1809, opened 2022-07-29
Fix retry logic


mattbjordan:
## 🐛 Bug Report

- Retries of reading (fetching) the same resource are treated as separate parallel processes. 
- 5 retries are required to be attempted before read's callback function is called. (Following the execution path from i18next.init(), the callback function is what allows the promise made by i18next.init() to resolve.)

If multiple resources which do not exist are requested, then maxParallelReads is reached and the deferred promise made within i18next.init() never resolves. 

This functionality exists in the read() function at i18next.js line 2107.

## To Reproduce

```ts
import i18next, { i18n, InitOptions } from "i18next";
import Backend, { BackendOptions } from "i18next-http-backend";

const namespaces: string[] = ["ns1", "ns2", "ns3"];

const backendOptions: BackendOptions = {
  loadPath: "locales/{{lng}}/{{ns}}.json",
  crossDomain: true
};

const initOptions: InitOptions = {
  backend: backendOptions,
  ns: namespaces,
  defaultNS: namespaces[0]
};

var myInstance: i18n = i18next.createInstance();

myInstance.use(Backend);

initPromise = (myInstance.init(initOptions) as unknown) as Promise<void>;

// Gets stuck here
await initPromise;
```

Oddly, it seems that the program does not actually get stuck awaiting... but it certainly is causing hanging in the background. If you trace through the execution in a debugger, you'll see that the callback function passed into read() is never called.

What I see happening in read() though, is that attempted requests for the same resource are added to the waitingReads list. If there are multiple resources which do not exist, then the waitingReads list fills up with maxParallelReads items before the 5th retry of each requested resource in the waitingReads list is tried, thus readingCalls is never decremented and the callback function is never called.

## Suggested behavior

Attempted reads on the same resource should not be added to the waitingReads list OR remove the previous try from the list before performing another attempt and decrement readingCalls.

Additionally, it would be nice for the user to be able to set how many retries to run and what the wait/timeout timeout time should be between retries. This includes the possibility of doing zero retries. 

It would also be nice to set the above per request or per resource. Say I have a default namespace that I add to every i18next instance that I create, yet most of the time there is no JSON file created for it. It would make sense to not retry requesting that resource 5 times because it is likely to not exist, yet if I added another namespaces in a particular instance of my program, then I wouldn't mind it retrying a few times. The default retry logic can add about 10 seconds to i18next initialization if it has to retry 5 times looking for a resource that doesn't exist. It would be nice to be able to change off of that default behavior.

[...]

PR #1812, opened 2022-07-31
More `BackendConnector.load` failure tests

huntharo:
- Confirm that the callback is called even when some namespaces fail completely
- While writing these tests it became very clear that there is an inability to set the retry delay etc of `read` when using `BackendConnector.load`, as noted in #1809
- All of these tests pass with the code as-is :+1:

#### Checklist

- [x] tests are included

### DIFF
--- a/test/backend/backendConnector.load.retry.spec.js
+++ b/test/backend/backendConnector.load.retry.spec.js
@@ -60,7 +60,7 @@ describe('BackendConnector load all fail - 1 namespace', () => {
   });
 });
 
-describe.only('BackendConnector load all fail - 10 namespaces', () => {
+describe('BackendConnector load all fail - 10 namespaces', () => {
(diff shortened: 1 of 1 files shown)


test/backend/backendConnector.load.retry.spec.js
```diff
@@ -33,6 +33,143 @@ describe('BackendConnector load retry', () => {
...
+      connector.load(['en'], ['fail'], function (err) {
+        expect(err).to.eql(['failed loading']);
+        expect(connector.store.getResourceBundle('en', 'fail')).to.eql({});
+        done();
+      });
+    }).timeout(12000);
+  });
+});
+
+describe.only('BackendConnector load all fail - 10 namespaces', () => {
```
adrai, line 63: remove the only please

PR #1828, opened 2022-09-07
FallbackLng function argument should also return false

martinlecs:
This PR fixes a minor type issue where the fallbackLng property from `InitOptions` accepts the following types:

``` 
false | string | readonly string[] | FallbackLngObjList
````

The fallbackLng property also accepts a function that returns a value of type:

```
(code: string) => string | readonly string[] | FallbackLngObjList
```

I'm doing some work in my dev environment where I want to disable the fallbackLng based on some condition ie. I want to extend the function return type to allow us to return `false`.

```
(code: string) => string | readonly string[] | FallbackLngObjList | false
```

This will fix what I think is an inconsistency with the fallbackLng type in InitOptions.

#### Checklist

- [ ] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,7 +9,7 @@ export type FallbackLng =
   | string
   | readonly string[]
   | FallbackLngObjList
-  | ((code: string) => string | readonly string[] | FallbackLngObjList);
+  | ((code: string) => string | readonly string[] | FallbackLngObjList | false);
 
 export type FormatFunction = (
   value: any,

### CODE REVIEW

**Comment by adrai** [comment]
every falsy value (undefined, null)
Where did you see about false?

**Comment by martinlecs** [comment]
@adrai Documentation here shows that you can set `fallbackLang` in InitOptions to `false` if you want to disable triggering of `fallbackLang`: https://www.i18next.com/overview/configuration-options

**Comment by adrai** [comment]
shouldn't the change be like this then?

```
export type FallbackLng =
  | string
  | readonly string[]
  | FallbackLngObjList
  | false,
  | ((code: string) => string | readonly string[] | FallbackLngObjList);
```

**Comment by martinlecs** [comment]
Type definition of fallbackLng on InitOptions which we allow `false`: https://github.com/i18next/i18next/blob/b027522e8e0f30af1beda3560fd62deb57126ce5/index.d.ts#L254

**Comment by adrai** [comment]
So already done, right?

**Comment by martinlecs** [comment]
@adrai False is already set on `fallbackLng`. My issue is that I want the function that we can pass into `fallbackLng` to also allow me to return `false` so that I can do:

```typescript
i18n.init({
  fallbackLng: (code: string) => { if (condition) { return false } return 'en' }
})
```

The current types don't allow me to return `false`, only `string | readonly string[] | FallbackLngObjList`

adrai:
When using the fallbackLng as function you should return an empty string and not false...

adrai:
ok, return false could also work, but not documented.

adrai:
Sorry, needed to revert it.
false is not allowed (works but not intensionally)
You should return an empty string, like: 

```js
i18n.init({
  fallbackLng: (code: string) => { if (condition) { return [] } return 'en' }
})
```

PR #1843, opened 2022-10-05
Add formatter cache

### LINKED ISSUE #1841: Caching Intl formatters
## 🚀 Feature Proposal

This is my next approach to #1776 (I apologize for creating new issue - I couldn't reopen that one and didn't have enough spare time back then to provide a POC).

I have been reading the source code recently and I have noticed that [using formatters creates new intl formatter instances on every usage](https://github.com/i18next/i18next/blob/9ca694db52fe5b6f10da5debf3e106e6f615b907/src/Formatter.js#L44-L59). It has negative effect on performance and could be avoided by caching and reusing already created formatters.

## Motivation

This is rather simple and minor change that should improve formatters performance. Similar improvement has been previously introduced to Angular codebase in this PR: https://github.com/angular/angular/pull/18054/files

If you agree, I would be happy to make a PR with suggested changes.

Here is a simple demo showing ~2x faster execution for 20 occurrences of formatter (I recommend using 6x slowdown in devtools for better results). The difference increases with more usages. 

https://jsfiddle.net/kv4gt3qs/

pwolaq:
This PR resolves #1841. 

Few considerations:
- maybe we should export `createCachedFormatter` as a utility function so that others may reuse it when creating custom formatters?
- I had no idea how to test that functionality apart from somehow stubbing Intl constructors and verifying that they are not called multiple times - I can try to add this test if you would like to have it
- as cache keys are always strings I decided to use object as caching structure instead of Map (which might cause compatibility issues in older browsers)

#### Checklist

- [ ] tests are included

### DIFF
--- a/src/Formatter.js
+++ b/src/Formatter.js
@@ -39,27 +39,45 @@ function parseFormatStr(formatStr) {
   };
 }
 
+function createCachedFormatter(fn) {
+  const cache = Object.create(null);
+  return function (val, lng, options) {
+    const key = lng + JSON.stringify(options);
+    let formatter = cache[key];
+    if (!formatter) {
+      formatter = fn(lng, options);
+      cache[key] = formatter;
+    }
+    return formatter(val);
+  };
+}
+
 class Formatter {
   constructor(options = {}) {
     this.logger = baseLogger.create('formatter');
 
     this.options = options;
     this.formats = {
-      number: (val, lng, options) => {
-        return new Intl.NumberFormat(lng, options).format(val);
-      },
-      currency: (val, lng, options) => {
-        return new Intl.NumberFormat(lng, { ...options, style: 'currency' }).format(val);
-      },
-      datetime: (val, lng, options) => {
-        return new Intl.DateTimeFormat(lng, { ...options }).format(val);
-      },
-      relativetime: (val, lng, options) => {
-        return new Intl.RelativeTimeFormat(lng, { ...options }).format(val, options.range || 'day');
-      },
-      list: (val, lng, options) => {
-        return new Intl.ListFormat(lng, { ...options }).format(val);
-      },
+      number: createCachedFormatter((lng, options) => {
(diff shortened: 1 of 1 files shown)


src/Formatter.js
```diff
@@ -39,27 +39,45 @@ function parseFormatStr(formatStr) {
   };
 }
 
+function createCachedFormatter(fn) {
+  const cache = Object.create(null);
```
adrai, line 43: `const cache = {};` is probably faster.

adrai:
As alternative to the exposure of the `createCachedFormatter` function, we could extend the add function... something like `add(name, fc, getFormatterFc)` and cache that automatically... just an idea before going to bed 😉

src/Formatter.js
pwolaq: no, it is not - `{}` creates object with prototype - when you try to access property it has to go through the whole prototype chain; also if your cache key happens to be one of the Object's properties (like for example `hasOwnProperty`) then it would falsely return cached object (it should not happen here because of key format however)

adrai: https://www.measurethat.net/Benchmarks/Show/13610/0/object-literal-vs-objectcreatenull 🤷‍♂️

adrai:
> > As alternative to the exposure of the `createCachedFormatter` function, we could extend the add function... something like `add(name, fc, getFormatterFc)` and cache that automatically... just an idea before going to bed 😉
> 
> what do you think about something like this?
> 
> ```js
> add(name, fc) {
>   this.formats[name.toLowerCase().trim()] = fc;
> },
> 
> addCached(name, fc) {
>   this.formats[name.toLowerCase().trim()] = createCachedFormatter(fc)
> }
> 
> 
> i18next.services.formatter.addCached('my-format', (lng, options) => {
>   const formatter = new Intl.NumberFormat(lng, options);
>   return (val) => formatter.format(val);
> })
> ```

Will you extend this PR or should I do?

adrai:
landed in v21.10.0

PR #1865, opened 2022-11-11
Change `DefaultTFuncReturn` to return `null` if `returnNull` typeOption is true

pedrodurek:
Closes https://github.com/i18next/react-i18next/issues/1559, https://github.com/i18next/react-i18next/issues/1576, https://github.com/i18next/react-i18next/issues/1574

Since `returnNull` option is `true` by default, `t` function can return `string` or `null`. By declaring the `resources` type, we can infer the exact return type, so this is only useful to people who can't declare the `resource` type .

This PR allows to change this ☝️  behaviour if to set the `returnNull` type to `false`.
```ts
// i18next.d.ts
import 'i18next';

declare module 'i18next' {
  interface CustomTypeOptions {
    returnNull: false;
  }
}
```

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -860,7 +860,7 @@ type NormalizeMultiReturn<T, V> = V extends `${infer N}:${infer R}`
     : never
   : never;
 
-export type DefaultTFuncReturn = string | undefined | null;
+export type DefaultTFuncReturn = string | (TypeOptions['returnNull'] extends true ? null : never);
 
 export type DefaultTFuncReturnWithObject = DefaultTFuncReturn | object | Array<string | object>;
 
--- a/test/typescript/custom-types/i18next.d.ts
+++ b/test/typescript/custom-types/i18next.d.ts
@@ -3,6 +3,7 @@ import 'i18next';
 declare module 'i18next' {
   interface CustomTypeOptions {
     defaultNS: 'custom';
+    returnNull: false;
     resources: {
       custom: {
         foo: 'foo';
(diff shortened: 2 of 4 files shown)


adrai:
tests are failing

adrai:
included in v22.0.5

Jonnboy91:
Might be a dumb question but after this was included, it removed undefined as one of the possible returns of DefaultTFuncReturn, which in some cases gives an error then. For example TextInput requires either string | undefined in the placeholder, and with this update I'm getting this error:

`  Overload 1 of 2, '(props: TextInputProps | Readonly<TextInputProps>): TextInput', gave the following error.
    Type 'DefaultTFuncReturn' is not assignable to type 'string | undefined'.
      Type 'null' is not assignable to type 'string | undefined'.
`

This can be fixed by setting customTypeOptions returnNull: false, but was just wondering is that the preferred way instead of having also undefined as an option 🤔

Jonnboy91:
Hi @pedrodurek Thanks, this is what I thought and meant by "This can be fixed by setting customTypeOptions returnNull: false", but yeah it used to have the option of undefined and now that it was deleted some places where translation was used were giving errors now, but I could fix it by setting this customTypeOptions, so this was more of a question is the undefined coming back or will I just use this 💪 😄

adrai:
> @adrai can correct if I'm wrong, but `t` function will never return `undefined`, that's why I removed it from the union.

Yes, I'm not aware of any use case returning undefined by purpose...

PR #1870, opened 2022-11-22
Update global t function TypeScript type to support DefaultNamespace

### LINKED ISSUE #1855: defaultNS with Typescript
## 🐛 Bug Report

It seems that `i18next` doesn't respect type of defaultNS in .d.ts file while using `t` function

## To Reproduce

Code example:

#### i18next.d.ts
```ts
import { resources, defaultNS } from '../src/i18next';

declare module 'i18next' {
    interface CustomTypeOptions {
        defaultNS: typeof defaultNS;
        resources: typeof resources['en'];
    }
}
```
#### i18next.ts
```ts
import en from './translations/en.json';

export const defaultNS = 'general';

export const resources = {
    en: {
        general: en,
    },
} as const;

i18n.use(initReactI18next)
    .init({
        resources,
        fallbackLng: 'en',
        ns: ['general'],
        defaultNS,
        compatibilityJSON: 'v3',
        interpolation: {
            escapeValue: false,
            skipOnVariables: false,
        },
        react: {
            useSuspense: false,
        },
    });
```
#### Use cases
```ts
// i18next
i18next.t('common.errors.required') // typescript error
i18next.t('general:common.errors.required') // valid

// react-i18next
t('common.errors.required'); // valid
```

## Expected behavior

```ts
// i18next
i18next.t('common.errors.required') // valid
i18next.t('general:common.errors.required') // valid
```

## Your Environment

- *runtime version*: react-native 0.70.3
- *i18next version*: 12.0.0
- *react-i18next version*: 22.0.2
- *os*: Mac
- *typescript*: 4.8.4

daveallie:
Update TypeScript type of `t` function when used directly from i18n module.

The behaviour of `t` is unchanged, this is simply a change to the types to reflect the underlying logic of `t` (accepting of either a namespace, or no namespace with a fallback to the default namespace).

Fixes #1855

#### Checklist

- [ ] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1130,7 +1130,7 @@ export interface ExistsFunction<
 
 export interface i18n {
   // Expose parameterized t in the i18next interface hierarchy
-  t: TFunction<FallbackOrNS<string>[]>;
+  t: TFunction<Namespace | DefaultNamespace>;
 
   /**
    * The default of the i18next module is an i18next instance ready to be initialized by calling init.


index.d.ts
```diff
@@ -1130,7 +1130,7 @@ export interface ExistsFunction<
 
 export interface i18n {
   // Expose parameterized t in the i18next interface hierarchy
-  t: TFunction<FallbackOrNS<string>[]>;
+  t: TFunction<Namespace | DefaultNamespace>;
```
pedrodurek, line 1133: Hey @daveallie, thank you for your contribution!
Unfortunately, the approach here won't work as expected. `Namespace` will return all namespaces, which means that all namespaces will be considered "Default Namespaces".
Additionally, we should filter the `DefaultNamespace` out if the `DefaultNamespace` is not included in the resources. Otherwise, it'll infer the namespace as `string`, which means that `t` function will accept any key.
Both problems can be easily solved. My biggest concern is regarding the compilation time since we'll iterate over the default translation keys twice. It can be solved but may require major changes in the Normalize type.

adrai:
@daveallie please check again with v22.4.1

PR #1879, opened 2022-12-10
Update `LanguageDetectorAsyncModule`

daomtthuan:
Add return type `Promise<void>` for `LanguageDetectorAsyncModule`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1067,10 +1067,10 @@ export interface LanguageDetectorAsyncModule extends Module {
   type: 'languageDetector';
   /** Set to true to enable async detection */
   async: true;
-  init(services: Services, detectorOptions: object, i18nextOptions: InitOptions): void;
+  init(services: Services, detectorOptions: object, i18nextOptions: InitOptions): void | Promise<void>;
   /** Must call callback passing detected language */
-  detect(callback: (lng: string | readonly string[] | undefined) => void): void;
-  cacheUserLanguage(lng: string): void;
+  detect(callback: (lng: string | readonly string[] | undefined) => void): void | Promise<void>;
+  cacheUserLanguage(lng: string): void | Promise<void>;
 }
 
 /**


index.d.ts
```diff
@@ -1067,10 +1067,10 @@ export interface LanguageDetectorAsyncModule extends Module {
   type: 'languageDetector';
   /** Set to true to enable async detection */
   async: true;
-  init(services: Services, detectorOptions: object, i18nextOptions: InitOptions): void;
+  init(services: Services, detectorOptions: object, i18nextOptions: InitOptions): void | Promise<void>;
```
adrai, line 1070: the init option does not return a promise

index.d.ts
```diff
@@ -1067,10 +1067,10 @@ export interface LanguageDetectorAsyncModule extends Module {
...
   /** Must call callback passing detected language */
-  detect(callback: (lng: string | readonly string[] | undefined) => void): void;
-  cacheUserLanguage(lng: string): void;
+  detect(callback: (lng: string | readonly string[] | undefined) => void): void | Promise<void>;
+  cacheUserLanguage(lng: string): void | Promise<void>;
```
adrai, line 1073: also the cacheUserLanguage does not return a Promise

index.d.ts
```diff
@@ -1067,10 +1067,10 @@ export interface LanguageDetectorAsyncModule extends Module {
   type: 'languageDetector';
   /** Set to true to enable async detection */
   async: true;
-  init(services: Services, detectorOptions: object, i18nextOptions: InitOptions): void;
+  init(services: Services, detectorOptions: object, i18nextOptions: InitOptions): void | Promise<void>;
   /** Must call callback passing detected language */
-  detect(callback: (lng: string | readonly string[] | undefined) => void): void;
-  cacheUserLanguage(lng: string): void;
+  detect(callback: (lng: string | readonly string[] | undefined) => void): void | Promise<void>;
```
adrai, line 1072: and also the detect function does not return a promise (only callback based)

daomtthuan:
@adrai 
Why do`detect` and `cacheUserLanguage` not return `promises` in the module `LanguageDetectorAsyncModule`, although it is `async`?

In this case
When use `LanguageDetector` with `AsyncStorage`
```ts
 const languageDetector: LanguageDetectorAsyncModule = {
    type: 'languageDetector',
    async: true,
    init: () => {
      // Do nothing.
    },
    detect: async (setLanguage) => {
      try {
        const language = await AsyncStorage.getItem(i18nStorageKey.language);

        setLanguage(language ?? 'en');
      } catch (exception) {
        console.error(exception);
      }
    },
    cacheUserLanguage: async (language) => {
      try {
        await AsyncStorage.setItem(i18nStorageKey.language, language);
      } catch (exception) {
        console.error(exception);
      }
    },
  };
```
because use `async/await`, `LanguageDetector` is `LanguageDetectorAsyncModule`, not `LanguageDetectorModule`.

index.d.ts
daomtthuan: When use `LanguageDetector` with `AsyncStorage`, getting language from `AsyncStorage` need detect return promise.
Please check my below comment

index.d.ts
daomtthuan: When use LanguageDetector with AsyncStorage, setting language to AsyncStorage to cache language need `cacheUserLanguage` return promise.
Please check my below comment

index.d.ts
daomtthuan: In my case, `init` no need return `promise`. But I think in some case, it will need.
Because this is module with `async: true`.
That is reason its name is `LanguageDetectorAsyncModule`, not `LanguageDetectorModule`
Please check my below comment

adrai:
Because the interface looks like this: https://www.i18next.com/misc/creating-own-plugins#languagedetector

daomtthuan:
@adrai 
So, can not use async/await for `detect` and `cacheUserLanguage`?

adrai:
This is sufficient:

```js
import AsyncStorage from '@react-native-async-storage/async-storage'

  const LANGUAGE_KEY = 'i18n.language';

  const languageDetector: LanguageDetectorAsyncModule = {
    type: 'languageDetector',
    async: true,
    detect: (setLanguage) => {
      AsyncStorage.getItem(LANGUAGE_KEY)
        .then((language) => setLanguage(language ?? 'en'));
    },
    cacheUserLanguage: (language) => AsyncStorage.setItem(LANGUAGE_KEY, language)
  };

```

That's how the interface of i18next is defined since the beginning...

I'll check later, if at least we can accept a promise for the detect function.... will let you know.

adrai:
ok, with v22.3.0 you can now do this:

```js
  import AsyncStorage from '@react-native-async-storage/async-storage'

  const LANGUAGE_KEY = 'i18n.language';

  const languageDetector: LanguageDetectorAsyncModule = {
    type: 'languageDetector',
    async: true,
    detect: () => AsyncStorage.getItem(LANGUAGE_KEY),
    cacheUserLanguage: (language) => AsyncStorage.setItem(LANGUAGE_KEY, language)
  };
```

daomtthuan:
Thank you <3

Issue #1884, opened 2022-12-13
Consider setting the default for returnNull to false in next major version (v23.0.0)


adrai:
Seems for TypeScript users (in combination with react) it would be better to set the `returnNull` value to `false` by default.

https://github.com/i18next/react-i18next/issues/1559
https://github.com/i18next/react-i18next/issues/1576
https://github.com/i18next/react-i18next/issues/1574
https://github.com/i18next/next-i18next/issues/1795
https://github.com/i18next/next-i18next/issues/2038
https://github.com/i18next/next-i18next/issues/2059
https://stackoverflow.com/questions/70914886/react-i18n-t-function-doesnt-accept-string-variables-typescript-no-over/74817767
https://stackoverflow.com/questions/74146925/im-having-issues-with-i18next-in-react-18/74811575
https://stackoverflow.com/questions/74143213/how-to-handle-i18next-new-type-defaulttfuncreturn/74808715

Let's keep `returnEmptyString` set to `true` by default (not necessary regarding TypeScript issues)

<img width="783" alt="image" src="https://user-images.githubusercontent.com/1086194/207426283-26428633-6554-4ed3-a0e7-3225048ce490.png">

I would say as soon as @pedrodurek reviewed (and maybe corrected/optimized) the latest TypeScript changes and issues, we'll go for a new major release that changes the defaults, this would maybe simplify the usage for TypeScript users.
## TODO before next major release
- [ ] verify t function changes, especially regarding performance: https://github.com/i18next/i18next/compare/v22.3.0...master
- [ ] check if those changes are ok for react-i18next
- [ ] also check if this was ok: https://github.com/i18next/i18next/pull/1878
- [ ] recheck https://github.com/i18next/react-i18next/issues?q=is%3Aissue+is%3Aopen+label%3Atypescript
- [ ] recheck https://github.com/i18next/i18next/issues?q=is%3Aissue+is%3Aopen+label%3Atypescript
- [ ] recheck https://github.com/i18next/next-i18next/issues?q=is%3Aissue+is%3Aopen+label%3Atypescript
- [ ] is there also a solution for https://github.com/i18next/next-13-app-dir-i18next-example-ts/blob/main/app/i18n/index.ts#L26 ?

## TODO
- [x] PR #1885 
- [x] merge PR #1885 
- [ ] docs changes: https://www.i18next.com/overview/configuration-options#translation-defaults and https://www.i18next.com/overview/typescript
- [x] also merge PR #1945 for next major
[...]

PR #1885, opened 2022-12-13
change returnNull default to false

adrai:
#1884


pedrodurek (approved):

adrai:
Let's wait to merge this until you "verified" all ts changes.

mr-july:
@adrai: are you aware of the problem with the codeclima check?

adrai:
> @adrai: are you aware of the problem with the codeclima check?

@mr-july was not related to this PR

pedrodurek:
Hey @adrai, omitting `null` from the return type will only work when the resources type is not assigned (when devs opt not to have the keys inferred). Otherwise, it'll iterate over every key and extract the actual literal value. If you move the test to `test/typescript/t.test.ts` it should work. 
If you prefer, we can also change the types for when the resources type is defined to fallback to `string` if the value is `null`, up to you!

adrai:
i18next will return the key if returnNull is false, so maybe it's more consistent if the type would return string in that case?

adrai:
returnNull false having a key value set to false behaves like a non-existing key... so i18next returns the defaultValue or the key name... type string

adrai:
I've moved the test, like you suggested... I think it's ok to have an error when using typesafe translations... (behaves like non-existing keys)

PR #1903, opened 2023-01-29
types: fix types with moduleResolution node16

### LINKED ISSUE #1900: Incorrect types when using i18next in an esm package with moduleResolution node16
## To Reproduce

https://codesandbox.io/s/vibrant-hellman-fgs4r9?file=/src/index.ts

```ts
import i18n from "i18next";

console.log(
  // This is actually undefined when running in browser or node
  typeof i18n.default,
  typeof i18n.default?.language,
  // ts error
  i18n.language
);
```

tsconfig.json:
```json
{
  "compilerOptions": {
    "strict": true,
    "module": "esnext",
    "esModuleInterop": true,
    "rootDir": "src",
    "moduleResolution": "node16"
  }
}
```

Note: change `moduleResolution` to `node` will make it work as expected.

package.json (It's ES Module):
```json
{
  "type": "module"
}
```

image

## Expected behavior

TS should not report an error for `i18n.language` , and instead, it should for `i18n.default`.

```js
// Paste the expected results here
```

## Your Environment

- *runtime version*: node v16
- *i18next version*: 22.4.9
- *os*: Mac
- *typescript version*: 4.9.4

weareoutman:
closes #1900

#### Checklist

- [ ] tests are included

Note: tests failed (`test:typescript:noninterop`), reporting:

```
ERROR: /Users/wangshenwei/github/i18next/test/typescript/backend.test.ts:1:8 - TypeScript@next compile error: 
Module '"/Users/wangshenwei/github/i18next/index"' can only be default-imported using the 'esModuleInterop' flag
...
```

How ever, I think that this library can't work with `esModuleInterop: false`, so the tests are supposed to fail?

I also find out that my PR is just a reversion of #1352 

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 // Helpers
 type MergeBy<T, K> = Omit<T, keyof K> & K;
-export type StringMap = { [key: string]: any };
+type StringMap = { [key: string]: any };
 type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (k: infer I) => void
   ? I
   : never;
@@ -11,7 +11,7 @@ type LastOf<T> = UnionToIntersection<T extends any ? () => T : never> extends ()
 /**
  * This interface can be augmented by users to add types to `i18next` default TypeOptions.
  */
-export interface CustomTypeOptions {}
+interface CustomTypeOptions {}
 
 /**
  * This interface can be augmented by users to add types to `i18next` default PluginOptions.
@@ -38,9 +38,9 @@ export interface CustomTypeOptions {}
  * }
  * ```
  */
-export interface CustomPluginOptions {}
+interface CustomPluginOptions {}
 
-export type TypeOptions = MergeBy<
+type TypeOptions = MergeBy<
   {
     /**
      * Allows null values as valid translation
@@ -87,7 +87,7 @@ export type TypeOptions = MergeBy<
   CustomTypeOptions
 >;
 
-export type PluginOptions<T> = MergeBy<
+type PluginOptions<T> = MergeBy<
   {
     /**
      * Options for language detection - check documentation of plugin
@@ -116,14 +116,14 @@ export type PluginOptions<T> = MergeBy<
   CustomPluginOptions
 >;
 
-export type FormatFunction = (
+type FormatFunction = (
   value: any,
   format?: string,
   lng?: string,
   options?: InterpolationOptions & StringMap,
 ) => string;
 
-export interface InterpolationOptions {
+interface InterpolationOptions {
   /**
    * Format function see formatting for details
    * @default noop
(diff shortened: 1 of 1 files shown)


adrai:
@pedrodurek 🤷‍♂️

adrai:
@rosskevin you originally did #1352 right?

pedrodurek:
This is a problem with `"moduleResolution": "node16"`, I believe typescript 5 is tackling this problem https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/?ck_subscriber_id=1968760352#moduleresolution-bundler. 

@weareoutman could you bump your TS version to 5.0.0-beta to check it?

weareoutman:
@pedrodurek Actually I have tried TS 5.0.0-beta with `moduleResolution: "bundler"`, the type works, no warnings, and package exports work too.

However, here `bundler` is not I really want. I want to restrict code style just like `node16` (disabling index-file resolution and extensionless lookups), the biggest reason maybe because I have some code scanning task in my build process, I want the scanning be simple.

rosskevin:
@adrai yes I did #1352, with research documented in #1351 (linked from #1352).

Namespaces were deprecated by typescript and are legacy (as far as I can remember).  I believe the reason is they have no meaning in ECMA script or the future direction, so I see no reason why we would re-introduce them.

Reasons for use of `esModuleInterop` are documented in #1351 and it is fully recommended by the Typescript team to be in use (as referenced in #1352 and #1351)  

Perhaps this issue is related to the movement forward to ESM and the lack of `"type": "module",` in this project's `package.json`?    

As a note, this is the combination of project attributes and research I have found to work well as demonstrated in the linked sample project.  https://github.com/rosskevin/ts-esm-workspaces/tree/bug-main-required-to-build#workaround

`moduleResolution` in typescript can get finicky, if you try to use `nodenext` which I believe is same as `node16`, all your dependencies need to be esm.  Since that is a non-starter for most at this point in time, we need to use the above workaround.  Here is more infor about `nodenext` resolution and it's challenges https://github.com/rosskevin/ts-esm-workspaces 

I think this issue is about resolution, not about `namespaces`.  My guess is that adding `"type": "module"` will move this forward, but as you can see from my links, it is still finicky and all the stars have to align in the toolchains.

rosskevin:
@pedrodurek has it right, his link is right on with my research.  I still do recommend considering adding `"type": "module"` as it is the one key signal (I have found) to indicate you are shipping ESM and for the resolution to look at the new `exports` specification.  As an fyi, you have to leave the `main` even if you are ESM only, as noted in my research here https://github.com/rosskevin/ts-esm-workspaces/tree/bug-main-required-to-build#workaround

Perhaps typescript 5.x (or vscode) will address this and the workaround will be no longer necessary.

pedrodurek:
Since we agreed to no move forward with the approach used here, I'll be closing this PR. Thank you @weareoutman for your contribution!

Issue #1894, opened 2023-01-01
22.4.5 broke the types


falsandtru:
## 🐛 Bug Report

Updating from 22.4.4 to 22.4.5 broke the following code.

## To Reproduce

```ts
      interface TransDataMap {
        'Greeting': { name: string; };
      }
      function trans<K extends keyof TransDataMap>(el: HTMLElement, key: K, data: TransDataMap[K]) {
        i18next.init((err, t) =>
          el.textContent = err
            ? ''
            : t(key, data) ?? ``);
      }
```

## Expected behavior

pass

## Your Environment

- *runtime version*: v19
- *i18next version*: 22.4.5
- *os*: Windows

falsandtru:
@adrai @pedrodurek

falsandtru:
@adrai @pedrodurek STOP THE BOT!!

falsandtru:
FUCK YOU

adrai:
Please try with i18next v23.0.1 and react-i18next v13.0.0

falsandtru:
Got errors.

```
Argument of type '[K, TransDataMap[K]]' is not assignable to parameter of type '[key: K | K[], options?: (TransDataMap[K] & InterpolationMap<DefaultTReturn<TransDataMap[K]>>) | undefined] | [key: K | K[], defaultValue: string, options?: (TransDataMap[K] & InterpolationMap<DefaultTReturn<TransDataMap[K]>>) | undefined]'.
  Type '[K, TransDataMap[K]]' is not assignable to type '[key: K | K[], defaultValue: string, options?: (TransDataMap[K] & InterpolationMap<DefaultTReturn<TransDataMap[K]>>) | undefined]'.
    Type at position 1 in source is not compatible with type at position 1 in target.
      Type 'TransDataMap[K]' is not assignable to type 'string'.
        Type '{ name: string; }' is not assignable to type 'string'.ts(2345)
(parameter) key: K extends "Greeting"
```

adrai:
@falsandtru please provide a reproducible example

falsandtru:
@adrai Same as the repro code at first in this issue.

adrai:
That's just a code snippet, and nod a complete but minimal reproducible example repository or similar...
I can't see the the reason why you are wrapping the init function like that... What is your use case?
Create a new github repository showing what you want to achieve...

falsandtru:
Then the next is how to repro:

1. Clone https://github.com/falsandtru/typed-dom
[...]

Issue #1889, opened 2022-12-18
getFixedT() types and namespace prefixes


kkuegler:
## 🐛 Bug Report

Using the `'otherNamespace:my.key'` syntax with a `getFixedT()` function bound to a namespace works at runtime, but leads to TS type errors in the editor.

The problem is related to #1887, but instead of less prefixed keys, I want more prefixed ones. 😄 

## To Reproduce

I created a [minimal reproducible example commit](https://github.com/kkuegler/i18next/commit/b6bbfece06b054772e697d344b57a1c36572eb77) based on the existing i18next "customtypes" tests.

The [repository](https://github.com/kkuegler/i18next/tree/master) itself is not "minimal", because it is i18next master plus the commit linked above.

Running
```sh
npm run test:typescript:customtypes
```
will fail to typecheck for the lines with a `// TS error, but works ok` comment. I also included these below:

```js
const t8 = i18next.getFixedT('en', 'alternate');
// this is not the most useful call, but maybe there are use-cases for prefixing everything:
t8('alternate:foobar.barfoo'); // TS error, but works ok
// other namespaces should be available via a key prefix:
t8('custom:foo'); // TS error, but works ok
```
In my mind, getFixedT() should allow the keys for the "fixed" namespace(s) without a prefix (as we currently do), but also allow the keys of *all* other namespaces available in i18next with the respective namespace prefix. This would match the current runtime behavior of `getFixedT()`. I understand, that in the IDE this would suggest a lot of keys that are not helpful in most cases, because one usually would only want to work with the "fixed" namespace's keys.

As an alternative, we might also keep basically the current behavior, but add another generic fallback overload to the `TFunction` definition, that just allows keys matching the type ``` `existing-ns-prefixes:${string}` ```, i.e. not generating every possible prefixed key, but just allowing any key with the correct prefix for any of the namespaces known to i18next. This cannot suggest existing prefixed keys or detect misspelled namespace-prefixed keys, but would not produce a giant list of possible keys.

I'm not a typescript expert, so I'll leave the judging to you. :)

## Expected behavior
The code above should typecheck just fine.

## Your Environment

- *runtime version*: Node.js v14.21.1, npm 9.2.0
- *i18next version*: 22.4.5 / master
- *os*: Windows

adrai:
Would you like to provide a PR?

kkuegler:
> Would you like to provide a PR?

I can try and come up with one for the second alternative.

kkuegler:
[...]

Issue #1867, opened 2022-11-16
Allow unknown keys, when a `defaultValue` is provided


janpaepke:
## 💥 Regression Report

This isn't really a functional regression, more like a convenience regression, it only concerns the new [Typescript enhancements](https://www.i18next.com/overview/typescript).
When using the t function it is now impossible to use with keys that don't exist in the default namespace, even when a default value is provided, which it could fall back to.

This forces me to always add every key to my default namespace's resource file.
Usually in our workflow we only add translations inline using the defaultValue, when developing a new feature.
This is because things often change.
Only once something is production ready, we run i18next-parser to extract all keys and add translations.

## Proposal

I propose to add additional overloads for when a default value is provided (one for `t('key', 'defaultValue')`, one for `t('key', { defaultValue: 'value' })`, where the key can be any string.

## Last working version

Worked up to version: 21

Stopped working in version: 22

## To Reproduce

```ts
t('unkown.key', 'default translation')
// ⬆ No overload matches this call.
//  The last overload gave the following error.
//    Argument of type '"unkown.key"' is not assignable to parameter of type 'Normalize<{ ... }>  | TemplateStringsArray ...`
```

## Your Environment

- *runtime version*: i.e. node v16
- *i18next version*: i.e. 22.0.5
- *os*: Mac

janpaepke:
I'll put it on my todo list. It's long. ;)

janpaepke:
So I had some time and adding the overloads was actually pretty straight forward.

here they are:
```ts
  <
    TDefaultResult extends DefaultTFuncReturn = string,
    TInterpolationMap extends object = StringMap,
  >(
    key: string,
    options?: TOptions<TInterpolationMap> & { defaultValue: string },
  ): TFuncReturn<N, string, TDefaultResult, TKPrefix>;
  <
    TDefaultResult extends DefaultTFuncReturn = string,
    TInterpolationMap extends object = StringMap,
  >(
    key: string,
    defaultValue: string,
    options?: TOptions<TInterpolationMap> | string,
  ): TFuncReturn<N, string, TDefaultResult, TKPrefix>;
```
I tested them and they work as expected.  
They even still prefer the known keys so key autocomplete still works, when a default value is provided.

I could create a pr, but since this is a types-only changes, I failed to find the appropriate testing suite (was looking for [tsd](https://www.npmjs.com/package/tsd)).

Can you point me in the right direction?

adrai:
[...]

PR #1911, opened 2023-02-11
Redesign `t` function types

pedrodurek:
### Description
The existing implementation of the type for the `t` function is characterized by being very complex, slow, hard to maintain, and not scalable, which are also hampered by various limitations. 

Our redesign endeavours to enhance the approach to parsing and inferring keys for the `t` function. Instead of performing a recursive examination of each key-value pair in resources associated with a specific namespace each time the `t` function is invoked, we generate a comprehensive set of keys from all namespaces just once. Then, we parse and filter the keys based on the designated namespace and key prefix. The outcome of this redesign is that the types have been greatly simplified, and are faster, smarter, more scalable, and are easier to maintain. This upgrade will allow us to mitigate numerous issues raised by devs.

### Features
...
### Fixes
...
### Breaking changes
...
### Benchmark results
...
### Limitations
...

#### Checklist

- [ ] only relevant code is changed (make a diff before you submit the PR)
- [ ] run tests `npm run test`
- [ ] tests are included


pedrodurek:
There are a few remaining tasks that I'll be working on in the next few days before merging this PR, that includes:

- [ ] Open a PR for `react-i18next` repo supporting the new types.
- [ ] Open a PR for `next-i18next` repo supporting the new types.
- [ ] Create new tests and update current ones.
- [ ] Update `i18next` documentation.
- [ ] Include benchmark results here.

kkuegler:
(this comment is somewhat tangential to this PR, but I think being aware of this can help shape this new typing approach)

Hi @pedrodurek,

earlier this year, I tried to come up with some type fixes for `getFixedT()` in https://github.com/i18next/i18next/pull/1898.

Maybe we can incorporate some of those ideas into your new typing approach. Also, that PR had some new test cases that might make sense to incorporate here/add later.

## getFixedT()
To me, it seems the types for `getFixedT()` should work somewhat differently to the `useTranslation()` ones:

Adapting your first example from the PR description, `getFixedT(null, ['ns1', 'ns2'])` should allow keys from both namespaces both without a ns prefix and with a prefix, i.e. `'key1' | 'key2' | 'key3' | 'key4' | 'ns1:key1' | 'ns1:key2' | 'ns2:key3' | 'ns2:key4'`.

`getFixedT(null, ['ns1'])` should allow `'key1' | 'key2' | 'ns1:key1' | 'ns1:key2'`. This seems to work fine with this PR.

`getFixedT(null, 'ns1')` (with a string instead of an array) should allow the same keys, but with this PR only allows `'key1' | 'key2'`.
To change this, one could change the `ParseKeysByNamespaces` type to this:
```
type $Flatten<A> = A extends ReadonlyArray<infer E> ? E : A;
type ParseKeysByNamespaces<
  Ns extends Namespace,
  Keys,
  UnionNsps = $Flatten<Ns>,
> = UnionNsps extends keyof Keys ? AppendNamespace<UnionNsps, Keys[UnionNsps]> : never;
```

As outlined in https://github.com/i18next/i18next/issues/1889 (the bug corresponding to the PR mentioned above), `getFixedT()` - like `i18next.t()` - should allow the keys of all other namespaces available in i18next with the respective namespace prefix - even if they are not mentioned via the namespace parameter to `getFixedT()`.

So in fact `getFixedT(null, ['ns1'])` and `getFixedT(null, 'ns1')` should allow `'key1' | 'key2' | 'ns1:key1' | 'ns1:key2' | 'ns2:key3' | 'ns2:key4'`. This is quite different to how `useTranslation()` seems to work.

## Lots of TS key suggestions
These are the same keys that are valid for your `i18next.t()` example. In the real world, there are much more translation keys. Also, the non-namespaced keys don't usually happen to be alphabetically before the namespaced keys like in the example. Due to this, TS will suggest a lot of keys and will happily mix ns-prefixed and unprefixed keys. I think this can be quite a mess to look at, making the key suggestions less helpful.

To cope with that I came up with the (hack-ish) type `DisableSuggestion` (see https://github.com/kkuegler/i18next/blob/18c6274451d133f8df98c91b53b8b98ba4db15d5/index.d.ts#L10). This can allow some keys, but keep them out of the TS suggestion list. I hoped to find more "official" ways to do this, but couldn't find any.

## The end :)

I'm not sure if you can incorporate some of these aspects into this PR or not, but in the long run I think we should address these, so the runtime behavior and the types for `getFixedT()` match.

jamuhl:
Perhaps we should make clear what `useTranslation` is...it's only reason to exist is to assert the namespaces provided are loaded (nothing more) before rendering (and is only needed in combination with using backends).

the `getFixedT` is just a helper to get the `t` function, but not bound to the default namespace and bound to namespace(s) passed in (bound means just not needing the namespace prefix `ns1:key`).

`t` function itself does not care about translations being loaded or not. It just looks up translations based on given namespace and key (and does this by using default namespace or given namespace as prefix to the key)


Under the hood `useTranslation` calls the `getFixedT` function and returns the provided bound `t` function (beside other stuff)

VictorGlindasPaf:
This looks like a fantastic improvement.
We're using  these updated types at work, and have some feedback. I know it's a W.I.P, but could be useful nonetheless.

- Autocomplete of the interpolated values seems to break when using multiple namespaces and accessing the key without the namespace prefix.

- The types are faster, but still very slow when we have many keys. We have about 1500 keys in our project, and strictly typing them made the `tsc` compilation time goes from ~120s to ~250s.

diegodorado:
Thanks for the thourough description of what the PR does @pedrodurek ! I am eager to see this merged, and I am specially interested in typed interpolation params.

Until this is merged, I have a quick question on how `useTranslation` would be better used:

With `i18next@22.4.11` and `react-i18next@12.2.0`, and having resources defined as
```ts
const ns1 = {
  key1: '',
}

const ns2 = {
  key2: '',
}

const resources = {
  en: {
    ns1,
    ns2,
  },
}
```

Depending if I call `useTranslation(nsArray)` or `useTranslation<NSType>()` I get different allowed keys for `t`.

With the first one:
```ts
  const { t } = useTranslation(['ns1', 'ns2'])
  t('key1') // no error
  t('ns1:key1') // no error
  t('key2') // TSC Error: No overload matches
  t('ns2:key2') // no error
```

While the later:
```ts
  const { t } = useTranslation<'ns1'|'ns2'>()
  t('key1') // no error
  t('ns1:key1') // TSC Error: No overload matches
  t('key2') // no error
  t('ns2:key2') // TSC Error: No overload matches

```

As I understand, using one or the other might be subject to preference or use cases, but curious if there is any reason to use one or the other or if the PR proposal is to deprecated one of these.

Thanks!

falsandtru:
Can this resolve #1894?

pedrodurek:
### Benchmark results in Mac M1:

**Num keys:** 11,429
**Num namespaces:** 55
**Compilation time:** ~6.27s
**VSCode autocomplete keys:** ~2.477s

I won’t share the results of the current types because it doesn’t handle these many keys/namespaces. 😅

I’ll share the project that I used for the benchmark once we merge this PR.

pedrodurek:
Hey @falsandtru, I'm not sure, but we can work something out once it's merged

falsandtru:
I see

VictorGlindasPaf:
We found another issue. In the new types.
When the namespace can be one of several ones, it doesn't complain if we access a key that only exists on one of them, but not the other.

Screenshot 2023-04-18 at 10 31 58

In the above example, `choose-amount.heading` is available only on `deposit`, but we get no error.

adrai:
fyi: ts tests are failing: https://app.circleci.com/pipelines/github/i18next/i18next/1104/workflows/55564641-6a5f-462f-af39-139d8e209ba8/jobs/908/parallel-runs/0/steps/0-105

danr-za:
Are there any plans on merging this in?

adrai:
> Are there any plans on merging this in?

@danr-za as soon as ready

wolthers:
Thanks for putting in this work @pedrodurek. Don't mean to be unappreciative, but could you give an update? [641](https://github.com/i18next/react-i18next/issues/1641) is a minimal repro that shows that any app that wants to bump typescript to v5 but also includes react-18next will fail typecheck.

adrai (approved):
lgtm... do you want me to merge it now? also let me know if I can/should start with the other changes relevant for the next major version.

Can you also adapt the docs afterwards, if necessary?
btw: https://github.com/i18next/react-i18next/pull/1615

adrai:
@pedrodurek does this example also need to be updated? https://github.com/i18next/i18next/tree/master/examples/typescript it seems to not work with the new index.d.ts
image

adrai (changes requested):
@pedrodurek does this example also need to be updated? https://github.com/i18next/i18next/tree/master/examples/typescript it seems to not work with the new index.d.ts
image

adrai:
It only works, if the index.d.ts file is included in the tsconfig...
image

is this intended to work like that @pedrodurek ?

pedrodurek:
Hey @adrai, it's safe to merge it, I'll update the other PRs once this one is merged.

> Do you also want to update the typescript dependency?

Done!

> Can you also adapt the docs afterwards, if necessary?

Do you mean [this](https://github.com/i18next/i18next-gitbook/pull/206)?

> @pedrodurek does this example also need to be updated?

I'll update the examples with the correct i18n version once the PR is merged.

adrai:
fyi: goal is to release a new major version with https://github.com/i18next/i18next/pull/1885 and https://github.com/i18next/i18next/pull/1945 and also addressing some other stuff: https://github.com/i18next/i18next/issues/1884
but first we need to address:
https://github.com/i18next/i18next/pull/1885#issuecomment-1582181256
https://github.com/i18next/i18next/pull/1911#issuecomment-1576121449

adrai:
> Do you mean [this](https://github.com/i18next/i18next-gitbook/pull/206)?

@pedrodurek I mean because of [this](https://github.com/i18next/i18next/pull/1911#issuecomment-1576121449)

adrai:
@pedrodurek now all but [this](https://github.com/i18next/i18next/pull/1911#issuecomment-1576121449) is ready to release the major version...
Can you check the example and describe how the new typesafe translation approach should be set up in a project?

adrai:
seems the compilerOptions needs to be strict: https://github.com/i18next/i18next/commit/b7075f2ede932ac9f08dee4b5d99f6a22d848735#diff-603812bc45f8ec4c682a4be24c9423d325e92c7ea7f10afe1bd84b6c46077164R10

adrai:
stuff to check before releasing this:
- why is [this interpolation](https://github.com/i18next/i18next/blob/e5071dd61176375dcf4bc659416ec692556424a2/test/typescript/custom-types/t.test.ts#L9) check working in [tests](https://github.com/i18next/i18next/blob/e5071dd61176375dcf4bc659416ec692556424a2/test/typescript/custom-types/t.test.ts#L9), but not in [example](https://github.com/i18next/i18next/tree/master/examples/typescript)
- why is [strict mode necessary for this to work](https://github.com/i18next/i18next/blob/e5071dd61176375dcf4bc659416ec692556424a2/examples/typescript/tsconfig.json#L13)? is there a workaround?

adrai:
> * why is [this interpolation](https://github.com/i18next/i18next/blob/e5071dd61176375dcf4bc659416ec692556424a2/test/typescript/custom-types/t.test.ts#L9) check working in [tests](https://github.com/i18next/i18next/blob/e5071dd61176375dcf4bc659416ec692556424a2/test/typescript/custom-types/t.test.ts#L9), but not in [example](https://github.com/i18next/i18next/tree/master/examples/typescript)

seems because typeof [here](https://github.com/i18next/i18next/blob/52e5712f1cddd6da08338fa2a29bca5c169b2725/examples/typescript/%40types/i18next.d.ts#L6) is just using string type for the translation values, so there is no info of the interpolation anymore...

adrai:
> * why is [strict mode necessary for this to work](https://github.com/i18next/i18next/blob/e5071dd61176375dcf4bc659416ec692556424a2/examples/typescript/tsconfig.json#L13)? is there a workaround?

`"strictNullChecks": true` is sufficient

janpaepke:
Hi guys. This update reintroduced an Issue I had fixed in the previous version:
For details see here: https://github.com/i18next/i18next/issues/1867

Essentially, when the t function receives a default value (either as a string, or as part of the `defaultValue` prop in the options), typescript should not enforce "known keys only".

That's the whole point of the default value option. If we're forcing people to always have in their main namespace, then this entire option is completely useless, as it will always be overwritten anyway.

The same is true by the way for the `Trans` component of `react-i18next`, but this is probably less of your concern.

adrai:
> Hi guys. This update reintroduced an Issue I had fixed in the previous version: For details see here: #1867
> 
> Essentially, when the t function receives a default value (either as a string, or as part of the `defaultValue` prop in the options), typescript should not enforce "known keys only".
> 
> That's the whole point of the default value option. If we're forcing people to always have in their main namespace, then this entire option is completely useless, as it will always be overwritten anyway.
> 
> The same is true by the way for the `Trans` component of `react-i18next`, but this is probably less of your concern.

@pedrodurek what do you think?

janpaepke:
#### addendum
My gut instinct would be to solve this via overloads:
```typescript
/**************************
 * T function declaration *
 **************************/
  <
    TOpt extends TOptions,
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    ...args:
      | [key: Key | Key[], options?: TOpt & InterpolationMap<Ret>]
      | [key: Key | Key[], defaultValue: string, options?: TOpt & InterpolationMap<Ret>]
  ): TFunctionReturnOptionalDetails<Ret, TOpt>;
  <
    TOpt extends TOptions,
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    ...args:
      | [key: string | string[], options?: TOpt & InterpolationMap<Ret> & { defaultValue: string }]
      | [key: string | string[], defaultValue: string, options?: TOpt & InterpolationMap<Ret>]
  ): TFunctionReturnOptionalDetails<Ret, TOpt>;
}
```
But you likely had a reason for eliminating them (probably speed?).
I haven't tested or benchmarked the above, but it might help as a starting point.

adrai:
@janpaepke can you try with v23.2.6 ?

janpaepke:
I did and it works. 👍
Thanks for your quick reaction!

It let's me set any string as the key, when a default value is provided and complains if there's not.
Is it possible to implement a similar solution for the Trans component? (default value is `children` in that case).
Or should I post this in a different repo?


There's another thing I noticed, but I am not entirely sure if it's related to my environment (IDE/ts version).

Key-Autocompletion for the t function doesn't seem to work for me any more.
Interestingly it does still work for the Trans Component.
And yes, this was also the case without this latest change.

Should I create a separate issue for that?

adrai:
@janpaepke intellisense is tracked here: https://github.com/microsoft/TypeScript/issues/54708 (probably fixed in next ts version and/or ide version)
Regarding Trans component, yes open a new issue in the react-i18next repo.

janpaepke:
@adrai thanks a lot!
Trans Component -> will add an issue.
re your workaround: How would leaving out the key work? It couldn't relate the respective translation vars, if available, right?
Plus i18next-parser can't extract it anymore... Sorry, if off topic.

adrai:
> In a React Component, when I'm loading a namespace `foo` (not configured to be the default ns !), I got an TS error by not specifying the namespace in the `t` function :
> 
> ```
> const { t } = useTranslation('foo');
> ...
> <div>
>   // TS error
>   {t('key')}
>   // No TS error
>   {t('foo:key')}
> </div>
> ```
> 
> The TS error says `Argument of type '["key"]' is not assignable to parameter of type '[key: TemplateStringsArray |.....`. I thought it was possible, not to specify the namespace but seems, it's requiered with v23.
> 
> Should you update this [documentation of loading namespace in react-i18next](https://react.i18next.com/latest/usetranslation-hook#loading-namespaces) ?
> 
> Thanks in advance for your answer

@MiryksV please provide a complete but minimal reproducible example repository or similar...

MiryksV:
I couldn't reproduce on my [sandbox](https://codesandbox.io/s/quirky-ritchie-764q9n?file=/src/Component.tsx), I'll take a deeper look in my project. Sry to bothering you for nothing. \
Thanks for your answer.

reckter:
I just adopted this, and the typescript compiler needs about 8.5GB ram now in order to type check our code. 
It takes about 1.7min (with a few errors left to fix).

We  use the interface generation, because we have .json files translations.
The interface file is 8888 lines long (I swear, I didn't make this up xD), with 26 namespaces.

Without the generated interfaces included, typescript needs about ~3s to run.

If you have any ideas on why it might take so long, I would really appreciate it!

reckter:
Seems like, once I had a no-error run the tsc cache really speeds up the compiler.
Every time I bust the cache (manual deletion, or different branches) it will take up to ~2.5 minutes and over 8GB of ram.
After a successful run it will be in the low seconds again.

semenov-ol:
Hey, before i can use generic type to specify which type would return
`i18n.t<string, string[]>('i18nkey', { returnObjects: true })`

How can i do the same with new version, when passing third param ts complain that 'Type  string[]  does not satisfy the constraint  string'?
`i18n.t<string, any, string[]>('i18nkey', { returnObjects: true })`

felixmeziere:
I have the same issue as @reckter with very long tsc run times because of this, any way to solve the issue ?

PR #1932, opened 2023-03-27
fix: avoid format function error without format parameter and refine related type

marcalexiei:
#### Summary

Currently calling `i18n.format()` without the second parameter throws a TypeError:
`undefined is not an object (evaluating '_format.split')`

Someone could accidentally forget to specify it and encounter the above error runtime (that would be me 😅).

Typescript isn't throwing any error so, since calling format function without the second parameter doesn't have much sense,  I would like to propose to make the `FormatFunction` `format` parameter mandatory in the definitions.
Also I changed the format function to issue a warning and return the value if `format` parameter isn't provided.
Let me know if you prefer that the function throws an error with the same message.

#### Example

```ts
import { default as i18n } from "i18next";

i18n.init({
  lng: "en",
  resources: {
    en: {
      translation: { text: "my {{value, uppercase}}" }
    }
  },
  debug: true
});

i18n.services.formatter.add("uppercase", (value: string) => value.toUpperCase(););

console.info(i18n.format("text"));
```

[codesandbox.io](https://codesandbox.io/s/i18next-formatter-2nd-param-required-kn35g5?file=/src/index.ts:458-489)

#### Additional changes

- I refined eslintrc `globals` using `env` to be able to use `beforeEach` and `afterEach` in tests
- I fixed some ESlint errors in `i18next.translation.formatting.spec.js`

#### Checklist

- [x] tests are included


adrai:
The format function is not a public interface....
You should not call i18next.format in userland.

adrai:
Please explain why you need to call the format function from userland or close this PR.

marcalexiei:
I didn't know that the function was for internal use.
I have some custom formatter and I need to process some "raw" data like string and number provided from API with them.

Using format function directly avoids to create a translation key like: `"formatWithFormatter": "{{value, formatter}}"`.

adrai:
I suggest you extract your custom format function and use that one instead of using i18next.format()
The format function could change in future... since internal.

marcalexiei:
Gotcha.
Maybe I can change the PR to add a comment about this on the `i18n.format` function declaration (maybe it can be marked with `@deprecated` to add a visible highlight to avoid its usage on userland).
If not you can close the PR.

PR #1934, opened 2023-03-27
Align `defaultTransParent` type with Trans `parent`

### LINKED ISSUE #1933: `defaultTransParent` type incompatible with Trans `parent`
## 🐛 Bug Report

https://github.com/i18next/i18next/blob/dd133af54f071deadbd8fd9dd95e0153b4e6deea/index.d.ts#L252

`defaultTransParent` type is missing `React.ComponentType<any>`. See `Trans` [types](https://github.com/i18next/react-i18next/blob/7026d5461409529145d99525cd2463ecd3111b0b/TransWithoutContext.d.ts#LL21C21-L21C45) definition.

## Expected behavior

it should be possible to set react component as `defaultTransParent`.

```ts
import { Text } from 'my-ui-lib';

const config: InitOptions = {
  lng: 'en',
  react: {
    defaultTransParent: Text,
  },
};
```

## Your Environment

- *i18next version*: latest master

fantua:
Fixes #1933 

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -249,7 +249,7 @@ export interface ReactOptions {
   nsMode?: 'default' | 'fallback';
   /**
    * Set it to the default parent element created by the Trans component.
-   * @default 'div'
+   * @default undefined
    */
   defaultTransParent?: string | React.ComponentType<any> | null;
   /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+import * as React from 'react';
+
 // Helpers
 type MergeBy<T, K> = Omit<T, keyof K> & K;
 export type StringMap = { [key: string]: any };
@@ -249,7 +251,7 @@ export interface ReactOptions {
    * Set it to the default parent element created by the Trans component.
    * @default 'div'
    */
-  defaultTransParent?: string;
+  defaultTransParent?: string | React.ComponentType<any> | null;
   /**
    * Set which events trigger a re-render, can be set to false or string of events
    * @default 'languageChanged'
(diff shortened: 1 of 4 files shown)


adrai:
@pedrodurek is this the only way to solve this?
Normally i18next has no react dependency.

fantua:
I need only for test:
https://github.com/i18next/i18next/blob/2fd96613e90b513bb4532dd31c659462b588f6c5/test/typescript/init.test.ts#L478-L483

adrai:
It's also in the index.d.ts file that's why I'm asking 🤷‍♂️

fantua:
True, looks like with change I also need to add react to peerDeps 🤔 Will create new ts project locally with i18n and play a bit

fantua:
If I don't install `@types/react` package TS will resolve `defaultTransParent` as `any` without errors. So maybe it makes sense to add `react` as optional peerDep?

adrai:
> If I don't install `@types/react` package TS will resolve `defaultTransParent` as `any` without errors. So maybe it makes sense to add `react` as optional peerDep?

idk, what do you think @pedrodurek ?

jamuhl:
I vote against adding anything react specific into this repo

pedrodurek:
Sorry for the delay, I haven't had much free time lately. I've been working overtime for the last weeks/months.
I also don't agree, react components are functions, so they can be represented like `(...arg: any) => any`, or we can copy the `ComponentType` type.

fantua:
> I vote against adding anything react specific into this repo

this repo already contains react specific configuration (`ReactOptions`) 🤷‍♂️ might be worth considering moving these config to `react-i18next` directly 

> I also don't agree, react components are functions, so they can be represented like (...arg: any) => any

then we can just change `defaultTransParent` type to `any`

> or we can copy the ComponentType type to here.

[type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b4fa1b97bcb81d1c5358c37c0086df279bba9824/types/react/index.d.ts#L76) is very complex, I'm not sure it's a good idea to copy and maintain after

Issue #1956, opened 2023-05-05
Typescript slow to show translation keys in intellisense


Bramgus12:
### Discussed in https://github.com/i18next/i18next/discussions/1955

<div type='discussions-op-text'>

<sup>Originally posted by **Bramgus12** May  5, 2023</sup>
Hi, I have some trouble with using typescript with i18n and react.

When I try to use the `t` function in my code and specifically as a child in tsx, typescript is slow with loading the types that I can use in that function. Can sometimes take up to 5 mins to load. While it is loading typescript itself is so busy getting the types that it won't do anything else. This is really annoying if you have to keep restarting your ts server to make everything work again. My colleagues also tried to do the same thing as I did and had almost no issues (max 3 seconds to load the types). Only difference is that I have a MacBook pro m1-pro while my colleagues are using windows/linux. The translation files that I have a really small (max 30 translations per file).
Anyone know what I am doing wrong?

Here is my setup: 
![image](https://user-images.githubusercontent.com/42220620/236499712-b8748195-eeac-4cc3-9fbe-cd9f12f4efab.png)

i18n.ts
```ts
import i18n from "i18next";
import { initReactI18next } from "react-i18next";

import Backend from "i18next-http-backend";
import LanguageDetector from "i18next-browser-languagedetector";
import componentsENGB from "./languages/en_GB/components.json";
import generalENGB from "./languages/en_GB/general.json";
import componentsNLNL from "./languages/nl_NL/components.json";
import generalNLNL from "./languages/nl_NL/general.json";
import layoutsNLNL from "./languages/nl_NL/layouts.json";
import layoutsENGB from "./languages/en_GB/layouts.json";
import decisionToolNLNL from "./languages/nl_NL/decisionTool.json";
import decisionToolENGB from "./languages/en_GB/decisionTool.json";

export const defaultNS = "general";
export const resources = {
    en_GB: {
        components: componentsENGB,
        general: generalENGB,
        layouts: layoutsENGB,
        decisionTool: decisionToolENGB,
    },
    nl_NL: {
        components: componentsNLNL,
        decisionTool: decisionToolNLNL,
        general: generalNLNL,
        layouts: layoutsNLNL,
    },
} as const;

export const i18nInit = i18n.use(Backend)
    .use(LanguageDetector)
    .use(initReactI18next)
    .init({
        fallbackLng: "en_GB",
        // TODO: debug should only be true on development
        debug: true,
        interpolation: {
            escapeValue: false,
        },
[...]

Issue #1921, opened 2023-03-07
Type instantiation is excessively deep and possibly infinite


bildungsroman:
## 🐛 Bug Report

I'm getting the error `Type instantiation is excessively deep and possibly infinite` when defining `CustomTypeOptions`. 

My type definition:

```js
// i18next.d.ts
import resources from "../i18n/locales/resources";

declare module "i18next" {
	interface CustomTypeOptions {
		returnNull: false;
		resources: typeof resources['en'];
	}
}
```

where `resources` are:
```js
// i18n/locales/resources
import definitions_de from './definitions/de.json';
import definitions_en from './definitions/en.json';
import errors_de from './errors/de.json';
import errors_en from './errors/en.json';

export const resources = {
	de: {
		definitions: definitions_de,
		errors: errors_de,
	},
	en: {
		definitions: definitions_en,
		errors: errors_en,
	},
<etc.>
}
```

In my code, I'm doing:

```js
// src/myCode
const t = i18nInstance.getFixedT(this.locale as Locale);

throw new Error(t('index.edgeError'));
```

## Expected behavior

I would expect the type definition to work. And the code to compile. However, while I don't get any tslint errors, my tests are failing with the error `Type instantiation is excessively deep and possibly infinite`, which also shows up (sometimes) in VS Code.

## Your Environment

- *runtime version*: node16
- *i18next version*: "^22.4.10", "typescript": "^4.9.5",
- *os*: Mac

sedge:
I'm seeing something very similar 👀

adrai:
https://github.com/i18next/react-i18next/issues/1601
https://github.com/i18next/react-i18next/issues/1600
https://github.com/i18next/i18next/pull/1911

adrai:
Please try with i18next v23.0.1 and react-i18next v13.0.0

tsuyoyo:
FYI: In our case, we needed `react-i18next v13.0.1` (not `v13.0.0`) to resolve `Type instantiation is excessively deep and possibly infinite` warning.

https://github.com/i18next/react-i18next/issues/1601#issuecomment-1602029316

yeliex:
the problem still exist with `i18next@23.2.5` and `react-i18next@13.0.1` when there are many namespaces(currently about 20 and would be 100+ in thre future)

adrai:
@yeliex an example repository may help @pedrodurek to investigate

esetnik:
[...]

Issue #1883, opened 2022-12-13
Run eslint -> JavaScript heap out of memory


ximex:
## 🐛 Bug Report

If i run eslint i got this error since i18next v2.4.4. It comes from the rules "@typescript-eslint/no-misused-promises". I don't know if it is a @typescript-eslint or i18next bug.
With v2.4.3 or before there are no problems.
```
> eslint ./


<--- Last few GCs --->

[12938:0x67d8210]    77956 ms: Mark-sweep (reduce) 1881.7 (1921.3) -> 1881.7 (1914.3) MB, 950.8 / 0.0 ms  (average mu = 0.564, current mu = 0.000) last resort; GC in old space requested
[12938:0x67d8210]    78893 ms: Mark-sweep (reduce) 1881.7 (1914.3) -> 1881.7 (1914.3) MB, 936.6 / 0.0 ms  (average mu = 0.379, current mu = 0.000) last resort; GC in old space requested


<--- JS stacktrace --->

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: 0xb6e500 node::Abort() [node]
 2: 0xa7e632  [node]
 3: 0xd47f20 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd482c7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xf137bf v8::internal::HeapAllocator::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 6: 0xef3b80 v8::internal::Factory::AllocateRaw(int, v8::internal::AllocationType, v8::internal::AllocationAlignment) [node]
 7: 0xeeb14c v8::internal::FactoryBase<v8::internal::Factory>::AllocateRawArray(int, v8::internal::AllocationType) [node]
 8: 0xeeb2c5 v8::internal::FactoryBase<v8::internal::Factory>::NewFixedArrayWithFiller(v8::internal::Handle<v8::internal::Map>, int, v8::internal::Handle<v8::internal::Oddball>, v8::internal::AllocationType) [node]
[...]

PR #1971, opened 2023-06-15
Add test

falsandtru:
Add a failed test requested in #1894.


test/typescript/t.test.ts
```diff
@@ -172,7 +172,7 @@ function datamap(t: TFunction) {
   interface TransDataMap {
     'Greeting': { name: string; };
   }
-  function trans<K extends keyof TransDataMap>(key: K, data: TransDataMap[K]): string {
+  function trans<K extends keyof TransDataMap>(key: K, data: TransDataMap[K]): string | TFunctionDetailedResult<string> {
```
falsandtru, line 175: This has a problem that it doesn't work with `el.textContent = trans(...)`. And mismatching of the return type has a function to prevent the data type form adding invalid properties. I'll improve the test.

adrai: What I don't understand is, why is `TOpt['returnDetails'] extends true` always true?

falsandtru: Because type `TOpt['returnDetails']` is not decided. So the result of the type function is both return types as all the possibilities.

adrai: quantum logic 😜

falsandtru: Fixed.

adrai:
If you know of how to best address this in the types, feel free to extend this PR...
I just know that @pedrodurek wanted to eliminate the overloads to improve the performance... 🤷‍♂️

adrai:
btw: fixing this will probably also help for: https://github.com/i18next/react-i18next/issues/1571

falsandtru:
Unfortunately I don't know of any solution to this other than overloads. And I don't know react. BTW, performance problems seem to be increased after eliminating the overloads: #1972, #1956, #1921, #1883. I don't know but do overloads generally worsen performance?

adrai:
Ok, than try with the overloads... at least we have a possible solution...

falsandtru:
Sounds good. I want to use the latest version.

adrai:
Probably something like this?

```ts
  <
    TOpt extends TOptions & { returnObjects: true },
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    key: Key | Key[],
    options: TOpt & InterpolationMap<Ret> & { returnObjects: true },
  ): TFunctionReturnOptionalDetails<Ret, TOpt>;

  <
    TOpt extends TOptions & { returnDetails: true},
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    key: Key | Key[],
    options: TOpt & InterpolationMap<Ret> & { returnDetails: true},
  ): TFunctionReturnOptionalDetails<Ret, TOpt>;

  <
    TOpt extends TOptions,
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    key: Key | Key[],
    options?: TOpt & InterpolationMap<Ret>,
  ): TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>;

  <
    TOpt extends TOptions & { returnObjects: true },
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    key: Key | Key[],
    defaultValue: string,
    options: TOpt & InterpolationMap<Ret> & { returnObjects: true },
  ): TFunctionReturnOptionalDetails<Ret, TOpt>;

  <
    TOpt extends TOptions & { returnDetails: true },
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    key: Key | Key[],
    defaultValue: string,
    options: TOpt & InterpolationMap<Ret> & { returnDetails: true },
  ): TFunctionReturnOptionalDetails<Ret, TOpt>;

  <
    TOpt extends TOptions,
    Ret extends TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>,
  >(
    key: Key | Key[],
    defaultValue: string,
    options?: TOpt & InterpolationMap<Ret>,
  ): TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>;
}
```

falsandtru:
Probably Ret should be removed.

```ts
  <
    TOpt extends TOptions & { returnObjects: true },
  >(
    key: Key | Key[],
    options: TOpt & { returnObjects: true },
  ): TFunctionReturnOptionalDetails<TFunctionReturn<Ns, AppendKeyPrefix<Key, KPrefix>, TOpt>, TOpt>;
```

This breaks some tests but the current definition is originally wrong. The next test is valid.

```ts
  const o: object = t('friend', { returnObjects: true });
```

But the actual type without type annotations is DefaultTReturn<{ returnObjects: true; }> not assignable to object.

```ts
  const o = t('friend', { returnObjects: true });
```

Current definitions essentially just fit the type to annotations as the next.

```ts
type $SpecialObject = object | Array<string | object>;
...
    Ret extends $SpecialObject,
```

adrai:
Need the advice of @pedrodurek here....

falsandtru:
Added the tests. Tests are confusing type casts and type annotations. They must use true type casts.

falsandtru:
Essential difference:

```ts
// Expected
declare function f(): string | object;
const a = f() as object; // Correct usage of type cast

// Actual
declare function g<a extends string | object>(): a; // Wrong usage of type parameter
const b: object = g(); // Wrong usage of type annotation
```

adrai:
I tried with some overloads, but I really don't know if this is the way to go... I would really appreciate @pedrodurek help here

adrai:
image

Does not work anymore:
https://github.com/i18next/i18next/pull/1971/files#diff-62b6742de09d87fba9495aac7a5ffb538e1d3c69af885232d60adb333b676151R23

falsandtru:
That error is correct. The return type of TFunction is undetermined to be string yet.

adrai:
> That error is correct. The return type of TFunction is undetermined to be string yet.

so can you update the test?

falsandtru:
Ah no, another problem. This is a correct behavior of overloads.

```ts
// Type '() => string' is not assignable to type '{ (): string; (): object; }'.
//  Type 'string' is not assignable to type 'object'.ts(2322)
const f:{
  (): string;
  (): object;
} = () => '';
```

You have to use another interface.

falsandtru:
However, `t('translationKey').trim()` will break anyway.

falsandtru:
If you want to implement TFunction, you have to implement polymorphism. `() => string` is a subtype of TFunction, not a complete type of TFunction.

adrai:
Since I'm not a TS user, we have to wait for @pedrodurek's comment.

falsandtru:
Well, keeping all tests is very difficult or impossible because tests are based on many broken types.

falsandtru:
Reverted. Update types and tests again. I can fix it before merging.

adrai:
<v23.x.x had a lot of issues for typed translations usage... but like said, I'm not a TS expert...

falsandtru:
Additionally, do you understand that CustomTypeOptions pollute and break all other instances? Current deep type safety is fake. In fact, that breaks type safety of all other instances. See added tests.

```ts
function anotherInstance() {
  // CustomTypeOptions pollute and break all other instances.
  const i = i18next.createInstance({
    lng: 'en',
    resources: {
      en: {
        translation: {
          'a': 'b',
        },
      },
    },
  });
  i.init();

  const t = i.t;
  t('a'); // Correct but error
  t('common').foo // Wrong but no error
}
```

falsandtru:
Fundamental misdesign. Too many and wide faults to fix all.

pedrodurek:
Hey @falsandtru, just wanted to clarify that runtime and compile-time are two different things. Typescript checks types during compile-time, so it can't infer types during runtime.

If your `resources` change dynamically, you have two options:

1. Set your resources from `CustomTypeOptions` with all possible options. However, since the resources may change over time, the type safety can no longer be guaranteed.
2. Don't set resources from `CustomTypeOptions`, which means keys will be assigned as `string | string[] | TemplateStringsArray`. But, this means that the `t` function won't have type-safe and auto-completion features.

We understand that the type enhancements here may not be suitable for everyone. That's why defining the `resources` type is optional.

falsandtru:
In short, setting the resource type of CustomTypeOptions is incompatible with another instances and options? It may be ok if setting it doesn't cause the performance problems. All the performance problems are caused by setting it?

pedrodurek:
@falsandtru, yes, that's correct. When you set `resources` under `CustomTypeOptions`, we map over all object keys within `resources` to infer the `t` function keys and the exact return type. However, depending on the project size or your configuration, it's not suitable for you. We added this information in the document https://www.i18next.com/overview/typescript.

falsandtru:
Can you explain why you are trying to get us to use a feature that will break down as the project grows?

pedrodurek:
Depending on your project's size, not even typescript without the `i18next` lib will work as expected. Talking from experience 😉

Anyway, it still performs really well, you can take a look at the benchmark results here: https://github.com/i18next/i18next/pull/1911

If you're not happy with the results here or if something doesn't fit your personal needs, you can always use Module Augmentation or [patch-package](https://www.npmjs.com/package/patch-package) to write your own types. Alternatively, you can create your own solution from scratch.

Issue #1982, opened 2023-06-21
Wrong return type


falsandtru:
Explained in #1971.

```ts
// ok
const a: object = t('friend', { returnObjects: true });
// ok but should be an error
const b: string = t('friend', { returnObjects: true });
```

pedrodurek:
We could constraint the return type to be `object` only when `returnObjects: true`, but I'm against doing the same if `returnObject` is set to `true` inside the `CustomOptionTypes` since it applies globally.

Up to you to decide @adrai

adrai:
sounds reasonable

adrai:
probably also returnDetails

pedrodurek:
`returnDetails` is already behaving correctly 😄

PR #1985, opened 2023-06-22
Constrain return type when `returnObjects` is `true`

pedrodurek:
Closes https://github.com/i18next/i18next/issues/1982

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -878,14 +878,11 @@ type ParseTReturn<Key, Res> = Key extends `${infer K1}${_KeySeparator}${infer Re
 
 type TReturnOptionalNull = _ReturnNull extends true ? null : never;
 type TReturnOptionalObjects<TOpt extends TOptions> = _ReturnObjects extends true
-  ? $SpecialObject
+  ? $SpecialObject | string
   : TOpt['returnObjects'] extends true
   ? $SpecialObject
-  : never;
-type DefaultTReturn<TOpt extends TOptions> =
-  | string
-  | TReturnOptionalObjects<TOpt>
-  | TReturnOptionalNull;
+  : string;
+type DefaultTReturn<TOpt extends TOptions> = TReturnOptionalObjects<TOpt> | TReturnOptionalNull;
 
 export type TFunctionReturn<
   Ns extends Namespace,
--- a/test/typescript/t.test.ts
+++ b/test/typescript/t.test.ts
@@ -16,6 +16,8 @@ function overloadedUsage(t: TFunction) {
 function returnCasts(t: TFunction) {
   const s: string = t('friend'); // same as <string>
   const s2: string = t`friend`;
+  // @ts-expect-error
+  const s3: string = t('friend', { returnObjects: true });
   const o: object = t('friend', { returnObjects: true });
   const sa: string[] = t('friend', { returnObjects: true });
   const oa: object[] = t('friend', { returnObjects: true });


index.d.ts
```diff
@@ -878,14 +878,11 @@ type ParseTReturn<Key, Res> = Key extends `${infer K1}${_KeySeparator}${infer Re
 
 type TReturnOptionalNull = _ReturnNull extends true ? null : never;
 type TReturnOptionalObjects<TOpt extends TOptions> = _ReturnObjects extends true
-  ? $SpecialObject
+  ? $SpecialObject | string
```
falsandtru, line 881: This makes the same problem when _ReturnObjects is true.

index.d.ts
```diff
@@ -878,14 +878,11 @@ type ParseTReturn<Key, Res> = Key extends `${infer K1}${_KeySeparator}${infer Re
...
   : TOpt['returnObjects'] extends true
   ? $SpecialObject
```
falsandtru, line 883: The same change is missing. But you cannot do it because if you do, the problem is not fixed.

index.d.ts
```diff
@@ -878,14 +878,11 @@ type ParseTReturn<Key, Res> = Key extends `${infer K1}${_KeySeparator}${infer Re
...
-  : never;
-type DefaultTReturn<TOpt extends TOptions> =
-  | string
```
falsandtru, line 886: string should be kept here. Just hid the problem behind _ReturnObjects is true.

falsandtru (commented):
Constrains nothing.

index.d.ts
pedrodurek: Check my comment: https://github.com/i18next/i18next/issues/1982#issuecomment-1603288015

pedrodurek: `_ReturnObjects` is out of question. Users can access the actual value (string) if they pass the full `key`

falsandtru: Then this pr cannot close the issue.

falsandtru: Actually, there seems to be no reason to change here.

falsandtru:
Note that the same problem is not fixed with objects. They need type casts.

```ts
  // ok
  (): object => t('friend', { returnObjects: true });
  // ok
  (): string[] => t('friend', { returnObjects: true });
  // ok
  (): object[] => t('friend', { returnObjects: true });
```

index.d.ts
pedrodurek: It's not because the `returnObjects` is set to `true` in the init options (`i18next.init({returnObjects: true})`), that `t` function will return an object. if the full key is passed, it'll return a string. That's why `string | $SpecialObject` is accurate here. For example:

```tsx
const translation = {
  en: {
    namespace: {
      foo: {
        bar: ''
      }
    }
  }
}

i18next.init({
  resources: translations,
  returnObjects: true
})

// return an object
i18next.t('namespace:foo');

//return an string
i18next.t('namespace:foo.bar');
```

falsandtru: Your example doesn't seem to set _ReturnObjects true.

pedrodurek: Answer updated!

falsandtru: Still got an error. Why you thought that returned a string?

```ts
// Property 'trim' does not exist on type 'DefaultTReturn<TOpt>'.
//   Property 'trim' does not exist on type 'object'.ts(2339)
i18next.t('namespace:foo.bar').trim()
```

pedrodurek: You need to narrow it down to ensure that's not an object, but a string.
```tsx
const result = i18next.t('namespace:foo.bar');
if (typeof result === 'string') {
  result.trim();
}
```

falsandtru: Then your comment `return a string` is a lie. The problem is not fixed.

pedrodurek: Based on your assumptions and answers, it seems like you might be lacking some knowledge and making incorrect assumptions about how Typescript works. Before making groundless opinions and wrong assumptions, I suggest you take some time to read TypeScript documentation or tutorials to get your facts straight.

falsandtru: Self-conceit answer. You are writing the wrong test due to your lack of even rudimentary knowledge that the difference of type casts and type annotations.

https://github.com/i18next/i18next/blob/5fb74d516ccd6bfee4a6a3e5339ef1bdd52c55ee/test/typescript/t.test.ts#L16-L22

And you don't even understand the types you wrote.

>@[falsandtru](https://github.com/falsandtru) falsandtru [1 hour ago](https://github.com/i18next/i18next/pull/1985#discussion_r1239127245)
Your example doesn't seem to set _ReturnObjects true.

>@[pedrodurek](https://github.com/pedrodurek) pedrodurek [48 minutes ago](https://github.com/i18next/i18next/pull/1985#discussion_r1239130847)
Answer updated!

***You don't even understand the types you wrote.***

falsandtru: I'm tired of reading your code full of mistakes. Do commit the mistake-ridden code as before.

PR #1994, opened 2023-07-07
Add brand to TFunction type so different namespaces' TFunctions are not treated as compatible

### LINKED ISSUE #1941: TFunction for two different namespaces are treated as compatible
## 🐛 Bug Report

Two `TFunction` types for different namespaces (`TFunction<"foo">` and `TFunction<"bar">`) are treated as compatible by TypeScript.

## To Reproduce

[TypeScript playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwCoDECuA7AxjYEMA0cMAvnAGZQQhwBEwAjABwYCmAHjLQNwBQvAE1ZYANgEMorOCAgC0IqfWZtOtRLzhxgGGKyjkxWKQGE0AZxjUkATzCsA8mFz4z6zZskw0UDADl5IgBcFGIiZqx87nBCBvIwvgDKwbTkEBA8Gu6SZhDeRmbBCJlRmqkQhcUlUezBFlDaAOYElVWa1rUw9RhNLe4kza1wAEYSFYNRYh1dPePuQ1ONA60klSSRcCsrvOSYOHgYxKwWDAAUMMGou84YADwpabQAfACUblmsXj7EJ7TstM98LY7bDXQ4WABMZwu6BB+zuIygT1eRSiumOZwBcAA9Fi4GYABa5EQCYZSMQHPRUKC8FZAA)

```ts
import {TFunction, t} from "i18next";

declare module "i18next" {
  interface CustomTypeOptions {
    returnNull: false;
    defaultNS: "foo";
    resources: {
        foo: {
            x: string,
            y: string,
        },
        bar: {
            a: string,
            b: string,
        }
    };
  }
}

function test1(t: TFunction<"foo">) {
    return t("x");
}

function test2(t: TFunction<"bar">) {
    test1(t); // should be an error
}
```

## Expected behavior

`test1(t)` should fail to compile because `test1` accepts `TFunction<"foo">`, but `t` is `TFunction<"bar">`.

## Your Environment

- *runtime version*: N/A
- *i18next version*: 22.4.14
- *os*: N/A
- *TypeScript version*: 5.0.2

jtbandes:
Previously, two `TFunction` types would be treated as compatible even if they referred to different namespaces. For example:

```ts
function bar(t: TFunction<"bar">) { }

const { t } = useTranslation("foo");
bar(t); // wrong, but no TS error :(
```

This PR adds a property to the `TFunction` interface, making it a ["branded" type](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d). Two `TFunction`s with different values of `Ns` will no longer be treated as compatible. The example above will now fail to compile.

Fixes #1941

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -907,6 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
+  $TFunctionBrand: Ns;
   <
     Key extends ParseKeys<Ns, TOpt, KPrefix> | TemplateStringsArray,
     TOpt extends TOptions,
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -152,3 +152,10 @@ function nullTranslations() {
 //   // context + plural
 //   t('dessert', { context: 'muffin', count: 3 }).trim();
 // }
+
+function expectErrorForDifferentTFunctions(t: TFunction<'ord'>) {
+  const fn: (t: TFunction<'plurals'>) => void = () => {};
+
+  // @ts-expect-error
+  fn(t);
+}


pedrodurek (approved):
Looks great to me! No negative impact @adrai
Thank you @jtbandes!

adrai:
included in v23.2.9

efstathiosntonas:
Hi, since this change how we should/could type `t`? 

previous versions:
```tsx
import { TFunction } from "i18next";

export interface ButtonProps {
  t: TFunction;
  title: string;
}
```

usage: 

```tsx

 const { t } = useTranslation(["action_sheets", "common"]);

<Button
   t={t}
   title={t("button.close", { ns: "common" })}
 />
```

it throws this error: 
```
TS2322: Type 'TFunction<[string, string], undefined>' is not assignable to type 'TFunction<"translation", undefined>'.   Type '[string, string]' is not assignable to type '"translation"'.
```

esetnik:
I have a similar issue to @efstathiosntonas.

```ts
function bar(t: TFunction<"bar">) { 
  // inside bar() t is constrained to only use translation keys from the "bar" namespace
}

const { t } = useTranslation(["foo", "bar"]);
bar(t); // should be correct, but gives error in v23.2.10
```

adrai:
@jtbandes can you have a look at this please?

esetnik:
I believe we should deprecate `useTranslation(namespaces: [])` signature because it is really complex to typecheck and has some unexpected behavior related to which namespace becomes the default. I'd much prefer a simplification of the API that only allows loading one namespace at a time.

```ts
const { t: tOther } = useTranslation('other');
```

If we did want to support loading multiple namespaces in a single `useTranslation` call then we should eliminate the confusion around default namespace and say that whenever multiple namespaces are loaded you must always use the namespace prefix, e.g.

```
  const { t } = useTranslation(['a', 'b']); // order doesn't matter
  const value = t('a:label'); // must always use `a:` there is no default
```

adrai:
@efstathiosntonas can you try with this?

```ts
import { TFunction } from "i18next";

export interface ButtonProps {
  t: TFunction<["action_sheets", "common"]>;
  title: string;
}
```

jtbandes:
How about this? https://github.com/i18next/i18next/pull/1997

adrai:
fyi: v23.2.11 should optimize this

jokerosky:
Does anyone face with type error after this merged?

In case i want to pass TFunction as parameter:
```
function someFunction(t: TFunction<any,any>){}

....

const {t} = useTranslation([Namespaces.SomeNamespace]);
someFunction(t) ← type error happens here
```

 Property '$TFunctionBrand' is missing in type 'TFunction<Namespaces.SomeNamespace[], undefined>' but required in type 'TFunction<any, any>'.ts(2345)

adrai:
> Does anyone face with type error after this merged?
> 
> 
> 
> In case i want to pass TFunction as parameter:
> 
> ```
> 
> function someFunction(t: TFunction<any,any>){}
> 
> 
> 
> ....
> 
> 
> 
> const {t} = useTranslation([Namespaces.SomeNamespace]);
> 
> someFunction(t) ← type error happens here
> 
> ```
> 
> 
> 
>  Property '$TFunctionBrand' is missing in type 'TFunction<Namespaces.SomeNamespace[], undefined>' but required in type 'TFunction<any, any>'.ts(2345)

Make sure you use the newest i18next version and remove the any, any generics

jokerosky:
The problem has appeared after new version released ( package.json has "i18next": "^23.2.2").
Removing `any` (although why generics does not work when they are last resort for typings?) makes no effect. 
```
function someFunction(t: TFunction<'namespace'[]>) {}

const { t } = useTranslation(['namespace']);
```

this is the error message (i've replaced project path with ****)

```
Argument of type 'import("****/node_modules/react-i18next/ts4.1/index").TFunction<"namespace"[], undefined>' is not assignable to parameter of type 'import("****/node_modules/i18next/index").TFunction<"namespace"[], undefined>'.
  Property '$TFunctionBrand' is missing in type 'import("****/node_modules/react-i18next/ts4.1/index").TFunction<"namespace"[], undefined>' but required in type 'import("****/node_modules/i18next/index").TFunction<"namespace"[], undefined>'.ts(2345)
index.d.ts(910, 3): '$TFunctionBrand' is declared here.
const t: TFunction<"namespace"[], undefined>
```

adrai:
@jokerosky can you provide a reproducible github repo?
@jtbandes can you check that?

jokerosky:
Strange, in fresh project there is no error https://github.com/jokerosky/i18next-types-bug-sample/blob/main/src/App.tsx#L1 will check my configuration and other packages.

Thank you for the quick response and attention ❤️

souvikjanatw:
Does anyone face this type issue after upgrading to v23.2.11?

```
// myFunction.spec.ts file
const tMock = jest.fn();

/* the following line is giving the error for the type of `tMock`: Argument of type 'Mock<any, any>' is not assignable to parameter of type 'TFunction<"translation", undefined>'.
  Property '$TFunctionBrand' is missing in type 'Mock<any, any>' but required in type 'TFunction<"translation", undefined>'.*/

myFunction(tMock);


// myFunction.ts file 
import { TFunction } from "i18next";

function myFunction(t: TFunction){
  // definition of the function 
}
```

adrai:
@jtbandes can you help @souvikjanatw ?

jtbandes:
Can you use `myFunction(tMock as TFunction)`?

souvikjanatw:
> Can you use `myFunction(tMock as TFunction)`?

@jtbandes It's giving the following error: `Conversion of type 'Mock<any, any>' to type 'TFunction<"translation", undefined>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property '$TFunctionBrand' is missing in type 'Mock<any, any>' but required in type 'TFunction<"translation", undefined>'.`

PR #1997, opened 2023-07-12
Fix $TFunctionBrand for non-string namespaces

jtbandes:
Follow-up to #1994 
Fix type brand validation for more complex TFunction use cases (like `TFunction<["a", "b"]>`).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: `${$FirstNamespace<Ns>}`;
+  $TFunctionBrand: $IsResourcesDefined extends true ? `${$FirstNamespace<Ns>}` : never;
   <
     Key extends ParseKeys<Ns, TOpt, KPrefix> | TemplateStringsArray,
     TOpt extends TOptions,
--- a/index.d.ts
+++ b/index.d.ts
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: Ns;
+  $TFunctionBrand: [Ns] extends [string] ? `${Ns}` : never;
   <
     Key extends ParseKeys<Ns, TOpt, KPrefix> | TemplateStringsArray,
     TOpt extends TOptions,
--- a/index.d.ts
+++ b/index.d.ts
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: [Ns] extends [string] ? `${Ns}` : never;
+  $TFunctionBrand: Ns extends string ? `${Ns}` : never;
   <
     Key extends ParseKeys<Ns, TOpt, KPrefix> | TemplateStringsArray,
     TOpt extends TOptions,
--- a/index.d.ts
+++ b/index.d.ts
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: Ns extends string ? `${Ns}` : never;
+  $TFunctionBrand: `${$FirstNamespace<Ns>}`;
   <
     Key extends ParseKeys<Ns, TOpt, KPrefix> | TemplateStringsArray,
     TOpt extends TOptions,
--- a/package.json
+++ b/package.json
@@ -103,12 +103,13 @@
     "watchify": "4.0.0"
   },
   "scripts": {
-    "pretest": "npm run test:typescript && npm run test:typescript:customtypes && npm run test:typescript:fallbacktypes && npm run test:typescript:noninterop",
+    "pretest": "npm run test:typescript && npm run test:typescript:customtypes && npm run test:typescript:defaulttypes && npm run test:typescript:fallbacktypes && npm run test:typescript:noninterop",
     "lint": "eslint src",
     "test": "npm run lint && npm run test:new && npm run test:compat",
     "test:new": "karma start karma.conf.js --singleRun",
     "test:compat": "karma start karma.backward.conf.js --singleRun",
     "test:typescript": "tslint --project tsconfig.json",
+    "test:typescript:defaulttypes": "tslint --project test/typescript/default-types/tsconfig.json",
     "test:typescript:customtypes": "tslint --project test/typescript/custom-types/tsconfig.json",
     "test:typescript:fallbacktypes": "tslint --project test/typescript/fallback-types/tsconfig.json",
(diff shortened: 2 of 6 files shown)


index.d.ts
```diff
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: Ns;
+  $TFunctionBrand: [Ns] extends [string] ? `${Ns}` : never;
```
adrai: why not `$TFunctionBrand: [Ns] extends [string] ? `${Ns}` : $FirstNamespace<Ns>;` ?

index.d.ts
```diff
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: Ns;
+  $TFunctionBrand: Ns extends string ? `${Ns}` : never;
```
jtbandes, line 910: I don't really understand why this `` `${Ns}` `` is necessary, but it doesn't work if I just use `Ns` 😕

index.d.ts
jtbandes: Can you explain? Why FirstNamespace?

jtbandes: BTW you can use double-\`... like ``` `` ...`${Ns}`... `` ``` :)

adrai: https://github.com/i18next/react-i18next/blob/0a4681e428c888fe986bcc0109eb19eab6ff2eb3/src/useTranslation.js#L62

jtbandes: Looks like that depends on `i18nOptions.nsMode`, which we can't really check at compile time right?

adrai: having FirstNamespace is ok imo

adrai: and nsMode is very rarely set to fallback

jtbandes: I tried your suggestion, but it looks like it doesn't work for  the use case mentioned [here](https://github.com/i18next/i18next/pull/1994#issuecomment-1631976127) `TFunction<[string, string]>`:

image

adrai: Is ``TFunction<[string, string]>`` really a use case?
Normaly, you have typed translations so ``TFunction<['realNS', 'realOtherNS']>`` or you don't have typed translations at all

jtbandes: I don't know — hope one of the maintainers can answer that question. I was just going based on the error message here: https://github.com/i18next/i18next/pull/1994#issuecomment-1631976127

jtbandes: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwK4GcCmAVKBDAdigG2xmAlwF84AzKCEOAIijWwGMYBaYARgA5c0ADxgMA3AChxVJLnalccGGhQwAFAEpE4uHFZkViRXEoBeOKkw58REmVUBtBtgYAaRgCMGAXXUSdAen8dYLgAPQB+cXIgA

adrai: I'm one of the maintainers ;-)
Based on the usage I'll argue like this: https://github.com/i18next/i18next/pull/1994#issuecomment-1632955910
So I suggest, we make the change with FirstNamespace... and if that will get problems, we come back to your "never" solution, ok?

index.d.ts
```diff
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: Ns;
+  $TFunctionBrand: `${$FirstNamespace<Ns>}`;
```
adrai, line 910: btw: that might also fix the [string, string] case => ``$TFunctionBrand: `${$FirstNamespace<Ns> | _DefaultNamespace}`;``

jtbandes: I get this error:

image

jtbandes: If `Ns` is `[string, string]`, then `$FirstNamespace<Ns>` will be `string`, so `` `${string | _DefaultNamespace}` `` will also be `string` (I think).

adrai: you're right

adrai: the problem is in a non typed translations setup, if you're using `useTranslation('myNS')` will now return the branded TFunction with myNS

adrai: this means you always need to use TFunction<'myNS'> also in a non typed translations setup...
hmmm....

adrai: and also the never approach would not fix that

adrai: This will never work with a branded TFunction:

```ts
function untypedTFunction() {
  const [t] = useTranslation('boo');
  interface Props {
    t: TFunction
  }
  const props: Props = {
    t
  }
}
```

adrai: @pedrodurek need some expert help here...
Currently I would say we merge this PR as is or remove the branded function again.

jtbandes: In a non typed setup, maybe we should avoid branding altogether or always use `never`/`string` or some consistent type for the brand?

adrai: I think it's `$IsResourcesDefined`

jtbandes: maybe this works... `{} extends CustomTypeOptions ? true : false`

adrai: image
image
here: https://github.com/i18next/react-i18next/blob/master/index.d.ts#L65
What do you think?

jtbandes: It seems to work, what do you think of this?

```diff
diff --git a/index.d.ts b/index.d.ts
index ae1b5dd..1c40f02 100644
--- a/index.d.ts
+++ b/index.d.ts
@@ -907,7 +907,7 @@ type AppendKeyPrefix<Key, KPrefix> = KPrefix extends string
  * T function declaration *
  **************************/
 export interface TFunction<Ns extends Namespace = _DefaultNamespace, KPrefix = undefined> {
-  $TFunctionBrand: `${$FirstNamespace<Ns>}`;
+  $TFunctionBrand: 'resources' extends keyof CustomTypeOptions ? `${$FirstNamespace<Ns>}` : never;
   <
     Key extends ParseKeys<Ns, TOpt, KPrefix> | TemplateStringsArray,
     TOpt extends TOptions,
diff --git a/test/typescript/custom-types/t.test.ts b/test/typescript/custom-types/t.test.ts
index 9a8e8e6..c65c4bc 100644
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -1,4 +1,4 @@
-import i18next, { TFunction } from 'i18next';
+import i18next, { CustomTypeOptions, TFunction } from 'i18next';
 
 function defaultNamespaceUsage(t: TFunction) {
   t('bar');
diff --git a/test/typescript/default-types/t.test.ts b/test/typescript/default-types/t.test.ts
index 413d7a1..ae1a78a 100644
--- a/test/typescript/default-types/t.test.ts
+++ b/test/typescript/default-types/t.test.ts
@@ -7,12 +7,15 @@ function expectErrorsForDifferentTFunctions(
   t4: TFunction<[string, string]>,
 ) {
   const fn: (t: TFunction<'plurals'>) => void = () => {};
+  const fn2: (t: TFunction) => void = () => {};
 
-  // @ts-expect-error
+  // no error - not checked when CustomTypeOptions["resources"] is not provided
   fn(t1);
-  // @ts-expect-error
   fn(t2);
-  fn(t3); // no error
-  // @ts-expect-error
+  fn(t3);
   fn(t4);
+  fn2(t1);
+  fn2(t2);
+  fn2(t3);
+  fn2(t4);
 }
```

adrai:
Thank you @jtbandes...
it's included in v23.2.11

PR #1999, opened 2023-07-18
Restrict interpolation types based on format; Include typing for formatParams

nicegamer7:
Currently, it's possible to pass any variable as an interpolation, which isn't great since it can be error prone. Take the following as an example:
```js
// Poorly named variable
const username = { name: 'some_user', ... };

// ...

// Assuming key fallback
t("Hello {{username}}", { username });
// -> "Hello [object Object]"
```

The above is not completely addressed in this PR. Instead, I'm first fixing a subset of this problem, which is formatted interpolations. Since the built-in interpolations must conform to the `Intl` formatters that they are passed to, this PR makes it so that any interpolation that is being formatted is restriced to the type that the corresponding `Intl` formatter expects. This PR also adds typing for `formatParams`.

#### Checklist

- [x] tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -852,13 +852,33 @@ export type ParseKeys<
 /*********************************************************
  * Parse t function return type and interpolation values *
  *********************************************************/
-type ParseInterpolationValues<Ret> =
+type FormatParam<Value extends string, Format> = Format extends 'number' | 'currency'
+  ? Record<Value, Parameters<Intl.NumberFormat['format']>[0]> & {
+      formatParams?: Partial<Record<Value, Intl.NumberFormatOptions>>;
+    }
+  : Format extends 'datetime'
+  ? Record<Value, Parameters<Intl.DateTimeFormat['format']>[0]> & {
+      formatParams?: Partial<Record<Value, Intl.DateTimeFormatOptions>>;
+    }
+  : Format extends 'relativetime'
+  ? Record<Value, Parameters<Intl.RelativeTimeFormat['format']>[0]> & {
+      formatParams?: Partial<Record<Value, Intl.RelativeTimeFormatOptions>>;
+    }
+  : Format extends 'list'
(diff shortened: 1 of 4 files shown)


adrai:
@pedrodurek and also here?

adrai:
I'm not sure we should enforce the format typings... what if someone uses the old formatting or custom formats?

adrai:
@marcalexiei what is your opinion on this?

marcalexiei (commented):
I tried to port the on `master` branch. 
(This PR is very old: `InterpolationMap` is inside `t.d.ts` files, here is inside `index.d.ts` 😅)

Right now tests are failing:
<details>
<summary>test errors</summary>
image
</details>

---

The possible breaking change is that someone could have already declared formatter with these names but with different parameters like @adrai pointed out.

### Simple solution

This computation should be performed only when `_JSONFormat` is set to `v4` like in `PluralSuffix`.

https://github.com/i18next/i18next/blob/2f3384ba5bf7422e387d320f3a347cf3d28e417a/typescript/t.d.ts#L35-L37

However if someone is using a custom parser with `JSONFormat` set to `v4` the issue will remain, so the only workaround I could think of is to add a new boolean flag under `CustomTypeOptions` to enable this kind of behaviour.

I personally don't find any of this two solutions quite appealing because I consider them a workaround. 

### Complex solution

The best, but more difficult, solution is to add a way to define formatters inside `CustomTypeOptions` so every user is able to customise them and they are inferred by the `TFunction` type when the key has them.
Basically something like what is already in place with `Resources`.

Unfortunately that is a lot of type gymnastic.

----

Besides that I think that more tests should be added to test the behaviour of `TFunction` when providing a formatter using `Intl` API.

adrai:
ok, will close this for now... if someone has time and motivation to take this over feel free to reopen or create a new PR

PR #2017, opened 2023-08-08
types: add customisation of `contextSeparator`

### LINKED ISSUE #2016: Type errors when using custom `contextSeparator`
## 🐛 Bug Report

When `contextSeparator` has been customized, types will be broken when using a `t` function providing a context.

The error will be something like this: `Type at position 1 in source is not compatible with type at position 1 in target.`

Screenshot 2023-08-08 alle 15 38 18

## To Reproduce

Your i18n config must have `contextSeparator: '|'`

// the namespace – custom.json
```json
{
  "example": "Example",
  "example|custom": "Example with context"
}
```

// the react component
```typescript
const { t } = useTranslation(['custom']);

// this would result in a type error, `myValue` is `never`
const myValue = t('custom:example', { context: 'custom' });
```

@marcalexiei will open a PR with a fix and an additional test case.

## Expected behavior

There are no type errors when using a custom `contextSeparator`.
i.e. by adding a way to override the default `contextSeparator` used by types.

## Your Environment

- *runtime version*: node v18
- *i18next version*: i.e. 23.4.2
- *os*: macOS 13.4.1

marcalexiei:
Allow usage of custom `contextSeparator` in `CustomTypeOptions` for correctly resolving keys in typescript.

Added a new folder `custom-context-separator` in `test/typescript` to test this specific scenario.

Fixes #2016 

#### Checklist

- [x] tests are included

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "watchify": "4.0.0"
   },
   "scripts": {
-    "pretest": "npm run test:typescript && npm run test:typescript:customtypes && npm run test:typescript:defaulttypes && npm run test:typescript:fallbacktypes && npm run test:typescript:noninterop",
+    "pretest": "npm run test:typescript && npm run test:typescript:customtypes && npm run test:typescript:defaulttypes && npm run test:typescript:fallbacktypes && npm run test:typescript:noninterop && npm run test:typescript:customcontextseparator",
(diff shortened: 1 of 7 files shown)


adrai:
Thank you for your awesome contribution.
It has been included in v23.4.3.

Issue #2046, opened 2023-09-28
returnDetails should return parameters used to drive interpolation


robross0606:
## 🚀 Feature Proposal

When `returnDetails` is set to `true`, the returned object includes the key, language, namespace and other important details.  However, it does not include any parameter values passed to `t()` and used in interpolation.

## Motivation

In order to understand what drove a particular translation, the details include everything you would need to trace the specific resource being used to drive `res`.  Everything, that is, except interpolated parameters specifically passed into `t()`.  These values are crucial details about the resolved translation if interpolation was involved.

## Example

In our specific situation, we need to be able to "pass thru" the context data for a message from a back end GraphQL API through to the API client so it might drive their own localization upon ingesting our API responses.   In other words, if we do this:
```json
{
  "hello": {
    "world": "Hello, {{user}}!"
  }
}
```
```js
const translation = ctx.i18n.t('hello.world', { returnDetails: true, user: 'John  Smith'})
```
The results in `translation` should be something like:
```js
'{
  "res": "Hello, John Smith!",
  "usedKey": "hello.world",
  "usedParams": {
    "user": "John Smith"
   },
  "exactUsedKey": "hello.world",
  "usedLng": "en",
  "usedNS": "my-namespace"
}'
```

This is similar to `returnObjects` in concept, but more "pass thru" in that we shouldn't have to declare every parameter and the structure of the returned object in our `en.json` file. This way, we can structure a post-processor to massage the translation to match our GraphQL schema in the resolver.

PR #2053, opened 2023-10-18
add interpolation data to response if returnDetails is true

jedlikowski:
Fixes https://github.com/i18next/i18next/issues/2046

What was changed?
1. Added a new field (`usedParams`) to the object returned from `t()` when `returnDetails = true`. It should contain all the fields passed to interpolation.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -121,13 +121,21 @@ class Translator extends EventEmitter {
             exactUsedKey: key,
             usedLng: lng,
             usedNS: namespace,
+            usedParams: this.getUsedParamsDetails(options),
           };
         }
         return `${namespace}${nsSeparator}${key}`;
       }
 
       if (returnDetails) {
-        return { res: key, usedKey: key, exactUsedKey: key, usedLng: lng, usedNS: namespace };
+        return {
+          res: key,
+          usedKey: key,
+          exactUsedKey: key,
+          usedLng: lng,
+          usedNS: namespace,
+          usedParams: this.getUsedParamsDetails(options),
+        };
       }
       return key;
     }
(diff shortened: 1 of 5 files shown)


test/translator/translator.translate.spec.js
```diff
@@ -163,12 +172,81 @@ describe('Translator', () => {
...
+              testParam: 'test-param',
+              count: 1,
+            },
           },
         },
       ];
 
       tests.forEach((test) => {
         it('correctly translates for ' + JSON.stringify(test.args) + ' args', () => {
+          console.log(t.translate.apply(t, test.args));
```
adrai, line 248: Probably the console.log is not wanted?

jedlikowski: fixed

src/Translator.js
```diff
@@ -544,6 +554,46 @@ class Translator extends EventEmitter {
     return this.resourceStore.getResource(code, ns, key, options);
   }
 
+  getUsedParamsDetails(options = {}) {
+    const optionsKeys = [
```
adrai, line 558: we need to remember to extend this array if in future there will be more option properties...

jedlikowski: yeah, I was thinking about pulling those keys from somewhere to not have to remember to update it. Is there a place I could pull them from?

adrai: Can you at least create a comment there?

jedlikowski: added

adrai:
included in v23.6.0
thank you for your contribution ;-)

adrai:
done ;-)

jedlikowski:
Thank you 👍 🎉

PR #2076, opened 2023-12-06
fix(types): FormatFunction should allow any value inside options parameter

### LINKED ISSUE #2074: Custom FormatFunction type broken
## 🐛 Bug Report

First: I don't know if the title is good :)

We upgraded from `23.4.6` to `23.7.7`.
After that update, we started to get typescript errors with `Type 'unknown' cannot be used as an index type.` that in our case related to writing a custom FormatFunction:

```
import { FormatFunction } from 'i18next';

export const i18nFormatFunctionOverwrite: FormatFunction = (value, _, lng, options) => {
  if (options && !options.interpolationkey) {
    // eslint-disable-next-line no-param-reassign
    options.interpolationkey = 'value';
  }

  const passedOptions = (options && options.formatParams && options.formatParams[**options.interpolationkey**]) || {};
  //  ...
}
```
were the **bold** give the above error.

I tried to track it down and it looks to be (as far as I understand) the following change: https://github.com/i18next/i18next/pull/2018/commits/b4abeafe99172e958f59224ed90315f4f1af43c7

Also casting `options.interpolationkey` does not work, as `options.formatParams` maps to a `{}` without a key signature somehow.

Only "solution" I found yet, is giving an explicit type for the options parameter like:
`options: InterpolationOptions & $Dictionary & { formatParams: Record<string, any>; interpolationkey: string }`

## To Reproduce

Copy the above code into a project that has i18next 23.7.7 installed.

## Expected behavior

In the end I expect, to not get an error :)

FormatFunction is a very generic type, although it's known what will be passed. I see two possible options without looking deeper into i18next code and without knowing what fits better into your concepts:
- Making the `FormatFunction` type more explicit by default
- Make the `FormatFunction` a generic type `FormatFunction<>` that allows extending the Dictionary, that is passed into the options

## Your Environment

- *i18next version*: 23.7.7
- *os*: Linux

marcalexiei:
Fixes #2074

#### Followup

I might been able to develop a more sophisticated solution using interface augmentation to allow user extends interpolation options so you can get type checking and autocomplete also on `t`  alongside `FormatFunction` type.

However I never used them so I need to check how they work.

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/format.test.ts
+++ b/test/typescript/format.test.ts
@@ -1,0 +1,9 @@
+import { FormatFunction } from 'i18next';
+
+export const customFormatFunction: FormatFunction = (value, _, lng, options) => {
+  const passedOptions = options?.formatParams?.[options.interpolationkey] || {};
+
+  // All keys should can be read from options without any error
+  const randomParameter = options?.random;
+  return '';
+};
--- a/typescript/options.d.ts
+++ b/typescript/options.d.ts
@@ -138,7 +138,7 @@ export type FormatFunction = (
   value: any,
   format?: string,
   lng?: string,
-  options?: InterpolationOptions & $Dictionary,
+  options?: InterpolationOptions & $Dictionary<any>,
 ) => string;
 
 export interface InterpolationOptions {


adrai:
included in v23.7.9

chbiel:
@adrai yeah thanks for the quick fix :) looks to work!

Issue #2070, opened 2023-11-29
Typescript performance really slow when using a function with `ParseKeys` return type in `t`


Gr3q:
## 🐛 Bug Report

As title. Even if `t` expects the exact same type that I give to my function return type it probably still matches all of the items against each other, but I don't know for sure.

I included the workaround in the repro too.

## To Reproduce

* Clone https://github.com/Gr3q/i18next-slow-repro and `npm i` 
* Run `time npx tsc --noEmit --skipLibCheck`
  ```fish
   time npx tsc --noEmit --skipLibCheck
   ________________________________________________________
   Executed in    8.39 secs    fish           external
      usr time   10.61 secs  745.00 micros   10.61 secs
      sys time    0.44 secs    0.00 micros    0.44 secs
   ```
* Comment out `t(ReallySlow())`
* Run `time npx tsc --noEmit --skipLibCheck`
    ```fish
   time npx tsc --noEmit --skipLibCheck

   ________________________________________________________
   Executed in    1.60 secs    fish           external
      usr time    3.21 secs  877.00 micros    3.21 secs
      sys time    0.24 secs    0.00 micros    0.24 secs
   ```

## Expected behavior

It should not take about 7 seconds to typecheck that 1 line, probably. VSCode typescript language features (engine) also hangs for that long (basically every time I save a file in the project autocomplete/etc don't come back until typechecking finishes).

## Your Environment

- *runtime version*: node v21.2.0, typescript 5.3.2
- *i18next version*: 23.7.7
- *os*: Linux

adrai:
feel free to provide a PR to improve this

Gr3q:
God I wish I knew, I can only spare an hour or 2 to look _this year_...

Gr3q:
bump...

Gr3q:
I update the repo to the latest version, still the same (obviously)

Greg-NetDuma:
I updated the repo to the latest version testing https://github.com/i18next/i18next/pull/2095 as well. All cases I could think of (including unlikely ones) are fast now.

Time on the latest with extra lines  in the repo:
```
~/R/i18next-slow-repro►time npx tsc --skipLibCheck --noEmit                                                                                                                                                                                                                                                                                       0.932s  (main|●💩?|MERGING↑1↓1) 10:26

________________________________________________________
Executed in    3.44 secs    fish           external
   usr time    5.33 secs  321.00 micros    5.33 secs
[...]

PR #2095, opened 2023-12-30
types: optimise ParseTReturn by reading `count` and `incremental` options

marcalexiei:
I have used the `ParseKeys` type to do define some keys that are then passed to a `t` function and I noticed that when there are a lot of keys `typescript` was really slow (similar to #2070).

Looking at the types it looks like the bottleneck was `ParseTReturnPlural`. 

To reduce build time I added the `TOpt` as a new type parameter so all plural and ordinal keys can be computed only when the relative option is provided to the `t` function. According to this `ParseTReturnPlural` has been divided into two separate types:
- `ParseTReturnPlural`
- `ParseTReturnPluralOrdinal`

Each one provide they key for the relevant plural scenario

https://github.com/marcalexiei/i18next/blob/70f498d128ccfdf10a3848981eae05ca548daed5/typescript/t.d.ts#L157-L169

```ts
type ParseTReturn<
  Key,
  Res,
  TOpt extends TOptions = {},
> = Key extends `${infer K1}${_KeySeparator}${infer RestKey}`
  ? ParseTReturn<RestKey, Res[K1 & keyof Res], TOpt>
  : // Process plurals only if count is provided inside options
  TOpt['count'] extends number
  ? TOpt['ordinal'] extends boolean
    ? ParseTReturnPluralOrdinal<Res, Key>
    : ParseTReturnPlural<Res, Key>
  : // otherwise access plain key without adding plural and ordinal suffixes
    Res[Key & keyof Res];
```

### Results

running tic inside the the `many-keys` typescript test scenario produced the following results:

perf-compare

Build time was done by ~40%.

In a more simple test scenario like the one of `test:typescript:customtypes` the build time was improved by 5/10%

Test machine is a MacBook Pro (15-inch, 2016).

### Support JSONFormat v3 in typescript

I added a new test scenario for typescript where JSONFormat v3 (which is the `custom-types` scenario with few key change for plural and ordinal keys).

#### Checklist

- [x] tests are included


adrai:
wow, this is awesome thank you very much...
Did you test it also with TS v4.x ?

adrai:
> Did you test it also with TS v4.x ?

I just tested it myself... looks good.
released with v23.7.13

Grazie Marco ;-)

marcalexiei:
Sorry I was opening #2096 😅

Thank you for testing the change with TS 4.x!

PR #2097, opened 2023-12-31
Feature: migrate typescript tests to `vitest`

### LINKED ISSUE #2096: Change typescript test system
## 🚀 Feature Proposal

It's not a feature proposal but more a suggestion for managing typescript tests 😅.

After working on #2095 I think about some improvements for the typescript tests:

Right now typescript tests are performed by `tslint` in a non usual way:
E.g.: `t(....).trim()` tests if the function returns a string 😁.

Since tslint is deprecated maybe using a tool like `vitest` might give some advantage like:

* type checking support (experimental): https://vitest.dev/guide/testing-types.html#run-typechecking
* Better API to make more accurate tests: https://vitest.dev/api/expect-typeof.html 
* Workspace support (each `tests/typescript/*` scenario could be a workspace): https://vitest.dev/guide/workspace.html
  which allows to remove all `test:typescript:*` scripts to have a more clean `package.json`
* Although also karma is deprecated so in the future all tests could be migrated to vitest

If you are interested I can create a PR in the next weeks 😀.

## Motivation

Explained above

## Example

N/A

marcalexiei:
Closes #2096.

Now all typescript tests are performed using `vitest`! 

Each folder inside `test/typescript` is a workspace. 
Due to this all files that were present directly in this directory have been moved inside a `misc` directory.

To see how all works I've you can check `CONTRIBUTING.md` and `vitest.workspace.mts`

### Code changes overview

- `tsconfig.nonEsModuleInterop.json` was just used for test file inside `test/typescript` so it has been moved inside `misc` directory
- `package.json` no longer need to include a script for each ts test scenario. This is now completely handled by `vitest` workspaces
- Debug instructions have been added to `many-keys` scenario (`test/typescript/many-keys/PERF_README.md`) 
- `tsconfig.json` in the root project now longer have `include` and `exclude` keys because every test scenario overrides them.
- `vitest.workspace.mts` has been added to `.npmignore`

### Test before and after

Thanks to vitest API test which is based on type assertion now can be more accurate:
Screenshot 2023-12-31 at 11 34 00

Also having `describe` and `it` allows a better test organisation.

### Additional changes

- I changed `prettierrc` adding `json` extension and adding `$schema` key
- `mts` files have been added to prettier formatted pattern. 

#### Checklist

- [x] tests are included


adrai:
wow... nice work for the last day of the year.... your awesome.... thank you

Do you think it would make also sense to change it in here? https://github.com/i18next/react-i18next/tree/master/test/typescript

marcalexiei:
> wow... nice work for the last day of the year.... your awesome.... thank you

You're welcome!

---

> Do you think it would make also sense to change it in here?

Yes, however as next step I was considering to add `@typescript-eslint` to have a better linting on both `test.ts` and `d.ts` files.
After that I can try to migrate also the test on `react-i18next`.

PR #2101, opened 2024-01-02
fix types for array access

Gotos:
The typing for accessing arrays was broken (yielding type errors on any access) and should be fixed with this PR. This should fix #2040 - I hope I spotted all the edge cases.

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [x] tests are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

Issue #2040, opened 2023-09-21
Typescript Error: accessing array entry does not comply to key type


sreucherand:
## 🐛 Bug Report

According to what it says here: https://www.i18next.com/translation-function/objects-and-arrays, the following code should be valid in a react+typescript environment,

```tsx
const result1 = i18next.t('arrayJoin', { joinArrays: '+' });
// -> "line1+line2+line3"

const result2 = i18next.t('arrayJoinWithInterpolation', { myVar: 'interpolate', joinArrays: ' ' });
// -> "you can interpolate"

const result3 = i18next.t('arrayOfObjects.0.name');
// -> "tom"
```

but in reality, this:

```tsx
const result3 = i18next.t('arrayOfObjects.0.name');
```

returns a TS error:

```sh
src/App.tsx:12:29 - error TS2345: Argument of type '["arrayOfObjects.0.name"]' is not assignable to parameter of type '[key: "arrayJoin" | TemplateStringsArray | "arrayJoinWithInterpolation" | "arrayOfObjects" | "translation:arrayJoin" | "translation:arrayJoinWithInterpolation" | "translation:arrayOfObjects" | ("arrayJoin" | ... 5 more ... | "translation:arrayOfObjects")[], options?: (TOptionsBase & ... 1 more ... & InterpolationMap...'.
  Type '["arrayOfObjects.0.name"]' is not assignable to type '[key: "arrayJoin" | TemplateStringsArray | "arrayJoinWithInterpolation" | "arrayOfObjects" | "translation:arrayJoin" | "translation:arrayJoinWithInterpolation" | "translation:arrayOfObjects" | ("arrayJoin" | ... 5 more ... | "translation:arrayOfObjects")[], options?: (TOptionsBase & ... 1 more ... & InterpolationMap...'.
    Type at position 0 in source is not compatible with type at position 0 in target.
      Type '"arrayOfObjects.0.name"' is not assignable to type '"arrayJoin" | TemplateStringsArray | "arrayJoinWithInterpolation" | "arrayOfObjects" | "translation:arrayJoin" | "translation:arrayJoinWithInterpolation" | "translation:arrayOfObjects" | ("arrayJoin" | ... 5 more ... | "translation:arrayOfObjects")[]'. Did you mean '"arrayOfObjects"'?

12   const result3 = i18next.t('arrayOfObjects.0.name');
                               ~~~~~~~~~~~~~~~~~~~~~~~
```

## To Reproduce

[...]

Issue #2103, opened 2024-01-03
The `keyPrefix` options is not resolved correctly as of 23.7.14 


pvdstel:
## 💥 Regression Report

The `keyPrefix` option in `react-i18next` is not recognized correctly. This is caused by `i18next` itself though, not the React package.

The line

```ts
const { t } = useTranslation('translations', { keyPrefix: 'deep' });
```

is resolved to type `TFunction<"translations", "deep">` in `23.7.13` and earlier, but resolves to `TFunction<"translations", KeyPrefix<"translations">>` now.

## Last working version

Worked up to version: 23.7.13.

Stopped working in version: 23.7.14.

## To Reproduce

I've added this zip file containing a repository which reproduces the issue. It only has the TypeScript, i18next, and `@types/react` dependencies. The package manager is pnpm. Run `type-check` to see the issue. Line 20 in `index.ts` should be producing an error.

[repro.zip](https://github.com/i18next/i18next/files/13817071/repro.zip)

The relevant section is reproduced here:

```ts
const { t } = useTranslation('translations', { keyPrefix: 'deep' });
t('deep.deep');
// @ts-expect-error
t('morning'); // this line should NOT work.
```

## Expected behavior

The type error should not be occurring.

## Your Environment

- *runtime version*: Node 20.
- *i18next version*: 23.7.14
- *os*: Windows
- *TypeScript*: 5.3.3

adrai:
@Gotos may this have influenced it? https://github.com/i18next/i18next/pull/2101

Gotos:
No, it won't fix this issue. I'm looking into it :thinking:

adrai:
these lines seems to be the issue:
https://github.com/i18next/i18next/blob/master/typescript/t.d.ts#L54-L57
https://github.com/i18next/i18next/blob/master/typescript/t.v4.d.ts#L54-L57

adrai:
@marcalexiei maybe you have an idea on why this happens?

adrai:
What is really strange is, this works:

```ts
const prefixedT = i18next.getFixedT('en', 'translations', 'deep')
prefixedT('deep.deep');

// @ts-expect-error
prefixedT('morning');
```

There's only a problem in combination with useTranslation from react-i18next

adrai:
v23.7.15 should fix this

Gotos:
Wow, your super quick with merging my contributions :D 

[...]

PR #2104, opened 2024-01-03
fix types for readonly array access

Gotos:
I noticed one edge case was missing in yesterdays #2101 - it was an easy fix tho. Now the sandbox example provided in #2040 should actually be fixed as well.

#### Checklist

- [x] tests are included


adrai:
will wait for a new release after addressing #2103

PR #2105, opened 2024-01-03
fix issue with key prefixes

### LINKED ISSUE #2103: The `keyPrefix` options is not resolved correctly as of 23.7.14

Gotos:
Fixes #2103 - the test I wrote as minimal of an example as I could write. To be quite honest: I'm not sure *why* this broke or why this fixes it, but as the fix removes some unnecessary code duplication, I guess it's a good change anyway.

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/array-access/i18next.d.ts
+++ b/test/typescript/array-access/i18next.d.ts
@@ -2,7 +2,7 @@ import 'i18next';
 
 declare module 'i18next' {
   interface CustomTypeOptions {
-    defaultNS: 'main';
+    defaultNS: 'prefix';
     resources: {
       main: {
         arrayOfStrings: ['zero', 'one'];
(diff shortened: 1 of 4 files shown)


adrai:
thank you @Gotos

PR #2121, opened 2024-01-19
types: fix support for defaultNS as array (fixes #2118)

marcalexiei:
Fixes #2118.

The main difference from #2119 is that I tried to take into account all namespaces included in `defaultNS` when providing it as an array.

----

Before marking this PR ready for review I would like to see what I can do to deduce test code. 
I hope to have something ready tomorrow.

#### Checklist

- [ ] only relevant code is changed (make a diff before you submit the PR)
- [ ] run tests `npm run test`
- [ ] tests are included
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

#### Checklist (for documentation change)

- [ ] only relevant documentation part is changed (make a diff before you submit the PR)
- [ ] motivation/reason is provided
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

PR #2122, opened 2024-01-20
types: fix typo in `CustomInstanceExtensions`

marcalexiei:
During #2121 I noticed a typo in `CustomInstanceExtensions` so I corrected it.
I don't know if this should be considered a breaking change or a bug but there is no mention in the doc of this interface (https://www.i18next.com/?q=CustomInstanceExtensions), hence I think it can be considered a fix.

#### Checklist

- (N/A) tests are included

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -217,10 +217,10 @@ export interface CloneOptions extends InitOptions {
   forkResourceStore?: boolean;
 }
 
-export interface CustomInstanceExtenstions {}
+export interface CustomInstanceExtensions {}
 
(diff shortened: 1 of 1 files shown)


adrai:
Only used here: https://github.com/i18next/react-i18next/blob/26f83a89b30c8b25a0d400d45d7d63525d5f9c28/index.d.ts#L40
so also react-i18next needs to be fixed.
currently afk... will check in a couple of hours

marcalexiei:
Gotcha, I'll open also a PR there.

adrai:
thanks a lot

PR #2123, opened 2024-01-20
types: support readonly defaultNS

EvertEt:
Follow-up of #2121 to also allow defaultNS to be readonly (This is the case when using `as const` on the defaultNS definitions and using typeof defaultNS in the types)

#### Checklist

- [x] tests are included

### DIFF
--- a/typescript/helpers.d.ts
+++ b/typescript/helpers.d.ts
@@ -12,4 +12,4 @@ export type $OmitArrayKeys<Arr> = Arr extends readonly any[] ? Omit<Arr, keyof a
 
 export type $PreservedValue<Value, Fallback> = [Value] extends [never] ? Fallback : Value;
 
-export type $NormalizeIntoArray<T extends unknown | unknown[]> = T extends unknown[] ? T : [T];
+export type $NormalizeIntoArray<T extends unknown | readonly unknown[]> = T extends readonly unknown[] ? T : [T];
--- a/test/typescript/custom-types-default-ns-as-array/i18next.d.ts
+++ b/test/typescript/custom-types-default-ns-as-array/i18next.d.ts
@@ -12,7 +12,7 @@ import {
 
 declare module 'i18next' {
   interface CustomTypeOptions {
-    defaultNS: ['custom', 'custom_b'];
+    defaultNS: Readonly<['custom', 'custom_b']>;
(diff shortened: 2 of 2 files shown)


adrai:
is this ok @marcalexiei ?

typescript/helpers.d.ts
```diff
@@ -12,4 +12,4 @@ export type $OmitArrayKeys<Arr> = Arr extends readonly any[] ? Omit<Arr, keyof a
 
 export type $PreservedValue<Value, Fallback> = [Value] extends [never] ? Fallback : Value;
 
-export type $NormalizeIntoArray<T extends unknown | unknown[]> = T extends unknown[] ? T : [T];
+export type $NormalizeIntoArray<T extends unknown | readonly unknown[]> = T extends readonly unknown[] ? T : [T];
```
marcalexiei, line 15: With this fix when using a normal array as `defaultNS` the types won't work. 
You can try out removing `Readonly<...> `from the `CustomTypeOptions.defaultNS.

You have to take into account both array types. 

This should do the trick:
```suggestion
export type $NormalizeIntoArray<T extends unknown | readonly unknown[] | unknown[]> = T extends
  | unknown[]
  | readonly unknown[]
  ? T
  : [T];
```

EvertEt: What version of TS did you test this with? I tested both variants and they seem to work. See this playground https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgzgQQE5IIYgDwBUB8UC8UWUEAHsBAHYAmcUSEq1A9pQDYhQCulA1pcwDulANoBdKAH4owJF2gAuKADNUbOBABQmgMas4weo2oB5dp0IiAjABoATDYDME1HT2UDu-YYFIAtmoEUNb2TmLaoJBQqEHwyGiYkRDMykZMZhw4mgD02VBQAHqSmklQAEaxiCjoGEkpUL4BbFm5+UVAA

marcalexiei: You are right, sorry.
I tested this on my local branch but probably I had something on my working copy that broke the test 😅

marcalexiei:
@adrai LGTM

adrai:
thank you both...
it's included in v23.7.18

PR #2128, opened 2024-01-28
Improve performance by reducing array slices and RegExp recreation

hsource:
#### Checklist

- [X] tests are included

## Motivation

Our app recently switched to using i18next. It works great, but the app felt slower. We have hundreds of strings on-screen at a time, so i18next code is run a lot.

After profiling, some i18next functions were the top contributors to time taken within our 20s stacktrace:

- `off`: used by react-i18next (and base library i18next) for detaching event handlers on unmount. It looks like it calls a array.filter according to [source code](https://github.com/i18next/i18next/blob/26844609a8bfa8db7b7c62b58329f3e9a7a3694a/src/EventEmitter.js#L14). This may be iterating through a very large array
- `getLastOfPath`: 158ms - [code](https://github.com/i18next/i18next/blob/26844609a8bfa8db7b7c62b58329f3e9a7a3694a/src/utils.js#L29) does a lot of recursive retrieval and also creates a RegExp, which may take longer
- `looksLikeObjectPath`: 293ms - [code](https://github.com/i18next/i18next/blob/26844609a8bfa8db7b7c62b58329f3e9a7a3694a/src/utils.js#L138) constructs RegExp
- `getResource`: 221ms - [code](https://github.com/i18next/i18next/blob/26844609a8bfa8db7b7c62b58329f3e9a7a3694a/src/ResourceStore.js#L31) runs concat and `deepFind`, which does a lot of O(n) array operations

Looking at the causes, the 2 main causes are:

1. Linear time array operations like `.slice`, `.concat`
2. RegExp construction, which may be slow in the React Native Android runtime in particular

## Fix

- Change EventEmitter to use a Map instead of an array, changing `off` from an `O(n)` to `O(1)`
- Cache regular expression construction for `looksLikeObjectPath`
    - Store expressions in a Map, there are only 32 possible `RegExp` that can be constructed so memory is trivial
- Improve performance of `resetRegExp`
    - Rather than creating a `new RegExp` every time, just set `lastIndex = 0` to reset a global `RegExp` state
- Improve `deepFind` performance
    - Change recursive call to iterative
    - Perform less array copies and `slice` , use indexing and string appending instead of `join`
    - Changed parts from `O(n^2)` to `O(n)`
- Improve `getLastOfPath` performance
    - Don't create a `new RegExp` every time, set `lastIndex = 0`
    - Use regular indexing `O(1)` rather than `array.shift` `O(n)`
- Improve `getResource` performance
    - Don't copy arrays multiple times, just create array once and `push` items into it

## Testing

1. Added new automated tests for `deepFind` and `eventEmitter.test.ts`
2. Used a branch of the code in our app and verified the performance improvements. The time spent in translation code decreased from about 700ms to 50-100ms in total

### DIFF
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -32,19 +32,21 @@ class EventEmitter {
 
   emit(event, ...args) {
     if (this.observers[event]) {
-      for (const [observer, numTimesAdded] of this.observers[event].entries()) {
+      const cloned = Array.from(this.observers[event].entries());
+      cloned.forEach(([observer, numTimesAdded]) => {
         for (let i = 0; i < numTimesAdded; i++) {
           observer(...args);
         }
-      }
+      });
     }
 
     if (this.observers['*']) {
-      for (const [observer, numTimesAdded] of this.observers['*'].entries()) {
+      const cloned = Array.from(this.observers['*'].entries());
+      cloned.forEach(([observer, numTimesAdded]) => {
         for (let i = 0; i < numTimesAdded; i++) {
           observer.apply(observer, [event, ...args]);
         }
-      }
+      });
     }
   }
 }
--- a/src/EventEmitter.js
+++ b/src/EventEmitter.js
@@ -1,12 +1,21 @@
 class EventEmitter {
   constructor() {
+    // This is an Object containing Maps:
+    //
+    // { [event: string]: Map<listener: function, numTimesAdded: number> }
+    //
+    // We use a Map for O(1) insertion/deletion and because it can have functions as keys.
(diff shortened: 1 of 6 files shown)


adrai:
very interesting... but currently afk... not able to review this... but the tests seems to fail... can you check this?

hsource:
I just pushed a fix to the linter error for using `for..of`: I went back to using `.forEach` which does the same thing. Take a look when you're free!

adrai:
Looks good. Thank you for your valuable contribution.
v23.8.0 has just been released.

PR #2000, opened 2023-07-20
Use improved type inference with key fallback

nicegamer7:
For TypeScript users who are using key fallback, the improved type inference introduced in v23 isn't working at the moment. This PR fixes that.

#### Checklist
- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [x] tests are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

PR #2129, opened 2024-01-28
types: support stricter typechecking for returnEmptyString and returnNull

hsource:
This is an adaptation of #2000 addressing the feedback and rebasing.

#### Checklist

- [X] tests are included

## Motivation

For TypeScript users who are using key fallback (`returnEmptyString` or `returnNull` set to false), the improved type inference introduced in v23 isn't working at the moment. This PR fixes that.

## Fix

Check the `returnEmptyString` and `returnNull` settings when deciding whether we can return null/empty string or fallback to the key.

## Testing

Added tests; ran on a large codebase and verified it typechecked still


adrai:
thank you....
included in v23.8.1

hsource:
Thanks for the fast pulls!

Issue #2142, opened 2024-02-20
Ideas: improve CI


marcalexiei:
As I mentioned on #2140 I have a few suggestions / ideas I would like to propose:

## 1. CI: add format check script (current `prettier` script)

During #2140 I noticed that `typescript/helpers.d.ts` wasn't formatted with prettier.
Ideally I would like to change lint_and_test workflow by dividing it in two jobs:
1. code quality: run lint + format
2. test:  run all tests like now with an additional optimisation (see next point)

## 2. CI: Reorganise test workspaces to reduce total execution time

Avoid to execute `runtime` and `compatibility` tests two times by creating separate workspace for test with coverage and typescript tests. 
This willl result in 3 workspaces:
- `runtime`: `runtime` + `compatibility` which will collect code coverage
- `typescript`: self-explanatory
- `local`: test to be executed only on a local machine

---

- [x] CI: add format check script (current `prettier` script) (#2146)
- [x] CI: Reorganise test workspaces to reduce total execution time (#2147)

adrai:
sounds good

marcalexiei:
`t.d.ts` dedupe moved into a separate issue (#2148)

PR #2146, opened 2024-02-23
ci: add `codeQuality` and `build` jobs

marcalexiei:
- Replaced `prettier` script with `format` and `format:fix` script
- Added a `codeQuality` job as described in the issue (#2142).
- In addition I also added a `build` job to assure that also build runs smoothly.
- `pretest` script no longer runs `lint` script
---

Due to the previous changes I renamed the `lint_and_test` workflow in `CI` to better reflect its purpose.

#### Checklist

- [ ] tests are included


adrai:
Can you also adapt the README because of the github worfklow renaming? https://github.com/i18next/i18next/blob/master/README.md?plain=1#L3

btw: does the lint script not run anymore completely?

marcalexiei:
> Can you also adapt the README because of the github worfklow renaming?

Sure!

---

> btw: does the lint script not run anymore completely?

The lint script is now executed explicitly inside `codeQuality` job. 
If you wish we could run it also on `pre-push` hook via `husky`.

adrai:
Yes, maybe also pre-push would be good.

marcalexiei:
I also added `format` script within the `pre-push` hook.
Better to have them fail locally than while running an action so you can have feedback quicker.

adrai:
thx a lot

PR #2147, opened 2024-02-24
test: move ts tests in a separate workspace, collect coverage only once

marcalexiei:
Point 2 of #2142

## Typescript tests

Typescript tests have been moved in a separate workspace.
To run them now you have to run `npm run test:typescript`, as a side effect all ts workspaces no longer have `ts-` prefix.

## CI: Test 

Now coverage should be collected only once (when running node 20 matrix value).
On node 18 tests are executed without using coverage.

## ~~CI: Typescript~~

~~Now CI should be able to test types with both typescript `4` and `5`.~~

See this [comment](https://github.com/i18next/i18next/pull/2147#issuecomment-1962305128)

#### Checklist

- [x] tests are included


marcalexiei:
Looks like testing both typescript supported version is not that simple. 
It looks like that typescript  doesn't read  `typesVersion` in package.json` locally.

Will think about it and than I'll open a separate PR since it looks more complicated than I expected 😅

marcalexiei:
Besides typescript issue the optimisation performed has reduced the CI execution time by ~50%:

https://github.com/i18next/i18next/actions/workflows/CI.yml?query=is%3Asuccess

Running test with or without coverage based upon matrix value, avoiding to run `runtime` and `compatibility` project twice,
paid off 😀.

Issue #2138, opened 2024-02-16
Typescript performance regression


TimoBbz:
## 💥 Regression Report

Checking types of i18next became way more memory-intensive, causing the command `yarn tsc` to crash out of memory. Running the command `NODE_OPTIONS="--max_old_space_size=8192" yarn tsc` works but not everyone I work with has that much available RAM.

It seems that the type complexification of the 23rd version caused this regression. Is there a way to restrict the use of the t-function (for example to namespace-prefixed keys only, such as _'ns1:whatever'_), and/or a way to generate some type files based on the namespaces to prevent an expensive type-computing?

Thanks in advance!

## Last working version

Worked up to version: 22.3.0

Stopped working in version: 23

## To Reproduce

Having a basic Next project containing 100 namespaces as json files, each with 50 keys.

```ts
export const BaseComponent = () => {
  const { t } = useTranslation(['ns1', 'ns2', 'ns3', 'ns4', 'ns5', 'ns6', 'ns7', 'ns8']);
  console.log(t('ns8:lastKey'));
}
```

Running `tsc --generateTrace trace` then `npx @typescript/analyze-trace trace`
It should return the `console.log` line as very slow.

## Expected behavior

It shouldn’t be that slow…

## Your Environment

- *runtime version*: node v18.17.0
- *i18next version*: 23.8.2
- *os*: macOS Ventura
- Everything else is last version

adrai:
I'm not a TS expert... not easy to make happy everyone...
Feel free to provide a PR...

adrai:
https://github.com/i18next/i18next/issues/1883
https://github.com/i18next/i18next/issues/2042
https://github.com/i18next/i18next/issues/2091

adrai:
Feel free to provide a PR if you know how this can be improved...
//cc @marcalexiei

felixmeziere:
This issue in the typescript repo mentions i18next a lot https://github.com/microsoft/TypeScript/issues/53087. What explodes between previous versions of i18next and this is the number of type instantiations and memory usage (from `tsc -d --diagnostics).

I've tried to reproduce the conditions of my repo in a new repo but didn't succeed : it could be due to a lot of different things. Only thing I know is that with i18next version pre-v23 `tsc` runs like a charm but crashes after that...

felixmeziere:
[...]

PR #2166, opened 2024-04-06
Fix typing performance regression

felixmeziere:
https://github.com/i18next/i18next/issues/2138#issuecomment-2037843106

### DIFF
--- a/typescript/t.d.ts
+++ b/typescript/t.d.ts
@@ -116,8 +116,8 @@ type ParseKeysByFallbackNs<Keys extends $Dictionary> = _FallbackNamespace extend
   ? Keys[UnionFallbackNs]
   : Keys[_FallbackNamespace & string];
 
-type FilterKeysByContext<Keys, TOpt extends TOptions> = TOpt['context'] extends string
-  ? Keys extends `${infer Prefix}${_ContextSeparator}${TOpt['context']}${infer Suffix}`
+type FilterKeysByContext<Keys, Context> = Context extends string
+  ? Keys extends `${infer Prefix}${_ContextSeparator}${Context}${infer Suffix}`
     ? `${Prefix}${Suffix}`
     : never
   : Keys;
@@ -128,12 +128,13 @@ export type ParseKeys<
   KPrefix = undefined,
   Keys extends $Dictionary = KeysByTOptions<TOpt>,
   ActualNS extends Namespace = NsByTOptions<Ns, TOpt>,
+  const Context extends TOpt['context'] = TOpt['context'],
 > = $IsResourcesDefined extends true
   ? FilterKeysByContext<
       | ParseKeysByKeyPrefix<Keys[$FirstNamespace<ActualNS>], KPrefix>
       | ParseKeysByNamespaces<ActualNS, Keys>
       | ParseKeysByFallbackNs<Keys>,
-      TOpt
+      Context
     >
   : string;
 


typescript/t.d.ts
```diff
@@ -116,8 +116,8 @@ type ParseKeysByFallbackNs<Keys extends $Dictionary> = _FallbackNamespace extend
   ? Keys[UnionFallbackNs]
   : Keys[_FallbackNamespace & string];
 
-type FilterKeysByContext<Keys, TOpt extends TOptions> = TOpt['context'] extends string
-  ? Keys extends `${infer Prefix}${_ContextSeparator}${TOpt['context']}${infer Suffix}`
+type FilterKeysByContext<Keys, Context extends any> = Context extends string
+  ? Keys extends `${infer Prefix}${_ContextSeparator}${Context}${infer Suffix}`
     ? `${Prefix}${Suffix}`
     : never
   : Keys;
```
marcalexiei, line 123: `FilterKeysByContext` is only used by `ParseKeys` which already check that `Context`  value is a `string` before  so it can be changed to something like this:

```ts
type FilterKeysByContext<Keys, Context extends string> = 
  ? Keys extends `${infer Prefix}${_ContextSeparator}${Context}${infer Suffix}`
    ? `${Prefix}${Suffix}`
    : never;
```

With this change we can also remove `extends any` which doesn't add any type specifity

marcalexiei (changes requested):
Thanks for the time and effort to find a fix for issue #2138 @felixmeziere!

I run the test locally and no errors were found.
@adrai when you have the change could you please approve workflow run so we can double check on GitHub actions?.

Reviewing the code I found a possible improvement on `FilterKeysByContext`.
@felixmeziere Could you please try if the change works also in your project?

typescript/t.d.ts
felixmeziere: @marcalexiei thanks, actually the most important part is the `const ...` so I improved the fix further so that the code is more readable (amended the commit sorry).

felixmeziere: I applied code formatting which changed other places in the file, not sure if it's legit or not, let me know :)

marcalexiei: > actually the most important part is the const ... so I improved the fix further so that the code is more readable

👌

---

> I applied code formatting which changed other places in the file, not sure if it's legit or not, let me know :)

Strange, code base  already follows formatting rules.
Actually these changes do not follow `prettier` styles so related PR check is failing.
Try running:

```sh
npm run format:fix
```

felixmeziere: done thanks, for some reason my IDE wouldn't pick up the proper prettier config...

adrai:
Thank you... it's included in v23.11.0

typescript/t.d.ts
```diff
@@ -128,12 +128,13 @@ export type ParseKeys<
   KPrefix = undefined,
   Keys extends $Dictionary = KeysByTOptions<TOpt>,
   ActualNS extends Namespace = NsByTOptions<Ns, TOpt>,
+  const Context extends TOpt['context'] = TOpt['context'],
```
dargmuesli, line 131: The added `const` here throws an [error TS1277: 'const' modifier can only appear on a type parameter of a function, method or class](https://github.com/maevsi/stomper/actions/runs/8605657170/job/23582395255?pr=203#step:7:304)

Shin-Ogata: I'm also encountering into the same problem.

marcalexiei: This issue will be tracked via #2168

Issue #1948, opened 2023-04-25
Don't transpile to es5


MonstraG:
## 🚀 Feature Proposal

We are in 2023, and maybe we can try targeting es6 (especially because this package dropped support for IE in 3 years ago).

This can be easily done by adding 
```json
{
  "browserslist": [
    "defaults"
  ]
}
```

## Motivation

I'm using `i18next` in several applications (trough `next-i18next`), and often, `i18next` is a single biggest dependency I have at 100kb (non-minified). 

`defaults` of `browserlists` targets es6, and this single change drops esm build size  down to 81kb. (approx. -20%)

## Example

Smaller builds, faster loads worldwide.

adrai:
we'll evaluate that for the next major

gunjam:
I only use i18next serverside in a nodejs application so I'm less concerned about package size, but I would also like a way in require() a non-babeled version of i18next. Babel has introduced a lot of es6 polyfils into the package which have a big performance impact, the i18next object spread polyfil is the hottest code path in my whole application, replacing the npm installed package with a non-babled version of the library increaed my requests per second by 10%!

adrai:
prepared for next major version

PR #2181, opened 2024-05-10
perf(interpolator): use object deconstruction

VIKTORVAV99:
The code used in this file can benefit from the use of object deconstruction and optional changing. Both of these reduce the amount of property lookups the engine needs to do which should help improve performance. Perhaps more important it cuts out unneeded code and allows for better mangling by rollup, cutting down the size of the minified files a bit, without impacting functionality.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -21,7 +21,7 @@ class Interpolator {
     this.logger = baseLogger.create('interpolator');
 
     this.options = options;
-    this.format = (options.interpolation && options.interpolation.format) || ((value) => value);
+    this.format = options.interpolation?.format || ((value) => value);
     this.init(options);
   }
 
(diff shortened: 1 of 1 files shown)


adrai:
We did not use the optional chaining operator etc... because in the past there has been some i18next usage in environments where this was not supported... for example we wanted to still support Node v12 (https://github.com/i18next/i18next/issues/1948) and that node version is not able to understand that syntax...
Do you have some tests that proofs that performance is significially better with your change?

VIKTORVAV99:
> We did not use the optional chaining operator etc... because in the past there has been some i18next usage in environments where this was not supported... for example we wanted to still support Node v12 (#1948) and that node version is not able to understand that syntax... Do you have some tests that proofs that performance is significially better with your change?

I have not made any performance benchmarking as I was mostly looking to cut down the size of the generated code bundles.

I'd be happy to skip the optional chaining if that means it can be merged quicker (with just the object deconstruction) but is supporting node 12 still something that is required from your side? Node 12 has been out of LTS for almost 4 years soon.

adrai:
> is supporting node 12 still something that is required from your side?

not from my side personally, but I'm sure there are some situations in userland that still uses old node versions...

So for now, yes if you remove the optional chaining we can merge that immediately...

VIKTORVAV99:
> > is supporting node 12 still something that is required from your side?
> 
> not from my side personally, but I'm sure there are some situations in userland that still uses old node versions...
> 
> So for now, yes if you remove the optional chaining we can merge that immediately...

Will do that then and perhaps open an issue about bumping the minimum node version so it can be discussed?

adrai:
> Will do that then and perhaps open an issue about bumping the minimum node version so it can be discussed?

we can do that for the next major version... but I don't think there are a lot of benefits to make a new major version to just drop older node.js versions...

VIKTORVAV99:
> > Will do that then and perhaps open an issue about bumping the minimum node version so it can be discussed?
> 
> we can do that for the next major version... but I don't think there are a lot of benefits to make a new major version to just drop older node.js versions...

Would it be fine if I opened a PR to switch to optional chaining across all of the files then? Then you can merge it whenever you want.

adrai:
> Would it be fine if I opened a PR to switch to optional chaining across all of the files then? Then you can merge it whenever you want.

You can, but that will not be merged so soon...

VIKTORVAV99:
> > Would it be fine if I opened a PR to switch to optional chaining across all of the files then? Then you can merge it whenever you want.
> 
> You can, but that will not be merged so soon...

That's fine from my side 👍🏻

adrai:
thank you... it's included in v23.11.4

Issue #2172, opened 2024-04-12
Not existing context not detected as type error if covered by string union


stefan-schweiger:
## 🐛 Bug Report

I was told to move this issue here from the `react-i18next` repo https://github.com/i18next/react-i18next/issues/1743

When you use a string union as the context parameter it's possible to supply values for which a context does not exist and it does not result in an error. If you just use the plain string value or a const it works.

## To Reproduce

A [minimal reproducible example](https://stackblitz.com/edit/vitejs-vite-yxak5c?file=src%2FApp.tsx,src%2Fi18n.ts&terminal=dev).

```json
{
  "translation": {
    "testContext1_Test1": "EN: Context1 Test1",
    "testContext1_Test2": "EN: Context1 Test2",
  },
}
```

```js
const App = () => {
  const { t } = useTranslation('translation');

  return (
    <div>
      <p>
        {t('testContext1', { context: 'Test1' as 'Test1' | 'Test2' | 'Test3' })}
      </p>
      <p>{t('testContext1', { context: 'Test2' })}</p>
      <p>{t('testContext1', { context: 'Test2' as const })}</p>
      <p>
        {t('testContext1', { context: 'Test3' as 'Test1' | 'Test2' | 'Test3' })}
      </p>
      {/* Error: Type '{ context: "Test3"; }' is not assignable to type 'string' */}
      <p>{t('testContext1', { context: 'Test3' })}</p>
      {/* Error: Type '{ context: "Test3"; }' is not assignable to type 'string' */}
      <p>{t('testContext1', { context: 'Test3' as const })}</p>
    </div>
  );
};
```

## Expected behavior

String unions should be correctly detected as errors if they have cases which are not covered by the context.

## Your Environment

- *i18next version*: 23.11.1

marcalexiei:
Honestly I have no idea on how to fix this 😢.
If anyone has any idea on how to do this feel free to open a PR.

As a workaround you can use `enums`. You can find example in this PR: #2173.

stefan-schweiger:
@marcalexiei enums actually have the same problem, I've updated [my reproduction](https://stackblitz.com/edit/vitejs-vite-yxak5c?file=src%2FApp.tsx,src%2Fmain.tsx&terminal=dev). But basically enums only work if you assign a known value, but if you assign something which could be any of the enum values it will also happily accept it instead of correctly determining a type error.

marcalexiei:
Gotcha, so it's related, if not the same, issue of the string union. 
[...]

PR #2182, opened 2024-05-10
Fix context type checking

stefan-schweiger:
Fixes https://github.com/i18next/i18next/issues/2172

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -156,11 +156,16 @@ describe('t', () => {
 
     it('should accept a default context key as a valid `t` function key', () => {
       expectTypeOf(t('beverage')).toMatchTypeOf('cold water');
+
+      expectTypeOf(t('beverage', { context: undefined })).toMatchTypeOf('cold water');
     });
 
     it('should throw error when no `context` is provided using and the context key has no default value ', () => {
       // @ts-expect-error dessert has no default value, it needs a context
       expectTypeOf(t('dessert')).toMatchTypeOf('error');
+
+      // @ts-expect-error dessert has no default value, it needs a context
+      expectTypeOf(t('dessert', { context: undefined })).toMatchTypeOf('error');
     });
 
     it('should work with enum as a context value', () => {
@@ -174,7 +179,7 @@ describe('t', () => {
       expectTypeOf(t('dessert', { context: ctx })).toMatchTypeOf<string>();
     });
 
-    it('should trow error with string union with missing context value', () => {
+    it('should throw error with string union with missing context value', () => {
       enum DessertMissingValue {
         COOKIE = 'cookie',
         CAKE = 'cake',
@@ -182,16 +187,53 @@ describe('t', () => {
         ANOTHER = 'another',
       }
 
-      const getRandomDesert = (): DessertMissingValue =>
+      const getRandomDessert = (): DessertMissingValue =>
         Math.random() < 0.5 ? DessertMissingValue.CAKE : DessertMissingValue.ANOTHER;
 
-      const ctxRandomValue: DessertMissingValue = getRandomDesert();
+      const ctxRandomValue: DessertMissingValue = getRandomDessert();
 
       // @ts-expect-error Dessert.ANOTHER is not mapped so it must give a type error
       expectTypeOf(t('dessert', { context: ctxRandomValue })).toMatchTypeOf<string>();
 
       // @ts-expect-error Dessert.ANOTHER is not mapped so it must give a type error
       expectTypeOf(t('dessert', { context: DessertMissingValue.ANOTHER })).toMatchTypeOf<string>();
+
+      expectTypeOf(
+        // @ts-expect-error no default context so it must give a type error
+        t('dessert', { context: 'cookie' as 'cookie' | 'another' }),
+      ).toEqualTypeOf<never>();
+
+      expectTypeOf(
+        // @ts-expect-error no default context so it must give a type error
+        t('dessert', { context: 'cookie' as 'cookie' | undefined }),
+      ).toEqualTypeOf<unknown>();
+
+      expectTypeOf(
+        // @ts-expect-error no default context so it must give a type error
+        t('dessert', { context: undefined }),
+      ).toEqualTypeOf<unknown>();
+    });
+
+    it('should not throw error with string union with undefined context value if it has a default context', () => {
+      enum BeverageValue {
+        BEER = 'beer',
+        WATER = 'water',
+      }
+
+      const getRandomDessert = (): BeverageValue | undefined =>
+        Math.random() < 0.5 ? BeverageValue.BEER : undefined;
+
+      const ctxRandomValue = getRandomDessert();
+
+      expectTypeOf(
+        t('beverage', { context: ctxRandomValue }),
+      ).toMatchTypeOf<'a classic beverage'>();
+
+      expectTypeOf(
+        t('beverage', { context: 'beer' as 'beer' | 'water' | undefined }),
+      ).toEqualTypeOf<'a classic beverage'>();
+
+      expectTypeOf(t('beverage', { context: undefined })).toEqualTypeOf<'a classic beverage'>();
     });
 
     it('should work with string union as a context value', () => {
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -238,7 +238,7 @@ describe('t', () => {
     });
 
     // @see https://github.com/i18next/i18next/issues/2172
-    it('should trow error with string union with missing context value', () => {
+    it('should throw error with string union with missing context value', () => {
       expectTypeOf(
         // @ts-expect-error
         t('dessert', { context: 'muffin' as 'muffin' | 'cake' | 'pippo' }),
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -167,10 +167,10 @@ describe('t', () => {
       // @ts-expect-error dessert has no default value, it needs a context
       expectTypeOf(t('dessert', { context: undefined })).toMatchTypeOf('error');
 
-      // @ts-expect-error dessert has no default value, it needs a context
-      expectTypeOf(t('dessert', { context: false as false | '' | 0 | null })).toMatchTypeOf(
-        'error',
-      );
+      expectTypeOf(
+        // @ts-expect-error dessert has no default value, it needs a context
+        t('dessert', { context: false as false | '' | 0 | null }),
+      ).toMatchTypeOf<unknown>();
 
       // TODO: edge case which is not correctly detected currently
       // expectTypeOf(
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -181,7 +181,6 @@ describe('t', () => {
 
     it('should throw error with string union with missing context value', () => {
       enum DessertMissingValue {
-        COOKIE = 'cookie',
         CAKE = 'cake',
         MUFFIN = 'muffin',
         ANOTHER = 'another',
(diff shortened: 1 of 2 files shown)


typescript/t.d.ts
```diff
@@ -205,6 +205,19 @@ export type KeyWithContext<Key, TOpt extends TOptions> = TOpt['context'] extends
   ? `${Key & string}${_ContextSeparator}${TOpt['context']}`
   : Key;
 
+export type ContextOfKey<
```
marcalexiei, line 208: This is used only on this file so I think we can remove the `export`
```suggestion
type ContextOfKey<
```

marcalexiei (changes requested):
## Unit test

Unit tests should be added to have this scenario tested properly.
You can add them here:
https://github.com/i18next/i18next/blob/9ce9150766763bc5b36b408d21b9b5aa316d0d73/test/typescript/custom-types/t.test.ts#L143

You can update / refine the ones added via #2173.

> [!NOTE]
> Be sure to update your branch: I have added a few more tests via #2186

## Performance

I performed a benchmark on `test/typescript/many-keys` test scenario  and `tsc` time went from:
```text
npx tsc --noEmit  4.09s user 0.10s system 247% cpu 1.693 total
```
to
```text
npx tsc --noEmit  5.86s user 0.23s system 206% cpu 2.942 total
```
Nearly ~2s up (on a MacBook Air M3).

I haven't tested in a real app.

During this tests no context value has been used, so I think that some optimisation need to be performed, especially due to the issues regarding out of memory and long type-checking time.

stefan-schweiger:
> ## Unit test
> Unit tests should be added to have this scenario tested properly. You can add them here:
> 
> https://github.com/i18next/i18next/blob/9ce9150766763bc5b36b408d21b9b5aa316d0d73/test/typescript/custom-types/t.test.ts#L143
> 
> You can update / refine the ones added via #2173.
> 
> Note
> 
> Be sure to update your branch: I have added a few more tests via #2186

Added/extended tests, still all green.

> ## Performance
> I performed a benchmark on `test/typescript/many-keys` test scenario and `tsc` time went from:
> 
> ```
> npx tsc --noEmit  4.09s user 0.10s system 247% cpu 1.693 total
> ```
> 
> to
> 
> ```
> npx tsc --noEmit  5.86s user 0.23s system 206% cpu 2.942 total
> ```
> 
> Nearly ~2s up (on a MacBook Air M3).
> 
> I haven't tested in a real app.
> 
> During this tests no context value has been used, so I think that some optimisation need to be performed, especially due to the issues regarding out of memory and long type-checking time.

I have no real clue about typescript performance optimisations. If you have any concrete suggestions please go ahead.

marcalexiei:
In my real application time  goes up from  `129.95s`  to  `147.57s`.

The performance is affected whether  you provide a `context` to a `t` function or not.
In my opinion this should not be merged as it is.
Up to @adrai to decide.

---

> I have no real clue about typescript performance optimisations

You can find some information about `ts` performance here:
https://github.com/microsoft/TypeScript/wiki/Performance#performance-tracing

and you can test them in i18next here:
https://github.com/i18next/i18next/blob/master/test/typescript/many-keys/PERF_README.md

---

> If you have any concrete suggestions please go ahead.

I don't have time to perform an in depth review at this time. 
The only suggestion I have, so far, is to try to find a way to compute `ContextOfKey` only when `context` key is provided.
This is something similar to what has been done in #2095

adrai:
The performance implication is to heavy... so please @stefan-schweiger try to find a way to way to compute `ContextOfKey` only when `context` key is provided, like suggested above.

stefan-schweiger:
@marcalexiei how do you arrive at those performance numbers? running `time npx tsc --noEmit` in the `test/typescript/many-keys` takes 5.9 seconds (M1 Pro) for me no matter if my code is there or not. Is there some cache or something I'm not taking into account?

stefan-schweiger:
I've now even cloned a completely fresh master branch from your repo and I still get the same performance time. Are you sure this is an issue caused by my PR?

marcalexiei:
`upstream` remote is `i18ntext`
`context-fix` is your repository

Screenshot 2024-05-13 at 10 27 06

marcalexiei:
I'm trying on another machine

marcalexiei:
After you added [`
improve readability of typing`](https://github.com/i18next/i18next/pull/2182/commits/f82a259d69181c34b7e1de16c30df170a9c45b8f) commit the performance issue  is no longer present

marcalexiei:
Screenshot 2024-05-13 at 10 59 47

stefan-schweiger:
I'm not sure I see how moving the code around a bit would make such a difference, but I guess I'll take it. So this can be merged then?

marcalexiei:
> I'm not sure I see how moving the code around a bit would make such a difference

Actually it does: you moved the `context` validation from the `ActualOptions` type parameter to the function parameters. This likely makes ts compute `ContextOfKey`  only when necessary.

> So this can be merged then?

Since the implementation is changed I would like to perform additional checks, including a real application. 
I'll review the new implementation later since I no longer have time today.

test/typescript/custom-types/t.test.ts
```diff
@@ -174,18 +179,61 @@ describe('t', () => {
...
+      expectTypeOf(
+        // @ts-expect-error 'another' is not mapped so it must give a type error
+        t('dessert', { context: 'cake' as 'cake' | 'another' }),
+      ).toEqualTypeOf<'a nice cake'>();
+
+      // TODO: edge case which is not correctly detected currently
+      // expectTypeOf(
+      //   // @ts-expect-error no default context so it must give a type error
+      //   t('dessert', { context: 'cake' as 'cake' | undefined }),
+      // ).toEqualTypeOf<never>();
```
stefan-schweiger, line 209: @marcalexiei while using this in my actual app I found this edge case which is uncovered. Wasn't able to figure out how to correctly handle this yet, maybe you have an idea. Overall I think it's not a blocker for the PR as it's still an improvement over the current functionality, but if possible to find a solution for this as well it would be great.

stefan-schweiger: also can in theory be covered by writing code like this `value ? t('dessert', { context: value }) : t('dessert')`. Which is not ideal but helps.

marcalexiei (changes requested):
I discovered a few regressions:
1. usage with a `t` function using multiple context
2. usage with a `t` function with an explicit namespace inside the  key  

I created a PR on master with the additional test cases: #2188

stefan-schweiger:
To be honest, this is getting above my knowledge of the type system, I have tried to tinker around a bit but with no real solution in sight. I'm having a hard time to figure out how the whole NS/Prefix detection is meant to work, so if you can provide any input it would be appreciated.

stefan-schweiger:
Ok, I'm not sure I understand the solution, but somehow I got the tests to be green again, please take a look if whatever I did make sense

marcalexiei (changes requested):
@stefan-schweiger request a review from me when you are done with your changes.
I don't want to risk performing tests on outdated code or code that you are working on.

stefan-schweiger:
> @stefan-schweiger request a review from me when you are done with your changes. I don't want to risk performing tests on outdated code or code that you are working on.

I'm still trying to figure out a solution for this `t('dessert', { context: 'cake' as 'cake' | undefined })` but so far no luck, for no I guess go ahead with the review, if I find a solution for this I will do it in a separate PR

marcalexiei (commented):
## Performance 

### Real app

| Current version (23.11.4) | With the fix |
|---------------------------|--------------|
| ~120.61s                   | ~129.01s      |

I think that the performance increase is within an acceptable limit,
probably it can be improved further later.

### Many keys test scenario (no context involved)

| Current version (23.11.4) | With the fix |
|---------------------------|--------------|
| ~4.3s                   | ~4.3s      |


## Breaking changes

While testing the patch I had to make some changes in the code.

> [!NOTE]
> All the following scenarios were working with version 23.11.4

### Empty string and falsy values

Those were accepted by the previous version when the context has a default value,
however now I have to set `undefined` to remove the type error.

```diff
// touch is a boolean
- touch && 'touch',
+ touch  ? touch : undefined,
```

```diff
- typeof activeStepIndex === 'number' ? 'step' : ''
+ typeof activeStepIndex === 'number' ? 'step' : undefined
```

### Default context value

When you access a context with a variable and that context has a default value but the i18n resource hasn't mapped all the value a type error is present. 
This makes the context default value not useful in this type of scenarios.

Given the following resource structure:
```json
"theme_setting": {
  "value": "system",
  "value|dark": "dark",
  "value|light": "light"
},
```

an this code

```ts
const currentTheme = 'system' as 'dark' | 'light' | 'system';
t('generic:theme_setting.value', { context: currentTheme })
```

To remove the error I had to remove the default context value and add the missing value (system):

```diff
-   "value": "system",
+   "value|system": "system",
```

## Conclusion

These breaking changes make context type-check much more strict than  the previous version.

I don't have a solution to make the new strict validation compatible with the old ts API.

Making a release at this point would be a major bump and due to the breaking changes above it might require a migration guide (from a TS validation standpoint).

stefan-schweiger:
> ### Empty string and falsy values
> Those were accepted by the previous version when the context has a default value, however now I have to set `undefined` to remove the type error.
> 
> ```diff
> // touch is a boolean
> - touch && 'touch',
> + touch  ? touch : undefined,
> ```
> 
> ```diff
> - typeof activeStepIndex === 'number' ? 'step' : ''
> + typeof activeStepIndex === 'number' ? 'step' : undefined
> ```

I think that should be solvable by simply adding `false | '' | 0` to the type definition. Tests are still green if I do that, but I might need to add additional tests.

> ### Default context value
> When you access a context with a variable and that context has a default value but the i18n resource hasn't mapped all the value a type error is present. This makes the context default value not useful in this type of scenarios.
> 
> Given the following resource structure:
> 
> ```json
> "theme_setting": {
>   "value": "system",
>   "value|dark": "dark",
>   "value|light": "light"
> },
> ```
> 
> an this code
> 
> ```ts
> const currentTheme = 'system' as 'dark' | 'light' | 'system';
> t('generic:theme_setting.value', { context: currentTheme })
> ```
> 
> To remove the error I had to remove the default context value and add the missing value (system):
> 
> ```diff
> -   "value": "system",
> +   "value|system": "system",
> ```

This might be a bit harder, but let me try to give it a go as well.

marcalexiei:
I personally don't have any problem changing the code to provide only valid context value,
so [support falsy values](https://github.com/i18next/i18next/pull/2182/commits/c2ba23abacf2b0096ef1339d2b4b3591d55e5dd8) commit can be reverted also since your last comment: 
> Btw is was recommended to me by members of your organisation before that contexts are for known values and should not use defaults. So for the second one you could maybe even argue it's a feature... but let me know if I still should try to fix the typing for it.

Probably someone will occur in these errors while upgrading. Up to @adrai to decide. 
FWIW this can be merged for me.

stefan-schweiger:
To be honest, I never 100% agreed with the assessment that context is only for known values and you even mention the fallback case in [your documentation](https://www.i18next.com/translation-function/context). In the end whatever you decide is fine by me, but it should be reflected in the documentation as well.

adrai:
> Btw is was recommended to me by members of your organisation before that contexts are for known values and should not use defaults. So for the second one you could maybe even argue it's a feature... but let me know if I still should try to fix the typing for it.
> 
> [i18next/react-i18next#1730 (comment)](https://github.com/i18next/react-i18next/issues/1730#issuecomment-1981535589)

I think that was a misunderstanding... 
Jan said that "context is more for a known set of option", but that does not mean it can't be used without a known value... at least at runtime this for sure is not a problem.
So this is/was just a TypeScript "problem".

So @marcalexiei and @stefan-schweiger should we merge this PR and release a new version? (patch/minor/major)?

stefan-schweiger:
@adrai in this case there is still the gap mentioned by @marcalexiei

> ### Default context value
> When you access a context with a variable and that context has a default value but the i18n resource hasn't mapped all the value a type error is present. This makes the context default value not useful in this type of scenarios.
> 
> Given the following resource structure:
> 
> ```json
> "theme_setting": {
>   "value": "system",
>   "value|dark": "dark",
>   "value|light": "light"
> },
> ```
> 
> an this code
> 
> ```ts
> const currentTheme = 'system' as 'dark' | 'light' | 'system';
> t('generic:theme_setting.value', { context: currentTheme })
> ```
> 
> To remove the error I had to remove the default context value and add the missing value (system):
> 
> ```diff
> -   "value": "system",
> +   "value|system": "system",
> ```

For `t` function parameters this is relatively easy to solve, but I'm currently pulling my hair out about the `TFunctionReturn` type because `ActualKey` is always defined to include the context which might not exist in the resources. So any help from your team would be appreciated.

marcalexiei:
> So @marcalexiei and @stefan-schweiger should we merge this PR and release a new version? (patch/minor/major)?

My main concern that is holding me back from approval is that now the types make context type-checking much more stricter than the JS API.

I understand the need to have a stricter validation on all possible values of a `context`, but is really necessary if I had a default value for the context?
The more I think about this, the more I think that strict validation should be performed only when there isn't a default value for the context.

Like I told earlier I'm too busy to perform an in depth review at this time.
I can find some time to review PRs if required, like I did here, but not much more.  
At least not for the next weeks.

marcalexiei:
Added a new test for https://github.com/i18next/i18next/pull/2182#pullrequestreview-2055968571 ("Default context value" section) via #2192

adrai:
@stefan-schweiger aren't you on this anymore?

PR #2194, opened 2024-05-20
perf: trim keys once

VIKTORVAV99:
Trims the key once and reuse it rather than calling trim multiple times. Also moves the logic inside the if condition instead of using an early return since it's the only logic in there.

This should improve both the performance and bundle size a tiny bit.

#### Checklist

- [x] tests are included


adrai:
included in v23.11.5

PR #2197, opened 2024-05-22
feat(logger): initOptions.debugOptionsFilter

sterlingwes:
#### Our Issue

We're using i18next for a react-native project and we'd like to have `debug` set to true in dev mode so that we can catch issues with missing strings/keys in the logs. The problem we found is that the `initialization` log event prints a massive object to the console that, in our case, takes 3-4 full screen heights of scrolling to get through. Since this is terminal output, the object isn't interactive or collapsed like in a browser console, so there's a lot of noise in the output (`[Object]`).

#### Proposed Solution

Allow for an option (`debugOptionsFilter`) that allows for passing a function to filter the options object down to the interesting components for users. If provided, the return value would be what's logged by the logger on initialization.

#### Other Notes

I've added two new test suites since this involves `init` and I can't re-init in the existing test suite. I've also added a separate debug suite without the filter to cover the existing behaviour when the filter function is not passed in.

Some tests were failing for me locally for dates, and I assume it's a timezone issue.

If this is accepted I can update the doc site repo.

#### Questions

* should i run the build command? (i notice there's a UMD file in the root of the repo)

#### Checklist

- [x] tests are included

### DIFF
--- a/test/runtime/i18next.debug.test.js
+++ b/test/runtime/i18next.debug.test.js
@@ -1,0 +1,21 @@
+import { describe, it, expect, beforeAll, vi } from 'vitest';
+import i18next from '../../src/i18next.js';
+import logger from '../../src/logger.js';
+
+describe('i18next debug', () => {
+  let logSpy;
+  beforeAll(async () => {
+    logSpy = vi.spyOn(logger, 'log');
+    await i18next.init({
+      foo: 'bar',
+      debug: true,
+    });
+    i18next.changeLanguage('en');
+  });
+
+  describe('init log options filtering', () => {
+    it('logs a subset of the options object', () => {
+      expect(logSpy).toHaveBeenCalledWith('initialized', i18next.options);
+    });
+  });
+});
--- a/typescript/options.d.ts
+++ b/typescript/options.d.ts
@@ -351,6 +351,12 @@ export interface InitOptions<T = object> extends PluginOptions<T> {
   debug?: boolean;
 
   /**
+   * Filter function that allows changing the logged option value when `debug` is set to true.
+   * @default undefined
+   */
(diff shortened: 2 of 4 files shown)


adrai:
we will review your pr... in the meantime, why not just pass your custom logger: https://www.i18next.com/misc/creating-own-plugins#logger ?

test/runtime/i18next.debug.test.js
```diff
@@ -0,0 +1,21 @@
...
+    logSpy = vi.spyOn(logger, 'log');
+    await i18next.init({
+      foo: 'bar',
+      debug: true,
+    });
+    i18next.changeLanguage('en');
+  });
+
+  describe('init log options filtering', () => {
+    it('logs a subset of the options object', () => {
```
adrai, line 17:
```diff
-it('logs a subset of the options object', () => {
+it('logs the options object', () => {
```

typescript/options.d.ts
```diff
@@ -350,6 +350,12 @@ export interface InitOptions<T = object> extends PluginOptions<T> {
    */
   debug?: boolean;
 
+  /**
+   * Filter function that allows changing the logged option value when `debug` is set to true.
```
adrai, line 354: but only for the log statement while init()... maybe the function name needs to respect this?

adrai:
Having this `debugOptionsFilter` option is very restrictive only for the log statement done in the init()...
Passing a custom logger module is more flexibel...

sterlingwes:
Oh very cool! Didn't come across the logger override in my searches 🤦🏻‍♂️ definitely a better approach here.

I'll close this... Thank you for the super quick review 🙏🏻

PR #2212, opened 2024-07-14
Use explicit imports for utils

VIKTORVAV99:
Changes the star imports (import `*`) to explicitly import the utils that are used in each file, this helps with dead code elimination and some code in the resulting build from:
```js
  function pushPath(object, path, newValue, concat) {
    const {
      obj,
      k
    } = getLastOfPath(object, path, Object);
    obj[k] = obj[k] || [];
    if (concat) obj[k] = obj[k].concat(newValue);
    if (!concat) obj[k].push(newValue);
  }
```
to:
```js
  function pushPath(object, path, newValue, concat) {
    const {
      obj,
      k
    } = getLastOfPath(object, path, Object);
    obj[k] = obj[k] || [];
    obj[k].push(newValue);
  }
```
This is because concat is not actually passed to the util function which means the build process can remove the branching as it will always have a deterministic output.

If that changes in the future the code will once again be included so this does not affect any functionality or source code in it self.

PS: I also changed the star imports in the test so it matches the rest of the code for consistency.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/BackendConnector.js
+++ b/src/BackendConnector.js
@@ -1,4 +1,4 @@
-import * as utils from './utils.js';
+import { pushPath } from './utils.js';
 import baseLogger from './logger.js';
 import EventEmitter from './EventEmitter.js';
 
@@ -105,7 +105,7 @@ class Connector extends EventEmitter {
 
     // callback if ready
     this.queue.forEach((q) => {
-      utils.pushPath(q.loaded, [lng], ns);
+      pushPath(q.loaded, [lng], ns);
       removePending(q, name);
 
       if (err) q.errors.push(err);
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -1,4 +1,10 @@
-import * as utils from './utils.js';
+import {
+  getPathWithDefaults,
+  deepFind,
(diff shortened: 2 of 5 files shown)


adrai:
Thank you Viktor... but I don't see any difference after the build dist/esm/i18next.js looks still the same?

VIKTORVAV99:
> Thank you Viktor... but I don't see any difference after the build dist/esm/i18next.js looks still the same?

ESM build should be the same as it bundles it all in one file and then strips the unused code.

adrai:
So this is only reflected in userland projects? Do you have an example?

VIKTORVAV99:
> So this is only reflected in userland projects? Do you have an example?

The only change I saw was the one mentioned in the PR description when running `npm build` in the top level `i18next.js` and `i18next.min.js`. I did not look into if this propagated to the `udm` or `cjs` bundles as well.

adrai:
I don't see that change there:
image

adrai:
ahhh, now I see it... seems because of new dependencies

adrai:
included in v23.12.1

PR #2222, opened 2024-08-05
languageOnly with LanguageDetector bugfix

JesusChrisse:
getBestMatchFromCodes(['en-US']) with languageOnly should return 'en' but currently return 'en-US'

#### Checklist

- [Y] only relevant code is changed (make a diff before you submit the PR)
- [Y] run tests `npm run test`
- [Y] tests are included
- [Y] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

- [N/A] only relevant documentation part is changed (make a diff before you submit the PR)
- [Y] motivation/reason is provided
- [Y] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

### DIFF
--- a/i18next.js
+++ b/i18next.js
@@ -936,7 +936,7 @@
       let found;
       codes.forEach(code => {
         if (found) return;
-        const cleanedLng = this.formatLanguageCode(code);
+        const cleanedLng = this.options.load === 'languageOnly' ? this.getLanguagePartFromCode(code) : this.formatLanguageCode(code);
         if (!this.options.supportedLngs || this.isSupportedCode(cleanedLng)) found = cleanedLng;
       });
       if (!found && this.options.supportedLngs) {
--- a/src/LanguageUtils.js
+++ b/src/LanguageUtils.js
@@ -77,7 +77,10 @@ class LanguageUtil {
     // pick first supported code or if no restriction pick the first one (highest prio)
     codes.forEach((code) => {
       if (found) return;
-      const cleanedLng = this.formatLanguageCode(code);
+      const cleanedLng =
+        this.options.load === 'languageOnly'
+          ? this.getLanguagePartFromCode(code)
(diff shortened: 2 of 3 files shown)


adrai:
Can you please add a test for this?

adrai:
maybe also a real use case example of why this needs to be changed....

JesusChrisse:
Here is my config:
image

Librairies version:
- i18next@23.12.1
- i18next-browser-languagedetector@8.0.0

The bug appends when in your browser accepted language there is a language 'en-US' or similar before 'en' (or without 'en').
The work arround is to use currentOnly for my config but it is counterintuitive to let this bug occur.
Technically, this is, indeed, a logic that should be done or at least be rechecked by i18next-browser-languagedetector.
But it is counterintuitive to let 'getBestMatchFromCodes' return a complete locale when the parameter 'languageOnly' is set.

JesusChrisse:
My current setup is not able to run the npm run test since it's on windows. I'll correct the thing tomorrow

adrai:
The tests do not run because you did not change the src files but the build files...

adrai:
Please provide a minimal reproducible example of your issue first... I'm not able to reproduce any issue with your config...
Also describe exactly what the bug is? Is an unexpected http request done? is there any error or similar? Or what is exactly your issue?

adrai:
Sorry, but you're only talking about the detected language... This is not related to which language is used/resolved to load the correct translations....
So again... what is the exact issue?
If you need to display the language somewhere, you can use i18next.resolvedLanguage => https://www.i18next.com/overview/api#resolvedlanguage

JesusChrisse:
The resolvedLanguage is indeed working.... I did not know that option. Sorry for the misunderstanding and loss of time.

PR #2226, opened 2024-08-13
Cache output of getRule to optimize performance

heharm:
# Motivation
While profiling scrolling our app, we saw that this function was taking a long time. This is because in React Native and possibly other runtimes, calling Intl functions can be slow.

# Change
Cached the output of getRule to improve performance.

# Testing
Using a consistent test process of scrolling 20 seconds in our app at the same starting point and then restarting the app after each trial, we measured the cumulative time that getRule was taking. The performance improvement is significant, whereas the memory usage difference was negligible since most apps ever use a 1 to a few codes per run, this cached object is tiny. The total universe is # codes * 2, which is still very small.

  | Time 1(ms) | Time 2(ms) | Time 3(ms)
-- | -- | -- | --
Before(no caching) | 630 | 581 | 619
After(with caching) | 2 | 2 | 3

#### Checklist

- [ ] tests are included

### DIFF
--- a/i18next.js
+++ b/i18next.js
@@ -1135,28 +1135,31 @@
         this.logger.error('Your environment seems not to be Intl API compatible, use an Intl.PluralRules polyfill. Will fallback to the compatibilityJSON v3 format handling.');
       }
       this.rules = createRules();
-      this.pluralRulesCache = new Map();
+      this.pluralRulesCache = {};
     }
     addRule(lng, obj) {
       this.rules[lng] = obj;
+    }
+    clearCache() {
+      this.pluralRulesCache = {};
     }
     getRule(code) {
       let options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};
       if (this.shouldUseIntlApi()) {
         try {
-          const locales = getCleanedCode(code === 'dev' ? 'en' : code);
+          const cleanedCode = getCleanedCode(code === 'dev' ? 'en' : code);
           const type = options.ordinal ? 'ordinal' : 'cardinal';
           const cacheKey = JSON.stringify({
-            locales,
+            cleanedCode,
             type
           });
-          if (this.pluralRulesCache.has(cacheKey)) {
-            return this.pluralRulesCache.get(cacheKey);
+          if (cacheKey in this.pluralRulesCache) {
+            return this.pluralRulesCache[cacheKey];
           }
-          const rule = new Intl.PluralRules(locales, {
+          const rule = new Intl.PluralRules(cleanedCode, {
             type
           });
-          this.pluralRulesCache.set(cacheKey, rule);
+          this.pluralRulesCache[cacheKey] = rule;
           return rule;
         } catch (err) {
           return;
--- a/i18next.js
+++ b/i18next.js
@@ -1135,6 +1135,7 @@
         this.logger.error('Your environment seems not to be Intl API compatible, use an Intl.PluralRules polyfill. Will fallback to the compatibilityJSON v3 format handling.');
       }
       this.rules = createRules();
+      this.pluralRulesCache = new Map();
     }
     addRule(lng, obj) {
       this.rules[lng] = obj;
@@ -1143,9 +1144,20 @@
       let options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};
       if (this.shouldUseIntlApi()) {
         try {
-          return new Intl.PluralRules(getCleanedCode(code === 'dev' ? 'en' : code), {
-            type: options.ordinal ? 'ordinal' : 'cardinal'
+          const locales = getCleanedCode(code === 'dev' ? 'en' : code);
+          const type = options.ordinal ? 'ordinal' : 'cardinal';
+          const cacheKey = JSON.stringify({
+            locales,
+            type
           });
+          if (this.pluralRulesCache.has(cacheKey)) {
+            return this.pluralRulesCache.get(cacheKey);
+          }
+          const rule = new Intl.PluralRules(locales, {
+            type
+          });
+          this.pluralRulesCache.set(cacheKey, rule);
+          return rule;
         } catch (err) {
           return;
         }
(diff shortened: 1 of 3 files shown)


adrai:
What is the performance gain?
file size increase vs. performance?

btw. Instead of Map, use normal object dictionary.

jamuhl:
really wonder if `getRule` is calculation heavy -> might be JSON.stringify has longer duration than getting that simple function from Intl

src/PluralResolver.js
```diff
@@ -102,6 +102,7 @@ class PluralResolver {
     }
 
     this.rules = createRules();
+    this.pluralRulesCache = new Map();
```
hsource, line 105: Can you add a comment about why we created this cache? Mention the performance costs in React Native and the trade-offs

src/PluralResolver.js
```diff
@@ -111,7 +112,18 @@ class PluralResolver {
   getRule(code, options = {}) {
     if (this.shouldUseIntlApi()) {
       try {
-        return new Intl.PluralRules(getCleanedCode(code === 'dev' ? 'en' : code), { type: options.ordinal ? 'ordinal' : 'cardinal' });
+        const locales = getCleanedCode(code === 'dev' ? 'en' : code);
```
hsource, line 115: Tiny nit: Instead of `const locales`, I think naming this `const cleanedCode` would be a bit clearer, since the output is still a single (not-plural) code for a locale

src/PluralResolver.js
```diff
@@ -111,7 +112,18 @@ class PluralResolver {
...
+        const type = options.ordinal ? 'ordinal' : 'cardinal';
+        const cacheKey = JSON.stringify({locales, type});
```
hsource, line 117: Can you properly format this as `JSON.stringify({ locales, type });`

src/PluralResolver.js
```diff
@@ -111,7 +112,18 @@ class PluralResolver {
...
+        if (this.pluralRulesCache.has(cacheKey)) {
+          return this.pluralRulesCache.get(cacheKey);
+        }
+
+        const rule = new Intl.PluralRules(locales, {
+          type
+        });
```
hsource, line 124: Tiny styling nit - can you change this to just `{ type }` since it should all fit on one line? That's a bit more readable

hsource (changes requested):
@heharm submitted this after I assigned this, so here are a few pieces of feedback.

Can you edit the Pull Request to include motivation/testing sections? In the testing section, share how long this piece of code took before and after. That'd make it easier to see the performance improvements.

 A few styling nits!

Also, it seems like a test is now failing because it expects the pluralResolver to have been called multiple times. We may want to expose a way to clear the cache for testing.

typescript/t.d.ts
```diff
@@ -46,8 +46,8 @@ type WithOrWithoutPlural<Key> = _JSONFormat extends 'v4' | 'v3'
   ? Key extends `${infer KeyWithoutOrdinalPlural}${_PluralSeparator}ordinal${_PluralSeparator}${PluralSuffix}`
     ? KeyWithoutOrdinalPlural | Key
     : Key extends `${infer KeyWithoutPlural}${_PluralSeparator}${PluralSuffix}`
-    ? KeyWithoutPlural | Key
-    : Key
+      ? KeyWithoutPlural | Key
```
hsource, line 49: Tiny nit - did these come built this way? If not, I think we may want to just revert the `t.d.ts`and `t.v4.d.ts` for a simpler diff since the only changes are indentation changes

sample.html
```diff
@@ -1,46 +1,51 @@
 <html>
   <head>
-    <script src="./i18next.min.js" ></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/i18next-browser-languagedetector/7.0.2/i18nextBrowserLanguageDetector.min.js" ></script>
```
hsource, line 4: Ditto on not formatting this and reverting it

.github/workflows/CI.yml
```diff
@@ -101,5 +101,5 @@ jobs:
       - name: Install dependencies
         run: npm install
 
-      - name: Test 
+      - name: Test
```
hsource, line 104: Ditto here

.github/stale.yml
```diff
@@ -1,18 +1,17 @@
...
 exemptLabels:
-  - "discussion"
-  - "feature request"
-  - "bug"
-  - "breaking change"
-  - "doc"
-  - "issue"
-  - "help wanted"
-  - "good first issue"
+  - 'discussion'
```
hsource, line 7: Ditto here

hsource (changes requested):
In general, I think it's best to avoid changes in files you didn't touch, so I recommend reverting all of the formatting-only changes you made

heharm:
I have edited the original description with more details including testing results for performance gain. 

> # Motivation
> While profiling scrolling our app, we saw that this function was taking a long time. This is because in React Native and possibly other runtimes, calling Intl functions can be slow.
> 
> # Change
> Cached the output of getRule to improve performance.
> 
> # Testing
> Using a consistent test process of scrolling 20 seconds in our app at the same starting point and then restarting the app after each trial, we measured the cumulative time that getRule was taking. The performance improvement is significant, whereas the memory usage difference was negligible since most apps ever use a 1 to a few codes per run, this cached object is tiny. The total universe is # codes * 2, which is still very small.
> 
>  	Time 1(ms)	Time 2(ms)	Time 3(ms)
> Before(no caching)	630	581	619
> After(with caching)	2	2	3
> #### Checklist
> * [x]  only relevant code is changed (make a diff before you submit the PR)
> * [x]  run tests `npm run test`
> * [ ]  tests are included
> * [x]  commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)
> 
> #### Checklist (for documentation change)
> * [ ]  only relevant documentation part is changed (make a diff before you submit the PR)
> * [ ]  motivation/reason is provided
> * [ ]  commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

adrai:
@hsource as soon as you approve, I'll merge

i18next.js
adrai, line 1: this i18next.js file does not need to be committed

hsource (approved):
LGTM!

adrai:
included in v23.13.0
Thank you @heharm  and @hsource

PR #2241, opened 2024-09-24
Chore/add warning on invalid fallback lang

AlemTuzlak:
I've tried to implement the warning for the issue outlined in the ticket below, although I am not sure about the following:
1. How do I test this and where? There are a lot of test files and I don't know what the best place is and what to use (mocks, spies etc)
2. Is this if check okay or could I have used optional chaining? I am not sure so I just checked if it exists and has a length
3. should the "dev" be ignored?
https://github.com/i18next/react-i18next/issues/1803

#### Checklist

- x] run tests `npm run test`
- [ ] tests are included


adrai:
1. Create a new test file and pass in a logger dummy like here: https://github.com/i18next/i18next/blob/master/test/runtime/i18next.hasLoadedNamespace.test.js#L4
2. if check is ok... no optional chaining
3. why should dev be ignored...

`fallbackLng` can be a string an array or an object: https://www.i18next.com/principles/fallback#fallback-to-different-languages

AlemTuzlak:
Some tests pass "dev" as a fallback even though it's not in supported languages so I wasn't sure if there was some behavior I'm not aware of, okay will fix what you just said!

adrai:
> Some tests pass "dev" as a fallback even though it's not in supported languages so I wasn't sure if there was some behavior I'm not aware of, okay will fix what you just said!

ok, keep the exception with the dev language

AlemTuzlak:
so I found something interesting, while trying to write the tests I actually got 2 logs from module called languageUtils that warned that the default language is not in the supported languages list, it seems to me that the feature is already there, maybe it's not called in the react package?

adrai:
totally forgot that... yes... it's already there ;-)

Yes, it should be called also via react-i18next...

adrai:
Just tested here (change the fallbackLng accordingly): https://codesandbox.io/p/sandbox/react-i18next-http-example-forked-n4fqct

seems to work:
image

So this PR can be closed, right?

AlemTuzlak:
I'll have to investigate this on my side, maybe it just doesn't trigger in Remix for some remix related reason. I'll raise the issue again if I figure out the root cause

PR #2243, opened 2024-10-05
fix(types): make context related keys detection stricter

### LINKED ISSUE #2242: [Typescript] Wrong key type when using context on TS5+
## 🐛 Bug Report

**Typescript 5+** infers wrong keys when using keys with context. 
On TS4, everything works as expected.
Screenshot 2024-10-05 at 00 15 22

## To Reproduce

```ts
import i18next from 'i18next';

const resources = {
    translation: {
        key: 'key value',
        key_context: 'key_context value',
    }
};

i18next.init({
    supportedLngs: ['en'],
    lng: 'en',
    resources: {
        en: resources,
    },
});

declare module 'i18next' {
    interface CustomTypeOptions {
        resources: typeof resources,
    }
}

const t = i18next.getFixedT('en', 'translation');

const test = t('keyontext');
//              ^ this is a valid key!
```

## Expected behavior

Only `key` and `key_context` keys should be inferred. 

## Your Environment
- *i18next version*: 23.15.1
- *typescript*: 5.6.2

marcalexiei:
Close #2242

----

Now `FilterKeysByContext` is more strict and should not suggest "stripped" keys  like reported in #2242

---

### Limitation

Considering the following namespace:

```ts
export type TestNamespaceContext = {
  dessert_cake: 'a nice cake';
  dessert_muffin_one: 'a nice muffin';
  dessert_muffin_other: '{{count}} nice muffins';

  beverage: 'a classic beverage';
  beverage_beer: 'fresh beer';
  beverage_water: 'cold water';
};
```

`dessert` context related keys have no default and with the current type architecture 
the `dessert` key can't be showed in suggestion. On the contrary all context values can.

E.g.:
image

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/custom-types/t.test.ts
+++ b/test/typescript/custom-types/t.test.ts
@@ -175,9 +175,16 @@ describe('t', () => {
         MUFFIN = 'muffin',
       }
 
-      const ctx = Dessert.CAKE;
+      expectTypeOf(t('dessert', { context: Dessert.CAKE })).toMatchTypeOf<string>();
 
+      const ctx: Dessert = Dessert.CAKE;
       expectTypeOf(t('dessert', { context: ctx })).toMatchTypeOf<string>();
+    });
+
+    it('should not provide partial match as option when keys include context separator', () => {
+      // https://github.com/i18next/i18next/issues/2242
+      // @ts-expect-error
+      t('beverageater');
     });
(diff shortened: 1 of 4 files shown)


adrai:
Thanks a lot @marcalexiei

PR #2244, opened 2024-10-13
perf: use Intl.getCanonicalLocale

VIKTORVAV99:
Since this code was created 9 years ago there has been standardisation around this and it's now part of the Intl object.

Replacing this should have no impact on support as it has had full browser support for several years now: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/getCanonicalLocales

This reduces the size by 519 bytes (`49 096 bytes -> 48 577 bytes`) and should be faster as it has a native implementation.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/LanguageUtils.js
+++ b/src/LanguageUtils.js
@@ -1,8 +1,6 @@
 import baseLogger from './logger.js';
 import { getCleanedCode } from './utils.js';
 
-const capitalize = (string) => string.charAt(0).toUpperCase() + string.slice(1);
-
 class LanguageUtil {
   constructor(options) {
     this.options = options;
@@ -33,28 +31,13 @@ class LanguageUtil {
   formatLanguageCode(code) {
     // http://www.iana.org/assignments/language-tags/language-tags.xhtml
     if (typeof code === 'string' && code.indexOf('-') > -1) {
-      const specialCases = ['hans', 'hant', 'latn', 'cyrl', 'cans', 'mong', 'arab'];
-      let p = code.split('-');
+      let formattedCode = Intl.getCanonicalLocales(code)[0];
 
       if (this.options.lowerCaseLng) {
-        p = p.map((part) => part.toLowerCase());
-      } else if (p.length === 2) {
-        p[0] = p[0].toLowerCase();
-        p[1] = p[1].toUpperCase();
-
-        if (specialCases.indexOf(p[1].toLowerCase()) > -1) p[1] = capitalize(p[1].toLowerCase());
-      } else if (p.length === 3) {
-        p[0] = p[0].toLowerCase();
-
-        // if length 2 guess it's a country
-        if (p[1].length === 2) p[1] = p[1].toUpperCase();
-        if (p[0] !== 'sgn' && p[2].length === 2) p[2] = p[2].toUpperCase();
-
-        if (specialCases.indexOf(p[1].toLowerCase()) > -1) p[1] = capitalize(p[1].toLowerCase());
-        if (specialCases.indexOf(p[2].toLowerCase()) > -1) p[2] = capitalize(p[2].toLowerCase());
+        formattedCode = formattedCode.toLowerCase();
       }
 
-      return p.join('-');
+      return formattedCode;
     }
 
     return this.options.cleanCode || this.options.lowerCaseLng ? code.toLowerCase() : code;
--- a/src/utils.js
+++ b/src/utils.js
@@ -246,7 +246,4 @@ export const deepFind = (obj, path, keySeparator = '.') => {
   return current;
 };
 
-export const getCleanedCode = (code) => {
-  if (code && code.indexOf('_') > 0) return code.replace('_', '-');
-  return code;
-};
+export const getCleanedCode = (code) => code && code.replace('_', '-');


adrai:
as long as there is V3 support that does not use the Intl api => https://github.com/i18next/i18next/blob/master/src/PluralResolver.js#L99 we can't merge this...
but we can consider this like (https://github.com/i18next/i18next/pull/2184) for the next major version

VIKTORVAV99:
> as long as there is V3 support that does not use the Intl api => https://github.com/i18next/i18next/blob/master/src/PluralResolver.js#L99 we can't merge this... but we can consider this like (#2184) for the next major version

Intl.getCanonicalLocale is a lot older than plural rules so even then it should be supported IMO. But yeah probably best to put it in a new major version. Might I also suggest removing the  `lowerCaseLanguage` option when that is done? (this one: `this.options.lowerCaseLng` I don't think it's adding any value).

adrai:
Just released v23.16.0 that will use `Intl.getCanonicalLocale` if available...

adrai:
regarding the lowerCaseLng... this was a user request long time ago... and part of i18next since v1.2.3... isn't it @jamuhl ?

VIKTORVAV99:
> Just released v23.16.0 that will use `Intl.getCanonicalLocale` if available...

Just as a FYI, it would be really nice if you used PRs so other people like myself can watch for new PRs. Commits to master does not have this capability.

VIKTORVAV99:
I would for example have loved to provide some feedback as the `typeof !== 'undefined'` checks are not needed in this case as a simple truthy check would suffice. typeof functions and comparisons are slower than a simple true check.

adrai:
> Just as a FYI, it would be really nice if you used PRs so other people like myself can watch for new PRs. Commits to master does not have this capability.

you're right... next time..

> I would for example have loved to provide some feedback as the typeof !== 'undefined' checks are not needed in this case as a simple truthy check would suffice. typeof functions and comparisons are slower than a simple true check.

Unfortunately, in some environments those truthy checks did not work... we experienced this also here: https://github.com/i18next/i18next/blob/master/src/i18next.js#L94

All these different environments and setups etc... are the reason why we want to keep the code compatibility as wide as possible... and not always use the newest syntax features etc...

VIKTORVAV99:
> > Just as a FYI, it would be really nice if you used PRs so other people like myself can watch for new PRs. Commits to master does not have this capability.
> 
> 
> 
> you're right... next time..
> 
> 
> 
> > I would for example have loved to provide some feedback as the typeof !== 'undefined' checks are not needed in this case as a simple truthy check would suffice. typeof functions and comparisons are slower than a simple true check.
> 
> 
> 
> Unfortunately, in some environments those truthy checks did not work... we experienced this also here: https://github.com/i18next/i18next/blob/master/src/i18next.js#L94
> 
> 
> 
> All these different environments and setups etc... are the reason why we want to keep the code compatibility as wide as possible... and not always use the newest syntax features etc...

I'm really curious to hear about which environments have defined internal objects that resolve truthy if they don't exist. If those exist they are not actually JavaScript compliant and I don't think anyone should go out of their way to support them.

And while I agree that the newest syntax should not always be used I'd say it's fair game to use any feature that is supported by all supported browsers and engines. If end users want to support older stuff than that I'd say it's up to them to enable that either by using older versions of the package or polyfiling/transpiling the code.

It should not be up to you as a maintainer to support 10 year old browsers.

But that's just my take on it.

jamuhl:
not all companies are fast movers - most are not, most big companies are not...locking them out just because of the latest greatest can be done - but should be done wisely (there is nothing worse than getting major versions every few months - breaking your codebase)...

adrai:
> It should not be up to you as a maintainer to support 10 year old browsers.

i18next is not only used in browsers... but also react-native, hexo, server side, meteor... etc...

adrai:
btw: [most users will just downgrade](https://github.com/i18next/react-i18next/issues/1797) to still continue to work with i18next... and if they find some bugs or missing functionality etc... we are then forced to address that also on older i18next versions... more work for everyone, just because of new syntax etc...

VIKTORVAV99:
> not all companies are fast movers - most are not, most big companies are not...locking them out just because of the latest greatest can be done - but should be done wisely (there is nothing worse than getting major versions every few months - breaking your codebase)...

That's the great part about pinning dependencies. Everything would work exactly as it does now for them. I don't see a problem here. 

If they only use the latest all the time that's on them.

> i18next is not only used in browsers... but also react-native, hexo, server side, meteor... etc...

Intl is supported in react-native, node, deno and other JS runtimes as well and have been for a long time.

It reached baseline support in September 2017.

> btw: https://github.com/i18next/react-i18next/issues/1797 to still continue to work with i18next... and if they find some bugs or missing functionality etc... we are then forced to address that also on older i18next versions... more work for everyone, just because of new syntax etc...

Which is my point, if they need to downgrade, from a major version they shouldn't have updated in the first place. Major versions indicate breaking changes and therefore extra care should be taken to ensure it works. Or to pin the dependency to the last supported version.

Newer applications and users should not have to suffer from bloated code or performance hits just because unsupported environments don't support a built in object or newer syntax.

For example i18n is 48.9 kb minified while Lingui is 7 kb, that is a lot of extra stuff for two libraries that aim to do the same thing.

adrai:
> Newer applications and users should not have to suffer from bloated code or performance hits just because unsupported environments don't support a built in object or newer syntax.

If bloated code or performance really hits in real world scenarios, I'm full ok with removing stuff for the newer (more modern) environments... but if it only brings a marginal improvement, then it's better not to cut off the user base straight away...

> For example i18next is 48.9 kb minified while Lingui is 7 kb, that is a lot of extra stuff for two libraries that aim to do the same thing.
i18next has a completely different approach compared to Lingui... and btw: what are 50kb vs 7kb when your application loads mbs of data...

But in general, I'm with you...
I already told you... for the next major, we will remove/change some stuff here...

VIKTORVAV99:
> i18next has a completely different approach compared to Lingui... and btw: what are 50kb vs 7kb when your application loads mbs of data...

Because we built a SPA that bundles all the languages and core libraries and only connect to our backend to get production data, not translations. But it's also an indicator that there is a lot more operations that are going on than there needs to be, which I'm more concerned about.

More importantly our app bundles are split up and i18next is included in the core bundle as it's used everywhere in our application. The core bundle is currently `286.85 kB` so removing `43 kb` from that (including any extra reductions from the react package) would mean lowering our index bundle by 15%. Since that bundle is blocking all other code from executing that is a pretty significant change.

jamuhl:
Sure smaller is better in general...but does your SPA really suffer from those 14kB i18next adds (if loaded gzipped)? Or could the time be invested better somewhere else?

I mean our app has a core module size of 900kB which gets cached after first load - and even if not - takes like 70ms to load. But we might be lucky our app is not used in environments having bad internet.

VIKTORVAV99:
> Sure smaller is better in general...but does your SPA really suffer from those 14kB i18next adds (if loaded gzipped)? Or could the time be invested better somewhere else?
> 
> I mean our app has a core module size of 900kB which gets cached after first load - and even if not - takes like 70ms to load. But we might be lucky our app is not used in environments having bad internet.

Our app is very much used in mobile envs and there is for sure other optimisations that can be done (and are being done) but these are outside the index bundle. So while they are important and still are being done they don't have the same impact on first load as the core packages like i18next.

jamuhl:
my last 50cents...looks like such changes have a wider impact as estimated...https://github.com/i18next/react-i18next/issues/1797#issuecomment-2409989919 comes to mind

if such changes break installments - I would consider not publishing this as just "patch" versions - but possible breaking major

PR #2247, opened 2024-10-19
perf: Optimize size of Formatter

VIKTORVAV99:
Some simple size optimizations, there are no breaking logic or syntax changes.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/Formatter.js
+++ b/src/Formatter.js
@@ -77,37 +77,30 @@ class Formatter {
 
     this.options = options;
     this.formats = {
-      number: createCachedFormatter((lng, opt) => {
-        const formatter = new Intl.NumberFormat(lng, { ...opt });
-        return (val) => formatter.format(val);
-      }),
-      currency: createCachedFormatter((lng, opt) => {
-        const formatter = new Intl.NumberFormat(lng, { ...opt, style: 'currency' });
-        return (val) => formatter.format(val);
-      }),
-      datetime: createCachedFormatter((lng, opt) => {
-        const formatter = new Intl.DateTimeFormat(lng, { ...opt });
-        return (val) => formatter.format(val);
-      }),
-      relativetime: createCachedFormatter((lng, opt) => {
-        const formatter = new Intl.RelativeTimeFormat(lng, { ...opt });
-        return (val) => formatter.format(val, opt.range || 'day');
-      }),
-      list: createCachedFormatter((lng, opt) => {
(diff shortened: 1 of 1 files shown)


adrai:
thx, included in v23.16.1

VIKTORVAV99:
> @VIKTORVAV99 @adrai This change might reduce code size, but it decreases runtime performance, as the formatters are not cached anymore outside the returned functions as before.
> 
> Before:
> 
> ```js
> number: createCachedFormatter((lng, opt) => {
>   const formatter = new Intl.NumberFormat(lng, { ...opt });
>   return (val) => formatter.format(val);
> }),
> ```
> 
> Returned arrow function uses same formatter on every call. This is even indicated by the function name "createCachedFormatter"!
> 
> After:
> 
> ```js
> number: createCachedFormatter(
>   (lng, opt) => (val) => new Intl.NumberFormat(lng, { ...opt }).format(val),
> ),
> ```
> 
> Returned arrow function creates new formatter on every call.

Do you have a way of actually showing they are not cached anymore? Logically it should be the same as before as new Intl.** was called inside the function return clause before as well, this simply removes the intermediate assignment to a const to take advantage of implicit returns.

jamuhl:
https://github.com/i18next/i18next/blob/master/src/Formatter.js#L67

before: it cached the returned function `return (val) => formatter.format(val);` which used the upfront created Intl.*** formatter in it's scope function

`(val) => formatter.format(val);`

now: it creates a new Intl.** formatter and calls format on it...

`(val) => new Intl.NumberFormat(lng, { ...opt }).format(val)`

danielrentz:
I mean calling the inner function that will be returned from the outer function will create a fresh formatter on every call now.

Before, `formats.number` was a function that uses the same formatter living in the closure of the "createCachedFormatter" call:
```js
formats.number === (val) => formatter.format(val)
```
where `formatter` is always the same (created once when the function "createCachedFormatter" has executed once when running the `Formatter` constructor).

Now, `formats.number` is a function that creates a new formatter on every call:
```js
formats.number === (val) => new Intl.NumberFormat(lng, { ...opt }).format(val)
```

VIKTORVAV99:
Hmm I see, might be good to revert that part of the PR then or refractor it so it only returns the outer function but don't call it and then just uses the cached inline function with a call signature instead.

adrai:
Sorry, I did not see that, when reviewing the PR... it's my fault.
Now reverted, and fixed with v23.16.4

PR #2248, opened 2024-10-20
perf(size): Adds a isString utility

VIKTORVAV99:
Adds a isString utility just like in https://github.com/i18next/react-i18next/pull/1763 to reduce code duplication. This alone reduces the size of the minified output by roughly 0.5 kB.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/BackendConnector.js
+++ b/src/BackendConnector.js
@@ -1,4 +1,4 @@
-import { pushPath } from './utils.js';
+import { pushPath, isString } from './utils.js';
 import baseLogger from './logger.js';
 import EventEmitter from './EventEmitter.js';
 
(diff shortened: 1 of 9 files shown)


VIKTORVAV99:
On a related note is it possible import helpers from the core i18next package (this one) in the framework specific ones?

adrai:
> On a related note is it possible import helpers from the core i18next package (this one) in the framework specific ones?

theoretically yes, but it's not part of the "public" i18next api (yet)

adrai:
included in v23.16.2

VIKTORVAV99:
> > On a related note is it possible import helpers from the core i18next package (this one) in the framework specific ones?
> 
> theoretically yes, but it's not part of the "public" i18next api (yet)

Might be nice to do so loggers and utilities can be reused without much effort.

adrai:
logger is already used via this.services

VIKTORVAV99:
> logger is already used via this.services

Then why is react-i18next defining its own warning functions in its utility file?

adrai:
I think because most warning needs only to be warned once... and there is not alway access to the i18next core... (am I right, @jamuhl ?)

adrai:
normal i18next plugins uses the core logger: https://github.com/i18next/i18next-http-backend/blob/master/lib/index.js#L157

jamuhl:
logger is more about using init option debug: true -> get information when needed (mute in production)

if it's a regular warning (like wrong usage of useTranslation before init) there is no need to make that dependent on debug true...it's a basic console.warn

PR #2257, opened 2024-11-20
types: remove JSONFormat v3 support from types

### LINKED ISSUE #2254: New major release: drop v3 typescript support
I checked [i18next-drop-legacy branch](https://github.com/i18next/i18next/tree/i18next-drop-legacy) changeling and I found:

https://github.com/i18next/i18next/blob/92a36d3f07750e3cb2d961110cbfc601f5eeafda/CHANGELOG.md?plain=1#L6-L7

Could the associated types and tests  of json v3 be removed from the source code?

marcalexiei:
Close #2254

#### Checklist

- [x] tests are included


adrai:
Should we set this to v4 only ?
https://github.com/i18next/i18next/blob/i18next-drop-legacy/typescript/options.d.ts#L628

marcalexiei:
Assuming this [comment](https://github.com/i18next/i18next/issues/2254#issuecomment-2488561881) applies also to `compatibilityJSON` option we surely can reduce the type to `v4`.

However `v4` match the only available and default value of `jsonFormat` which brings the question: 
is this option still useful in the next major? 
If not I would consider adding a message informing the user that the typescript types are not supported for older version.
If a new version comes out we can extend the type and remove the message.

adrai:
yes, I have no idea what that is

marcalexiei:
I check the doc and I found it: https://www.i18next.com/overview/typescript#custom-type-options
It an option related only to typescript:

image

However it seems to mirror `compatibilityJSON`. 
I need some time to see when and why it has been introduced to understand if it can be removed / replaced with `compatibilityJSON`.

marcalexiei:
Since `jsonFormat` is only related to `typescript` and `compatibilityJSON` has some runtime usage (with resolver and backward compatibility utils), I think that `jsonFormat` can be replaced with `compatibilityJSON `.

Effects on user side should be minimal but I would rather add a note into the migration guide.
I also need to remove `jsonFormat` from the documentation.

---

If there are no objection I can open a PR in the documentation repository.
Also does make sense to add  `compatibilityJSON` to https://www.i18next.com/overview/configuration-options?

adrai:
Yes, that sounds good

adrai:
@marcalexiei let me know if you see something that needs to be done (or could be done) in that branch for the next major... if not, I'll create a major version...

marcalexiei:
I still have to work on the pr doc. 
I’ll open it in an hour, then I’ll recheck if something else can be done.

adrai:
will do the docs after the release... ;-)
but, will wait for your OK

marcalexiei:
No further action required by me, you can release.

Issue #2267, opened 2024-12-03
Intl.getCanonicalLocales throws with custom regions


ChromeQ:
## To Reproduce

I am using custom regions in my Australian locales, for example some words are different depending on which state you are in. A "Pint" of beer in most of Australia is a "Pint", but in South Australia (SA) it is known as an "Imperial Pint" along with a [bunch of others](https://notoneoffbritishisms.com/2020/03/05/schooner/).

So I have set the locale as "en-AU-SA"

```js
const { t } = useTranslation('drinkTypes', { lng: 'en-AU-SA' });
```

It throws a RangeError:  `ERROR  [RangeError: com.facebook.hermes.intl.JSRangeErrorException: Invalid subtag: SA [at index 6]]` from within the native code for Intl.getCanonicalLocales

## Expected behavior

In v23 of i18next it used to work due to the `formatLanguageCode` - https://github.com/i18next/i18next/blob/e5170ad9d024c4279e8682a21f741491316ccd9e/src/LanguageUtils.js#L33

Even after [v23.16 introduced the Intl.getCanonicalLocales](https://github.com/i18next/i18next/commit/c3742325a9be3222e7fec75aafefbebb90cf3a0a) it worked because of the try catch and fallthrough to the previous behaviour - https://github.com/i18next/i18next/blob/1736965d1f21612c80e391b25574fe657de840a6/src/LanguageUtils.js#L36

## Fix

Although I understand my region is not "correct", it should allow me to get translations as they exist in my resource translation files.

From my tinkering I believe a simple try/catch around the formatLanguageCode and returning the code should suffice. I realise it is then up to me to ensure the code is formatted correctly.

```js
formatLanguageCode(code) {
    // http://www.iana.org/assignments/language-tags/language-tags.xhtml
    if (isString(code) && code.indexOf('-') > -1) {
      try {
        let formattedCode = Intl.getCanonicalLocales(code)[0];
        if (formattedCode && this.options.lowerCaseLng) {
          formattedCode = formattedCode.toLowerCase();
        }
        if (formattedCode) return formattedCode;

        if (this.options.lowerCaseLng) {
          return code.toLowerCase();
        }
      } catch {}

      return code;
    }

    return this.options.cleanCode || this.options.lowerCaseLng ? code.toLowerCase() : code;
  }
```

## Your Environment

- *runtime version*: react-native@0.74.5
- *i18next version*: 24.0.2

adrai:
addressed with v24.0.3

PR #2268, opened 2024-12-03
fix: incorrect locale detected

perrin4869:
Some language detectors may return invalid languages which cause `Intl.getCanonicalLocales` to throw. In this case, we need to swallow the error and ignore the detected language.

### DIFF
--- a/src/LanguageUtils.js
+++ b/src/LanguageUtils.js
@@ -31,7 +31,16 @@ class LanguageUtil {
   formatLanguageCode(code) {
     // http://www.iana.org/assignments/language-tags/language-tags.xhtml
     if (isString(code) && code.indexOf('-') > -1) {
-      let formattedCode = Intl.getCanonicalLocales(code)[0];
+      let formattedCode;
+      try {
+        formattedCode = Intl.getCanonicalLocales(code)[0];
+      } catch (e) {
+        // this can happen if the detected language is not valid
+        // for example: if the path detector is used and the requested path is `/providers`
+        if (!(e instanceof RangeError) || e.message !== 'Incorrect locale information provided') {
+          throw e;
+        }
+      }
       if (formattedCode && this.options.lowerCaseLng) {
         formattedCode = formattedCode.toLowerCase();
       }
--- a/test/runtime/languageUtils.test.js
+++ b/test/runtime/languageUtils.test.js
@@ -319,6 +319,7 @@ describe('LanguageUtils', () => {
       { args: [['c']], expected: 'en' },
       { args: [['e']], expected: 'en' },
       { args: [['user-id']], expected: 'en' },
+      { args: [['en-AU-SA']], expected: 'en' },
       { args: [[]], expected: 'en' },
     ];
 
--- a/test/runtime/languageUtils.test.js
+++ b/test/runtime/languageUtils.test.js
@@ -318,6 +318,7 @@ describe('LanguageUtils', () => {
       { args: [['ru']], expected: 'en' },
       { args: [['c']], expected: 'en' },
       { args: [['e']], expected: 'en' },
+      { args: [['user-id']], expected: 'en' },
       { args: [[]], expected: 'en' },
     ];
 


adrai:
Thank you  @perrin4869 ...
Can you also address https://github.com/i18next/i18next/issues/2267 since it touches the same function?

perrin4869:
I think this would close that issue too actually as-is

adrai:
Then can you add 'en-AU-SA' in the test?

adrai:
Wouldn't it be better to log instead of to throw and return the code?

adrai:
@ChromeQ what's your opinion?

perrin4869:
if the function throws it is not returning the code
you want to swallow a specific error when the code isn't valid for `Intl.getCanonicalLocales` but still throw in the case of other errors (which may not even exist at this point)

adrai:
you're right, it makes sense as is....

adrai:
included in v24.0.3

ChromeQ:
I just copied over the changes to my project. Unfortunately the logic is a bit incorrect and it is still throwing. I think it is meant to be:
```js
if (!(e instanceof RangeError || e.message !== 'Incorrect locale information provided')) {
```

The parenthesis around the `e instanceof RangeError` means it falls into the `||` and a RangeError will not have a message matching that string and so will throw

adrai:
I tested it locally, and it did not throw with en-AU-SA

adrai:
```js
const i18next = require('i18next')
i18next.init({
  fallbackLng: 'en',
  resources: {
    en: {
      translation: {
        k: 'hi there'
      }
    },
    'en-AU-SA': {
      translation: {
        k: 'hi there en-AU-SA'
      }
    }
  }
})

console.log(i18next.t('k'))
console.log(i18next.t('k', { lng: 'en-AU-SA' }))
```

ChromeQ:
With additional logging I am certainly seeing it fall into the throw, I'm testing on a react-native project on a real android device, perhaps the error message is different on other OS's?
image

adrai:
so, maybe better ignoring that error completely, and maybe just log it...?

ChromeQ:
My suggestion above should fix it...
```js
// for my case
if (!(e instanceof RangeError || e.message !== 'Incorrect locale information provided')) {
//    ^-------- true -------^    ^---------------------  not run --------------------^
//  ^----------------- flipped to `false` and skips conditional ------------------------^

// for other case
if (!(e instanceof RangeError || e.message !== 'Incorrect locale information provided')) {
//    ^-------- false ------^    ^----------------------- true ----------------------^
//  ^----------------- flipped to `false` and skips conditional ------------------------^
```

adrai:
This would also work, right?

```diff
formatLanguageCode(code) {
    // http://www.iana.org/assignments/language-tags/language-tags.xhtml
    if (isString(code) && code.indexOf('-') > -1) {
      let formattedCode;
      try {
        formattedCode = Intl.getCanonicalLocales(code)[0];
      } catch (e) {
-        // this can happen if the detected language is not valid
-        // for example: if the path detector is used and the requested path is `/providers`
-        if (!(e instanceof RangeError) || e.message !== 'Incorrect locale information provided') {
-          throw e;
-        }
+       this.logger.warn(`failed to format code: ${code}`);
+       /* fall through */     
      }
      if (formattedCode && this.options.lowerCaseLng) {
        formattedCode = formattedCode.toLowerCase();
      }
      if (formattedCode) return formattedCode;

      if (this.options.lowerCaseLng) {
        return code.toLowerCase();
      }

      return code;
    }

    return this.options.cleanCode || this.options.lowerCaseLng ? code.toLowerCase() : code;
  }
```

This would simplify the code and not throw in general...

ChromeQ:
Yes it works, it just depends if you want to ignore other unexpected or unaccounted errors as @perrin4869 mentioned

> you want to swallow a specific error when the code isn't valid for `Intl.getCanonicalLocales` but still throw in the case of other errors (which may not even exist at this point)

adrai:
I think that would be ok...
For these use cases it doesn't matter:

```js
const i18next = require('i18next')
i18next.init({
  fallbackLng: 'en',
  resources: {
    en: {
      translation: {
        k: 'hi there'
      }
    },
    'en-AU-SA': {
      translation: {
        k: 'hi there en-AU-SA'
      }
    }
  }
})

console.log(i18next.t('k')) // hi there
console.log(i18next.t('k', { lng: 'en-AU-SA' })) // hi there en-AU-SA
console.log(i18next.t('k', { lng: 'user-id' })) // hi there

console.log(i18next.resolvedLanguage) // en
i18next.changeLanguage('en-AU-SA')
console.log(i18next.resolvedLanguage, i18next.language, i18next.languages) // en-AU-SA en-AU-SA [ 'en-AU-SA', 'en-AU', 'en' ]

i18next.changeLanguage('user-id')
console.log(i18next.resolvedLanguage, i18next.language, i18next.languages) // en user-id [ 'user-id', 'user', 'en' ]
```

adrai:
please try with v24.0.4 in your environment

ChromeQ:
Tested with latest and can confirm no errors thrown. Thanks for the quick fix 🚀

perrin4869:
hm... I don't think we want to `warn` there, it would flood the console with irrelevant warnings. You want then to swallow `RangeErrors` and throw all else, and ignore the error message

perrin4869:
If we want to log, I think the correct loglevel would be `logger.debug`?

adrai:
It depends on what we want to achieve...
if you don't like to have "invalid" formatted codes, then logger.error would be better...
if you accept non-standard codes, then logger.debug or no log message at all is better...

What do we want to do?

perrin4869:
Many of the detectors in the ecosystem return incorrect codes and that's the way they are expected to work, so it's not an error. I'd say that in an express server setup to output warnings, this message would be triggered constantly, so debug makes more sense to me

adrai:
debug or no log at all?

perrin4869:
I think debug makes sense

adrai:
I just remembered... there is no debug level ;-)
there is just log, warn or error

adrai:
so keep as is, or remove the log statement?

perrin4869:
Or add a debug level lol
I think add a debug level or remove the warning.
It could also be called `logger.debug?.()` to protect against loggers that don't have a debug loglevel

adrai:
logger.debug is already used for the debug flag... so I'll remove the warning.

adrai:
v24.0.5

perrin4869:
Thanks! I'll try later when I get back home 😄

PR #2274, opened 2024-12-18
feat(typescript): Add `strictKeyChecks` option to enforce checking key existence when `defaultValue` is used

### LINKED ISSUE #2043: Wrong or missing key is not highlighted when there is a fallback
## 🐛 Bug Report

Wrong or missing key is not highlighted when there is a fallback. 

## To Reproduce

image

## Expected behavior

In both cases I'm referencing a key that doesn't exist, but in the first case I have a fallback, and the wrong key is not highlighted. In the second case, where i don't have a fallback, I get it highlighted.

## Your Environment

- "i18next": "^23.5.1"
- "react-i18next": "^13.2.2"
- "i18next-resources-for-ts": "^1.3.3"

**i18next.d.ts**

```
import Resources from './resources';

declare module 'i18next' {
  interface CustomTypeOptions {
    defaultNS: 'myfile';
    resources: Resources;
  }
}
```
Resources generated following [this step](https://github.com/locize/react-i18next-example-app-ts/blob/751f704984c206076d08638442ae34b3507aa7ad/package.json#L35)

masnormen:
In my experience, I use `defaultValue` for quick prototyping, but sometimes I forgot to add the actual key to my translation JSON file and it goes unnoticed because it doesn't produce errors.

This PR introduces a new option `strictKeyChecks`, which is customizable via user-augmented `CustomTypeOptions` interface. Enabling this option will make it illegal to reference a key that doesn't exist, **even when** `defaultValues` are provided.

If accepted, this PR closes #2043

Examples:

```ts
// public/locales/en/common.json
{
  "button": {
    "save": "Save"
  }
}
```

### without strictKeyChecks
```ts
// component.ts
export default function I18nextTest() {
  const { t } = useTranslation('common');

  t('button.save'); // ✅
  t('button.save', 'Default'); // ✅
  t('button.save', { defaultValue: 'Default' }); // ✅

  t('wrong'); // ❌ ts error
  t('wrong', 'Default'); // ✅ <= still pass typecheck 
  t('wrong', { defaultValue: 'Default' }); // ✅ <= still pass typecheck 
}
```

### with strictKeyChecks = true

```ts
// i18next.d.ts
declare module 'i18next' {
  interface CustomTypeOptions {
    strictKeyChecks: true;
    ...
  }
}


// component.ts
export default function I18nextTest() {
  const { t } = useTranslation('common');

  t('button.save'); // ✅
  t('button.save', 'Default'); // ✅
  t('button.save', { defaultValue: 'Default' }); // ✅

  t('wrong'); // ❌ ts error
  t('wrong', 'Default'); // ❌ ts error
  t('wrong', { defaultValue: 'Default' }); // ❌ ts error
}
```

#### Checklist

- [x] tests are included


typescript/t.d.ts
```diff
@@ -287,10 +288,16 @@ export interface TFunction<Ns extends Namespace = DefaultNamespace, KPrefix = un
...
-      | [key: Key | Key[], options?: ActualOptions]
-      | [key: string | string[], options: TOpt & $Dictionary & { defaultValue: DefaultValue }]
-      | [key: string | string[], defaultValue: DefaultValue, options?: TOpt & $Dictionary]
+    ...args: _StrictKeyChecks extends true
+      ? [
+          key: Key | Key[],
+          optionsOrDefaultValue?: ActualOptions | string,
+          options?: TOpt & $Dictionary,
+        ]
+      :
```
masnormen, line 297: Maybe there are better ways, but this totally works. So we have to kind of "overload" the second argument. I have tried to do this like below (which is ideal i think), but the types are messed up.

```ts
...args: _StrictKeyChecks extends true
      ? | [key: Key | Key[], options?: ActualOptions]
        | [key: Key | Key[], defaultValue: string; options?: TOpt & $Dictionary]
```

adrai:
Nice...
Does this come with any performance implication?
As soon as @marcalexiei approves this, we merge it.

marcalexiei (changes requested):
You code is not working alongside `returnObjects` option:

image


To test it add the following snippet to `test/typescript/strict-key-checks/t.test.ts`

```ts
describe('t defaultValue with strictKeyChecks == true and `returnObjects`', () => {
  const t = (() => '') as unknown as TFunction<'alternate'>;

  it('should work alongside `returnObjects`', () => {
    expectTypeOf(t('foobar', { returnObjects: true })).toMatchTypeOf<{
      barfoo: 'barfoo';
      deep: {
        deeper: {
          deeeeeper: 'foobar';
        };
      };
    }>();
  });
});
```

It refers to `TestNamespaceCustomAlternate` namespace

https://github.com/i18next/i18next/blob/c51950ecd987c62adfb75834812948b81c9ec01b/test/typescript/test.namespace.samples.ts#L16-L26


This code should work on  `test/typescript/custom-types`. Try pasting it here:

https://github.com/i18next/i18next/blob/c51950ecd987c62adfb75834812948b81c9ec01b/test/typescript/custom-types/t.test.ts

masnormen:
fixed and added the test. could you please recheck @marcalexiei?

masnormen:
You're right. I think the current solution is too 'hacky' and therefore break other things... Let me think of a cleaner solution for a bit

marcalexiei:
> I think the current solution is too 'hacky'

Function overloads are quite a pain to deal with. 
I've already tried to improve type checking, 
but managing a variable number of parameters that depend on various type parameters, 
without introducing breaking changes, is tricky. 
I’ve struggled to improve context type checking and auto-suggestions, 
but so far, the overloads have always defeated me 😅.

---

> Let me think of a cleaner solution for a bit

Sure! Thanks for tackling it anyway!

typescript/t.d.ts
```diff
@@ -294,4 +317,9 @@ export interface TFunction<Ns extends Namespace = DefaultNamespace, KPrefix = un
   ): TFunctionReturnOptionalDetails<TFunctionProcessReturnValue<$NoInfer<Ret>, DefaultValue>, TOpt>;
 }
 
+export type TFunction<
+  Ns extends Namespace = DefaultNamespace,
+  KPrefix = undefined,
+> = _StrictKeyChecks extends true ? TFunctionStrict<Ns, KPrefix> : TFunctionNonStrict<Ns, KPrefix>;
+
```
masnormen, line 324: Still not sure what are the possible implications of converting this `interface` into a `type` (although it will eventually resolve into an interface)

masnormen:
Some weird effects of this (or maybe just my lack of knowledge 🫠) is we sometimes have to use `as unknown` when casting the type of TFunction for tests.

```ts
// _StrictKeyChecks = true
const t = (() => '') as unknown as TFunction<'alternate'>;

// _StrictKeyChecks = false | works without additional casting
const t = (() => '') as TFunction<'alternate'>;
```

However, when I flipped the type position, the condition which requires `as unknown` is flipped.

```diff
- _StrictKeyChecks extends true ? TFunctionStrict<Ns, KPrefix> : TFunctionNonStrict<Ns, KPrefix>;
+ _StrictKeyChecks extends true ? TFunctionNonStrict<Ns, KPrefix> : TFunctionStrict<Ns, KPrefix>;
```

```ts
// _StrictKeyChecks = true | now this one works without additional casting
const t = (() => '') as TFunction<'alternate'>;

// _StrictKeyChecks = false | but this needs it
const t = (() => '') as unknown as TFunction<'alternate'>;
```

marcalexiei:
@masnormen you requested my review but you are still pushing commits, request a review when you have finished.

masnormen:
> @masnormen you requested my review but you are still pushing commits, request a review when you have finished.

Sorry @marcalexiei, my bad! I'm trying to make sure there are no perf issues and got some problem with vitest. Should be done now.

---

By the way, I followed the tutorial on CONTRIBUTING.md to test multiple tsconfig within the same scenario. But it turns out that Vitest will do multiple writes into the same `tsconfig.vitest-temp.json` file when there are multiple `tsconfig.*.json`, resulting in a broken temp tsconfig JSON and therefore broken test run:

image

The temp file name also seems to be [hardcoded here](https://github.com/vitest-dev/vitest/blob/78b62ffe1dc078d3e48573b3517f19bbeb898726/packages/vitest/src/typecheck/parse.ts#L80) so can't really do anything atm.

So for now I just copied the `many-keys` folder and adjusted `many-keys-strict` to check against the `strictKeyChecks: true` option.

marcalexiei (commented):
> Sorry @marcalexiei, my bad!

No problem! I haven't started the PR review yet, but better to avoid push after requesting a review😉

----

> I'm trying to make sure there are no perf issues

Looks like you did it ✨!
Done a few benchmarks on my biggest package in a mono-repo
Here are the results of running `tsc --noEmit`:

| With current types | With your fix |
|--------|--------|
| 18.479s |18.610s|
| 19.948s|18.420s|
| 19.077s |18.698s|
| 18.479s |18.804s|

I experienced similar results in `test/typescript/many-keys` test scenarios!

---

> By the way, I followed the tutorial on CONTRIBUTING.md to test multiple tsconfig within the same scenario. But it turns out that Vitest will do multiple writes into the same tsconfig.vitest-temp.json file when there are multiple tsconfig.*.json, resulting in a broken temp tsconfig JSON and therefore broken test run:

~~I know, that why I used the singular sentence:~~

~~better to add a warning to avoid this kind of situation in the future. 
Thanks for bringing this up!~~

I'll fix thisissue in a separate PR

---

> So for now I just copied the many-keys folder and adjusted many-keys-strict to check against the strictKeyChecks: true option.

I'm going to create a PR on your fork to avoid namespace duplication ~~and add a warning regarding testing vite with multiple `--project` flags. After that we can merge this~~.

marcalexiei (commented):
I created an issue on vitest repository about the `tsconfig.vitest-temp.json`: 
https://github.com/vitest-dev/vitest/issues/7107

Anyway if the CI will be green this can be merged

adrai:
Thanks to both...
it's included in v24.2.0

typescript/t.d.ts
adrai: That's the implication ;-) https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/i18next-sprintf-postprocessor/index.d.ts#L5
https://github.com/i18next/i18next/issues/2278

PR #2275, opened 2024-12-20
Standardize null/undefined checks across codebase

sebastiancarlos:
This PR standardizes the pattern of checking for null/undefined values using `== null` across the codebase. This is a well-known JavaScrip idiom, which was already being used at the codebase on the file `src/utils.js`.

This simplifies several conditionals while maintaining the same functionality. It particularly simplifies the check for `defaultNS`, in which extra care was taken to prevent the case of the variable being `false`.

#### Example:
```js
// Before
if (key === undefined || key === null)

// After
if (key == null)
```

#### Checklist

- [x] tests are included


adrai:
With this PR this will crash:

```js
console.log(i18next.t())
console.log(i18next.exists())
console.log(i18next.t(null))
console.log(i18next.exists(null))
```

If you update your branch, you'll see some failed tests

sebastiancarlos:
Hi @adrai, thank you for reviewing this PR and for your work on this project!

I tried to reproduce the issue, but I’m seeing consistent behavior on both master and my branch. Here’s what I tried:

1. Added a `test.js` file with the following content:
```javascript
const i18next = require('./index.js')

console.log(i18next.t())
console.log(i18next.exists())
console.log(i18next.t(null))
console.log(i18next.exists(null))
```

2. Ran it using the Node.js version specified in `.nvmrc` (20.18.0). For both `master` and my branch, the output was the same:
```shell
$ node test.js
undefined
undefined
undefined
undefined
```

Additionally:
- My branch is up to date. The commit history aligns with `master`, and GitHub confirms “This branch has no conflicts with the base branch.”
- All tests pass on both my branch and `master`:
```shell
$ npm run test
...
 Test Files  51 passed (51)
      Tests  894 passed | 9 skipped (903)
   Start at  12:41:00
   Duration  1.73s (transform 482ms, setup 0ms, collect 2.36s, tests 1.56s, environment 5ms, prepare 2.40s)
```

Could you share more details about the failing tests or the exact setup where you found these issues?

Thanks again for your time!

adrai:
sorry, I'm an idiot... I tried with === instead of ==

adrai:
No need to create a new release with this, right? Or do you have any issue because of this?

PR #2296, opened 2025-03-26
only log loaded translations in debug mode

mikebolt:
#### Checklist

- [ x ] only relevant code is changed (make a diff before you submit the PR)
- [ x ] run tests `npm run test`
  - 3 failing unit tests are related to date formatting
- [ ] tests are included
  - This is a very small change.
- [ x ] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

There's a lot of console output and there's no way to change the log level to only show warnings or errors. This PR would cause i18next to only log all translations when the debug config option is set to true.


mikebolt:
Sorry. I missed that when I was looking at logger.js. The issue is somewhere else.

Issue #1605, opened 2021-04-24
Multiple changeLanguage call may result in wrong order


waivital:
## 🐛 Bug Report

When call `changeLanguage` multiple times,  if `loadResources` inside the `changeLanguage` resolved in different order, then the final language will be the last resolved `changeLanguage` call, instead of the last `changeLanguage` call.
 
## To Reproduce

```js
    describe('changeLanguage', () => {
      it('it should keep changeLanguage call in order', () => {
        const instance = i18next.createInstance();
        const makeLang = (lan, delay) => ({
          delay,
          value: {
            foo: `${lan}`,
          },
        });
        const langMap = {
          en: makeLang('en', 100),
          zh: makeLang('zh', 500),
          es: makeLang('es', 300),
          jp: makeLang('jp', 200),
        };

        return new Promise(resolve => {
          instance
            .use({
              type: 'backend',
              read: function(language, namespace, callback) {
                const lang = langMap[language];

                setTimeout(() => {
                  callback(null, lang.value);
                }, lang.delay);
              },
            })
            .init(
              {
                fallbackLng: 'en',
              },
              (err, t) => {
                resolve(
                  Promise.all([
                    instance.changeLanguage('en'),
                    instance.changeLanguage('zh'),
                    instance.changeLanguage('jp'),
                    instance.changeLanguage('es'),
                  ]).then(() => {
                    expect(t('foo')).to.equal('es');
                  }),
                );
              },
            );
        });
      });
    });
```

## Expected behavior

Test not fail. The value of `t('foo')` is `es`

## Your Environment

- *runtime version*:  node v14.16.0
- *i18next version*: 20.2.1
- *os*:  Linux

jamuhl:
do you have a real use case you run into this? can‘t think of a situation where users toggle language like this

waivital:
I recently faced this issues when combined i18next and react-i18next with Server Side Render. 

The server side resolved language is 'zh' and the browser detector answer is 'en', because i18next init called before react hydrate, therefore changeLanguage first called with 'en' then 'zh', but the 'en' resource load a bit slower then the other, so the page result in wrong language.

When there are multiple place can cause language change (or change language twice shortly in one place), may need add some code keep it done in order.

I'm not quite sure about if its better do this work inside i18next or outside.

jamuhl:
[...]

PR #2298, opened 2025-04-10
Fix multiple changeLanguage calls may resolve to a wrong language

waivital:
Fix https://github.com/i18next/i18next/issues/1605

#### Checklist

- [x] tests are included

### DIFF
--- a/src/i18next.js
+++ b/src/i18next.js
@@ -345,11 +345,13 @@ class I18n extends EventEmitter {
 
     const done = (err, l) => {
       if (l) {
-        setLngProps(l);
-        this.translator.changeLanguage(l);
-        this.isLanguageChangingTo = undefined;
-        this.emit('languageChanged', l);
-        this.logger.log('languageChanged', l);
+        if (this.isLanguageChangingTo === lng) {
+          setLngProps(l);
+          this.translator.changeLanguage(l);
+          this.isLanguageChangingTo = undefined;
+          this.emit('languageChanged', l);
+          this.logger.log('languageChanged', l);
+        }
       } else {
         this.isLanguageChangingTo = undefined;
       }
--- a/test/runtime/i18next.test.js
+++ b/test/runtime/i18next.test.js
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeAll } from 'vitest';
+import { describe, it, expect, beforeAll, vi, beforeEach, afterEach } from 'vitest';
 import i18next from '../../src/i18next.js';
 import { get as getDefaults } from '../../src/defaults.js';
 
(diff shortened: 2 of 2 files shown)


adrai:
this PR will increase the package by approx. 3kb

adrai:
maybe this is already sufficient?
image

waivital:
@adrai  Oh, you are right, this is indeed much simpler, and the test case still passed.

adrai:
I think the new option is not necessary. Please remove it.
We just do it always.

waivital:
But it may introducing breaking changes. If someone relies on the old behavior, their code might stop working after the upgrade.

adrai:
The old behaviour is not reliable by itself... so this does not matter.

waivital:
I still feel that it is a bit too hasty to change it directly, after all, this problem has existed for so long. Lots of project already adopted with i18next's old behaviour.

adrai:
The old wrong behaviour occurs only if the changeLanguage is used in a wrong way. So this is a fix after all... But if you think this is breaking we will release a new major release... no problem. But I don't like to introduce a new option because of a "bug" or "wrong usage" behaviour.

adrai:
> But it may introducing breaking changes. If someone relies on the old behavior, their code might stop working after the upgrade.
> 
> And another concern is here
> 
> https://github.com/i18next/i18next/blob/94e4e7bb771597d19586f480216291975ba2467d/src/i18next.js#L348-L354
> 
> The `languageChanged` event no longer represents the real language changed event. To fix that, maybe here need to reject the outdated defer and pass an error to the callback.

feel free to optimize this...

waivital:
Got it, I'll look into this lately.

waivital:
I was thinking if we could optimize the changeLanguage function. It seems to involve three things: detect the language using a detector if language not provided, normalize the detected language, and load the language resources then set the related state. Maybe we can separate these steps to make the process clearer.

I created a new branch for this idea, if it looks good, I'll create a new PR for it.
https://github.com/waivital/i18next/commits/optimize-changelanguge-function/

jamuhl:
> I was thinking if we could optimize the changeLanguage function. It seems to involve three things: detect the language using a detector if language not provided, normalize the detected language, and load the language resources then set the related state. Maybe we can separate these steps to make the process clearer.
> 
> I created a new branch for this idea, if it looks good, I'll create a new PR for it. https://github.com/waivital/i18next/commits/optimize-changelanguge-function/

Not sure if I like the idea - changeLanguage does no detection with that change -> leads to a breaking version for no real benefit. Also adding `detectLanguage` and `normalizeLng` to public API seems to be unnecessary - at least for me it just adds more clutter.

waivital:
> > I was thinking if we could optimize the changeLanguage function. It seems to involve three things: detect the language using a detector if language not provided, normalize the detected language, and load the language resources then set the related state. Maybe we can separate these steps to make the process clearer.
> > I created a new branch for this idea, if it looks good, I'll create a new PR for it. https://github.com/waivital/i18next/commits/optimize-changelanguge-function/
> 
> Not sure if I like the idea - changeLanguage does no detection with that change -> leads to a breaking version for no real benefit. Also adding `detectLanguage` and `normalizeLng` to public API seems to be unnecessary - at least for me it just adds more clutter.

It's ok if `detectLanguage` and `normalizeLng` are public API or private API for me, it just a quick demo for the idea. I think the benefit of this change is that `changeLanguage` now does just one thing—find the matched language and load the related resources. If the language isn't found, it informs the user that there’s an issue with the configuration.

adrai:
I also don't like that change... For example, there are scenarios out there, where there is a lng passed via options but also a language detector...
And also that new thrown error there might be an issue "in the wild"...
So without these things, it's just splitting the content of the changeLanguage function into multiple functions... (no real benefit in userland)
So I would say we merge this PR only to fix the order of changeLanguage calls...

waivital:
Sure, we can discuss more on that later. :smile:

Issue #2318, opened 2025-06-10
Incomplete type definition for getFixedT() return value


emersion:
## 🐛 Bug Report

`getFixedT()` returns a special `TFunction` which supports an additional `keyPrefix` option:

https://github.com/i18next/i18next/blob/defa00f1a3bc79a32645d95f737b5ce4dcace14b/i18next.js#L1988

However the `TFunction` type definition doesn't reflect this and is missing the `keyPrefix` option.

## To Reproduce

```ts
const t = i18next.getFixedT('foo', 'bar');
t('hello', { keyPrefix: 'baz' });
```

This breaks when strict key checks are enabled, because `keyPrefix` is not taken into account. This is also confusing for documentation purposes: one needs to look into the source code to find out that this option exists.

## Expected behavior

The returned `TFunction` should have a `keyPrefix` field in its options type definition.

## Your Environment

- *runtime version*: node v24
- *i18next version*: 25.2.1
- *os*: Linux

adrai:
That's not the original T function => i18next.t()... this is the result of the getFixedT function => https://github.com/i18next/i18next/blob/defa00f1a3bc79a32645d95f737b5ce4dcace14b/src/i18next.js#L404

Feel free to provide a PR if this is not ok for you in TS...

emersion:
Yes, it is the return value of `getFixedT()`. Its type is incorrect, which is what this issue is about.

emersion:
I don't have time right now to work on a PR, would maybe have some in the future, could we re-open this issue to track the problem?

adrai:
v25.8.6 should be better

PR #2007, opened 2023-07-28
Typescript - Support `const` modifier for `options` and `key`

pedrodurek:
Closes https://github.com/i18next/i18next/issues/1929, https://github.com/i18next/i18next/issues/1950

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [x] tests are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

#### Checklist (for documentation change)

- [x] only relevant documentation part is changed (make a diff before you submit the PR)
- [x] motivation/reason is provided
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

PR #2319, opened 2025-06-11
feat: new autocompletion overload

### LINKED ISSUE #2317: Path computation is eager
## 🚀 Feature Proposal

Hello i18next team!

First let me say that I'm really impressed with the amount of engineering that has gone into the type-level algorithm for pre-computing all the translation paths.

I was wondering if you'd be open to a PR that simplifies the algorithm by taking a different approach?

The basic idea is to make path computation lazy (on-demand).

## Motivation

Currently the algorithm is eager, which means in order to get autocompletion, i18next needs to compute all possible paths ahead of time.

What if we could make this feature work more like autocomplete does natively?

## Example

Given this translation dictionary:

```
{
    greeting: {
        one: 'hello 1',
        two: 'hello 2',
    }
}
```

I'm proposing that autocomplete work like this:

Image

ahrjarrett:
Would close #2317

This is a draft PR that I'm opening to get feedback on a potential solution to the problem outlined in #2317.

### Problem

Currently when users use i18next with TypeScript, the compiler needs to do a significant amount of work upfront in order to type check even a single call to the `t` function.

To measure this cost, I used a tool called [@arktype/attest](https://github.com/arktypeio/arktype/tree/main/ark/attest) that tracks the number of allocations the TypeScript compiler needed to perform in order to type check a given expression.

Importantly, these instantiations continue to grow over time as users inevitably add more translations; in one example (included in this PR), a translation set of ~3,000 keys created [over 800,000](https://github.com/i18next/i18next/compare/master...ahrjarrett:ts-perf-optimization?expand=1#diff-2a4db2768405fdc5185bbf5cb95434e88af704a61c22f9873e05ccc403c77a85R7-R9) instantiations.

Not only does a larger set of instantiations cause IDE performance to drag, but since these instantiations are stored on the heap, this particular performance characteristic tends to interact poorly when used in conjunction with other tools such as `@typescript-eslint`, and can even cause OOM crashes in CI/CD pipelines (this issue is what originally motivated me to work on this problem).

Bugs like these are particularly insidious since their reproduction depends heavily on environmental factors. This makes them difficult to debug, and nearly impossible to reproduce.

### Proposed solution

This PR attempts to solve this problem by making path computation lazy.

It does so by using a bit of [TypeScript magic](https://github.com/i18next/i18next/compare/master...ahrjarrett:ts-perf-optimization?expand=1#diff-40c32b2c6c27c0460c561b514c69b66ac89624757c6ca1057da4c11d050fbb3eR289-R299) to mimic the native autocompletion API.

Usage looks like this:

i18next-example

#### Benefits 

1. Significantly fewer instantiations on the compiler's part.

For example, the same 3,000 keys mentioned above created only [771 instantiations](https://github.com/i18next/i18next/compare/master...ahrjarrett:ts-perf-optimization?expand=1#diff-343a8626d471dd7acbc585b56a15f3699ffe81d652423a99380451295e87c7f3R7-R9), which is a **1,000x reduction**

2. Translations are more discoverable

As translation sets continue to grow, "full-path" autocompletion becomes less ergonomic.

To demonstrate this, here's an example of ~1,800 translations with the current API:

i18next-eager

Arguably, translations become more discoverable with the proposed API:

i18next-lazy

#### Limitations

This approach is not without its limitations.

1. There's no easy way to integrate this solution with certain parts of the i18next API, namely _context_ and _interpolation_.

The reason for this is complicated, but tl;dr: TypeScript sometimes get confused when the 1st argument (the key) _depends on_ the value of the 2nd argument (the options object).

I spent considerable time getting it to work with context, specifically, and while it did mostly work, in some cases autocompletion broke, and I needed to restart the TS language server to get it to work again.

When @adrai tagged me in #2318 and I realized that we might also need to support the `keyPrefix` option, I changed my strategy for this PR.

Therefore, in its current state, **this PR only covers the case where a user does _not_ pass an options object**.

As soon as a user passes options, autocompletion falls back to eager resolution.

Furthermore, since users can choose to "stop" at any point in the object, this feature is only available to users when they select `returnObjects` to `true` -- which means with this PR, _this feature is currently disabled by default_.

### TODO

There is still more to do if the maintainers would like to merge this PR. 

- [ ] Add additional type-level tests; while all of the TypeScript tests are currently passing, I have a feeling there are some edge cases that are not tested
- [ ] Add additional benchmarks
- [ ] Add documentation for the new API

## Closing thoughts

Like I said in the beginning, I've opened this PR as a draft to get feedback, so if you have any thoughts or questions, please let me know!

In the end, since this change does not solve the entire problem, I'm personally on the fence about whether this is the best path forward. On the one hand, it does seem to improve DX, and definitely leads to far fewer allocations, but I also wouldn't want to ship this prematurely, only to discover that I'd missed an edge case that would be impossible to support, or arguably worse: that this feature ends up being costly for me (or other i18next contributors) to maintain.

#### Checklist

- [x] tests are included


marcalexiei:
Hi! Thanks a lot for opening this PR — really appreciate your work on this.

Just a heads-up: the CI won’t run until the conflicts with the base branch are resolved.

At the moment, I don’t have the capacity to do a full review, especially since this looks like a complete rewrite of the typings and includes some missing features. 
From my perspective, it makes sense to aim for feature parity with the current type system before merging anything into the main branch.

I totally see the value in improving the DX, but we also need to make sure we don’t lose type accuracy or coverage in the process.

Your POC is looking promising though! It might be a good idea to reach out to the maintainers about possibly merging this into a separate branch. That way, it could be published under an experimental tag for testing and feedback without affecting stable releases.

Thanks again for the effort — looking forward to seeing how this evolves!

typescript/t.d.ts
```diff
@@ -126,15 +126,15 @@ type ParseKeysByFallbackNs<Keys extends $Dictionary> = _FallbackNamespace extend
...
-export type FilterKeysByContext<Keys, Context> = Context extends string
+type FilterKeysByContext<Keys, Context> = Context extends string
   ? Keys extends
       | `${infer Prefix}${_ContextSeparator}${Context}${_PluralSeparator}${PluralSuffix}`
       | `${infer Prefix}${_ContextSeparator}${Context}`
     ? Prefix
     : never
   : Keys;
 
-export type ParseKeys<
```
marcalexiei, line 137: Beware about removing these exports some plugins / users might be using it.
`ParseKeys` is extremely useful also in user land

ahrjarrett: Ah, this was my IDE I think.

This is a quirk of `*.d.ts` files -- they behave like ambient namespaces (other way around actually, ambient namespaces behave like declaration files) in that all types that they contain are implicitly exported.

I'm happy to revert these lines, but just know that all types in `t.d.ts` are available to be used in userland.

ahrjarrett:
> Hi! Thanks a lot for opening this PR — really appreciate your work on this.
> 
> Just a heads-up: the CI won’t run until the conflicts with the base branch are resolved.
> 
> At the moment, I don’t have the capacity to do a full review, especially since this looks like a complete rewrite of the typings and includes some missing features. From my perspective, it makes sense to aim for feature parity with the current type system before merging anything into the main branch.
> 
> I totally see the value in improving the DX, but we also need to make sure we don’t lose type accuracy or coverage in the process.
> 
> Your POC is looking promising though! It might be a good idea to reach out to the maintainers about possibly merging this into a separate branch. That way, it could be published under an experimental tag for testing and feedback without affecting stable releases.
> 
> Thanks again for the effort — looking forward to seeing how this evolves!

Thank you for the review @marcalexiei -- just to clarify, in its current state, this PR is an additive change only -- none of the existing types were changed.

This PR simple adds a new overload that only applies when the user does not pass an options object, since I don't think it's possible to get the autocomplete behavior in cases where the set of available keys depends on which options are passed.

Either way, I'd say don't worry about doing a full review for now -- I'm working on validating another approach that I think would give the maintainers a more interesting set of tradeoffs to choose between.

typescript/t.d.ts
marcalexiei: > I'm happy to revert these lines, but just know that all types in t.d.ts are available to be used in userland.

Yes I know... they were always exported to userland. 
At some point they were moved into separate files to support different version of typescript.
Check the following diff https://github.com/i18next/i18next/pull/2007/files#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8

ahrjarrett:
Closed in favor of #2322

PR #2325, opened 2025-06-30
feat: support formatting function with plural and interpolation

### LINKED ISSUE #2323: Use formatting functions with interpolation and plurals in the nesting "t" function
## 🚀 Feature Proposal

Use formatting functions with interpolation and plurals in the nesting "$t" function.

## Motivation

This is useful if you need to reference other keys in a translation and need to change both the pluralization and the formatting simultaneously.

## Example

```json
{
    "boy": "Boy",
    "boy_other": "Boys",
    "girl": "Girl",
    "girl_other": "Girls",
    "girlsAndBoys": "They have {{girls}} $t(girl, { \"count\": {{girls}} }, lowercase) and {{boys}} $t(boy, { \"count\": {{boys}} }, lowercase)",
}
```

Result:

```javascript
i18next.t("girlsAndBoys", { girls: 3, boys: 2 }); // -> "They have 3 girls and 2 boys"
```

Currently, both `$t(girl, { \"count\": {{girls}} }, lowercase)` and `$t($t(boy, { \"count\": {{boys}} }), lowercase)` will not work. The error message of the latter: `i18next::translator: missingKey en $t(boy $t(boy`

I can PR based on #1419.

otomad:
Close #2323 .

#### Checklist

- [x] tests are included

#### Feature

Use formatting functions with interpolation and plurals in the nesting "$t" function.

Example:

```javascript
$t(ns:key, { count: {{count}}, arg1: "{{arg1}}", arg2: "{{arg2}}" }, formatter1, formatter2)
```

Note: you should not put formatters before the nesting option object. Because this is not only illogical, but also increases the complexity of the code. For example, the following code will not work properly.

```javascript
$t(ns:key, formatter1, formatter2, { "count": {{count}}, "arg1": "{{arg1}}", "arg2": "{{arg2}}" })
$t(ns:key, formatter1, { "count": {{count}}, "arg1": "{{arg1}}", "arg2": "{{arg2}}" }, formatter2)
```

##### `doReduce`

https://github.com/i18next/i18next/blob/a7f7d067f56d06feea74d247130ce26b404cbd52/src/Interpolator.js#L281
https://github.com/i18next/i18next/blob/a7f7d067f56d06feea74d247130ce26b404cbd52/src/Interpolator.js#L301

Here is a variable `doReduce`, but I don't understand what it does, so I replaced it with `formatters.length`.

##### `nestingRegexp`

I found that the part that determines the `$t()` function currently uses a simple regular expression to match. Ignoring custom configuration, its default value is `/\$t\((.+?)\)/`. This means that if you include a right parenthesis inside the function, it will break the function from working properly. For example:

```javascript
$t(key, { "name": "foo(bar)baz" })
```

This will break the function. However, since this part is beyond the scope of the target issue, I will not fixed it yet. I'm just giving a heads up.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -280,15 +280,14 @@ class Interpolator {
        */
       const keyEndIndex = /{.*}/.test(match[1])
         ? match[1].lastIndexOf('}') + 1
-        : match[1].indexOf(',');
+        : match[1].indexOf(this.formatSeparator);
       if (keyEndIndex !== -1) {
-        const key = match[1].slice(0, keyEndIndex);
         formatters = match[1]
           .slice(keyEndIndex)
           .split(this.formatSeparator)
           .map((elem) => elem.trim())
           .filter(Boolean);
-        match[1] = key;
+        match[1] = match[1].slice(0, keyEndIndex);
       }
 
       value = fc(handleHasOptions.call(this, match[1].trim(), clonedOptions), clonedOptions);


adrai:
Currently reviewing your PR....

> I found that the part that determines the `$t()` function currently uses a simple regular expression to match. Ignoring custom configuration, its default value is `/\$\((.+?)\)/`. This means that if you include a right parenthesis inside the function, it will break the function from working properly. For example:
> 
> ```js
> $t(key, { "name": "foo(bar)baz" })
> ```
> 
> This will break the function. However, since this part is beyond the scope of the target issue, I will not fixed it yet. I'm just giving a heads up.

Can you create a new PR for this?

Issue #2341, opened 2025-08-24
enableSelector: true doesn't work


vchirikov:
Hi, I tried new `enableSelector` feature and found a couple of [bugs](https://github.com/ahrjarrett/i18next-selector/issues/109). 

1. InitOptions<T> doesn't have `enableSelector`, so I use workaround to add it there

```ts
  interface CustomPluginOptions {
    enableSelector: true;
  }
```

1. after this vscode/Typescript works well, autocomplete works as expected, but:
1. i18next doesn't work with proxy 

<img width="354" height="158" alt="Image" src="https://github.com/user-attachments/assets/8a81cdc5-e85a-4565-b4c4-fab65d5f7fc4" />

client side i18n initialized like this:

```ts
export const initI18next = (
  initLocaleContent: ResourceLanguage | null,
  initLocale: Locale,
  initNamespace: Namespace = defaultNamespace
) => {
  const instance = createInstance();
  void instance.use(resourcesToBackend(resourceLanguageLoader)).init({
    enableSelector: true,
    partialBundledLanguages: true,
    debug: false,
    ...(initLocaleContent
      ? {
          resources: { [initLocale]: { [initNamespace]: initLocaleContent } },
          lng: initLocale,
          preload: [initLocale],
        }
      : {
          lng: 'cimode',
        }),
    supportedLngs: locales,
    fallbackLng: defaultLocale,
    defaultNS: defaultNamespace,
    fallbackNS: defaultNamespace,
    ns: initNamespace,
    interpolation: {
      // react handles this
      escapeValue: false,
    },
    react: {
      useSuspense: true,
    },
    nsSeparator: ':::',
    keySeparator: '::',
    pluralSeparator: '##',
    contextSeparator: '##',
  });
  return instance;
};
```

As you can see on screenshot looks like proxy doesn't use `keySeparator`

## Your Environment

versions:
```
    "i18next": "25.4.1",
    "i18next-resources-to-backend": "1.2.1",
```

vchirikov:
It's interesting, that this works

<img width="566" height="160" alt="Image" src="https://github.com/user-attachments/assets/f1d213de-db6e-44da-bde8-85814d5fe409" />

I think that something wrong with fix for `keySeparator` in 25.4.1 @adrai / @ahrjarrett

vchirikov:
@adrai I've added tests for this and made a PR - #2342

adrai:
Can you try with v25.4.2 ?

adrai:
[...]

PR #2342, opened 2025-08-24
fix: fix selector api with custom separators (fix for #2341)

vchirikov:
`typeof options === 'object'` is useless, because it always undefined (code should check this.options instead `options`, also 1. `this.options` can't be non-object 2. even with `undefined` / `null` fix will work fine)

---
#### Checklist

- [x] tests are included

### DIFF
--- a/src/Translator.js
+++ b/src/Translator.js
@@ -90,8 +90,7 @@ class Translator extends EventEmitter {
       /* eslint prefer-rest-params: 0 */
       opt = this.options.overloadTranslationOptionHandler(arguments);
     }
-    if (typeof options === 'object') opt = { ...opt };
-    if (!opt) opt = {};
+    if (!opt) opt = { ...this.options };
 
     // non valid keys handling
     if (keys == null /* || keys === '' */) return '';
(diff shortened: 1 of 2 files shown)


src/Translator.js
```diff
@@ -90,8 +90,7 @@ class Translator extends EventEmitter {
       /* eslint prefer-rest-params: 0 */
       opt = this.options.overloadTranslationOptionHandler(arguments);
     }
-    if (typeof options === 'object') opt = { ...opt };
-    if (!opt) opt = {};
```
ahrjarrett, line 94: If the problem is that the keySeparator isn't being propagated to `keysFromSelector`, I think there's a less risky way to fix it.

This is the only line that changed in this file when we introduced the selector API:

```javascript
// line 98
if (typeof keys === 'function') keys = keysFromSelector(keys, opt);
```

@vchirikov can you test that this also fixes the issue?

```javascript
// line 98
if (typeof keys === 'function') keys = keysFromSelector(keys, { ...this.options, ...opt });
```

test/runtime/translator/translator.translate.selector.test.js
```diff
@@ -0,0 +1,56 @@
...
+    beforeAll(() => {
+      const rs = new ResourceStore({
+        en: {
+          root: {
+            foo: {
+              bar: 'foobar'
+            }
+          },
+        },
+      });
```
adrai, line 22: This test is wrong... sorry.
The keySeparator options etc... needs also to be passed to the resource store...
Will close this PR... a fix is coming...

PR #2361, opened 2025-10-25
fix: TFunctionReturn fallback

### LINKED ISSUE #2360: t() return type is never if using key from fallbackNS
## 🐛 Bug Report

If you setup an i18n instance with 2 ns(the default `translation` and `custom`) and with `fallbackNS: translation`
and then create a `t<"custom">` instance, calling `t("keyFromTranslationNS")` will return "never" instead of "string"

## To Reproduce

I modified one of the react-i18next examples: made a [repo](https://github.com/leobenzol/i18next-react-i18next-master-example-react-typescript_simple) for convinience. The last commit is all the changes i made over the example

## Expected behavior

Using t("fallbackKey") should return "string", not "never"

I tested it and i'm prettu sure it would be as simple as changing the line: https://github.com/i18next/i18next/blob/master/typescript/t.d.ts#L247
with: `$PreservedValue<ParseTReturn<ActualKey, Resources[$FirstNamespace<ActualNS>], TOpt>, ParseTReturn<ActualKey, Resources[_FallbackNamespace], TOpt>>`

## Your Environment

- *runtime version*: node 20
- *i18next version*: 25.6.0
- *os*: Mac

leobenzol:
Closes #2360 

I had some problems trying to support fallbackNS *arrays*, which made it a bit more complicated than the simple fix I proposed in the issue. I'm not a typescript expert so there might be a simpler way.

I think the tests I changed already cover the fix so I didn't add any extra.

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/custom-types/getFixedT.test.ts
+++ b/test/typescript/custom-types/getFixedT.test.ts
@@ -51,7 +51,7 @@ describe('getFixedT', () => {
     expectTypeOf(t('bar')).toEqualTypeOf<'bar'>();
 
     expectTypeOf(t('foobar.barfoo', { ns: 'alternate' })).toEqualTypeOf<'barfoo'>();
-    expectTypeOf(t('fallbackKey')).toEqualTypeOf<never>();
+    expectTypeOf(t('fallbackKey')).toBeString();
 
     // @ts-expect-error
     assertType(t('alternate:foobar.barfoo'));
--- a/test/typescript/fallback-types/getFixedT.test.ts
+++ b/test/typescript/fallback-types/getFixedT.test.ts
@@ -17,7 +17,7 @@ describe('t', () => {
   });
 
   it('should accept fallback keys', () => {
-    expectTypeOf(t('fallbackKey')).toEqualTypeOf<never>();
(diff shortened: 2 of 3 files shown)


leobenzol:
hey @ahrjarrett I looked into attest and got it working. I tried the following test before and after the change:
```ts
import { bench } from '@arktype/attest';
import { } from './i18next';
import type { TFunction } from 'i18next';

declare const t: TFunction;

bench('t() return type resolution', () => {
    t('bar');
}).types([1, 'instantiations']);

bench('fallback t() return type resolution', () => {
    t('fallbackKey');
}).types([1, 'instantiations']);

bench('fallback2 t() return type resolution', () => {
    t('anotherFallbackKey');
}).types([1, 'instantiations']);
```
You have 15600 instantiations before in all 3 cases and 17400 after, or about a 11.5% increase

This isnt the best test as it doesn't only measure the return type instantiation. If I add 500 synthetic keys to all 3 namespaces it goes to 556k instances before and 650k after, or a 17% increase.

In any case it is not even close to the 2x that we both imagined, so imo it's worth it, up to you.

adrai:
included (and adapted) in v25.8.0

PR #2384, opened 2026-02-03
fix(types): Selector API - fix Namespace inference for selector `ns` option

### LINKED ISSUE #2383: Selector API: TFunction with specified namespace doesn't restrict types to the provided namespaces
## 🐛 Bug Report

`TFunction` doesn't report a type error when using a namespace that isn't available.

This is a problem with `react-i18next` where `useTranslation` accepts one or more namespaces and the resulting `t` should only accept keys coming from those namespaces, rather than  from all namespaces.

This is especially problematic when loading namespaces asynchronously, 
because you might end up using keys from namespace that haven't been loaded yet.

## To Reproduce

https://github.com/i18next/i18next/blob/da8d29cc5421911055c730767bcfce6ee52a90ef/test/typescript/selector-custom-types/t.test.ts

```ts
  it('should error when using a defined namespace that is not present in TFunction NS type parameter', () => {
    const t: TFunction<['alternate']> = (() => '') as never;

    // @ts-expect-error
    t($ => $.foo, { ns: 'custom' }) // currently no error is reported
  })
```

## Expected behavior

```js
t($ => $.foo, { ns: ['custom'] })
t($ => $.foo, { ns: 'custom' })
```

should report a type error since the function only accepts `alternate` namespace

## Your Environment

- *runtime version*: not relevant 
- *i18next version*: 25.8.0
- *os*: not relevant 

marcalexiei:
- closes #2383

#### Checklist

- [x] tests are included

### DIFF
--- a/test/typescript/selector-custom-types/t.test.ts
+++ b/test/typescript/selector-custom-types/t.test.ts
@@ -355,4 +355,12 @@ describe('t', () => {
       t(($) => $.keypath_deep, { suffering: { in: { the: 'yes' } } });
     });
   });
+
+  // https://github.com/i18next/i18next/issues/2383
+  it('should error when using a defined namespace that is not present in TFunction NS type parameter', () => {
+    const t: TFunction<['alternate']> = (() => '') as never;
+
+    // @ts-expect-error
+    t(($) => $.foo, { ns: 'custom' });
+  });
 });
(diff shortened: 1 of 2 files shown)


adrai:
thx, included in v25.8.1

PR #2393, opened 2026-02-15
types(i18n): add missing toJSON() declaration

nagosu:
## What
Add missing `toJSON()` declaration to the `i18n` interface.

## Why
Runtime already exposes `I18n#toJSON()` but TypeScript types did not, so `i18next.toJSON()` was not type-safe / discoverable.

## How
- Add `toJSON()` signature to `index.d.ts`.
- Add a TypeScript regression test under `test/typescript/misc`.

## Tests
- npm run format
- npm run lint
- npm run test:typescript -- --project misc

## Risk
Type-only change; no runtime behavior changes. Return shape matches runtime `toJSON()` implementation.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -395,6 +395,17 @@ export interface i18n extends CustomInstanceExtensions {
   cloneInstance(options?: CloneOptions, callback?: Callback): i18n;
 
   /**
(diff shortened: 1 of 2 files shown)


adrai:
thx, included in v25.8.8

PR #2394, opened 2026-02-16
fix(interpolator): escape nestingOptionsSeparator in nesting option parsing

nagosu:
## What
Fix nesting option parsing by escaping `nestingOptionsSeparator` before constructing the internal `RegExp` in `Interpolator#nest()`.

## Why
`nestingOptionsSeparator` is user-configurable (default: `,`).  
When it contains regex meta characters (e.g. `+`, `*`, `|`, `(`), the current `RegExp` construction can throw or break the split logic.

## How
- Escape `sep` with `regexEscape(sep)` when building the `RegExp` in `handleHasOptions()`.
- Add a runtime regression test covering a separator with a regex meta character (e.g. `+`).

## Tests
- npm run format
- npm run lint
- npm run test:runtime

## Risk
Internal parsing fix only. Default separator behavior remains unchanged. No public API changes.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -228,7 +228,7 @@ class Interpolator {
       const sep = this.nestingOptionsSeparator;
       if (key.indexOf(sep) < 0) return key;
 
-      const c = key.split(new RegExp(`${sep}[ ]*{`));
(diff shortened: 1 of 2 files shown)


adrai:
thx, included in v25.8.9

PR #2430, opened 2026-05-07
fix(types): add type predicate overload to ExistsFunction

### LINKED ISSUE #2425: Type error assigning `.exists` to `ExistsFunction` since v25.10.4
## 🐛 Bug Report

We have some code that creates an `.exists` function with some defaults set (e.g. to set-up the namespace and language) in an object we type using i18next’s `ExistsFunction` type, e.g. something like this:

```ts
import i18next, { type ExistsFunction } from "i18next";

interface Utils {
	exists: ExistsFunction;
}

const obj: Utils = {
  exists = (key, options) => i18next.exists(key, { lng: 'en', ns: 'ourNamespace', ...options }),
}
```

When updating to 25.10.4 or higher, this no longer works as expected because `ExistsFunction` includes the type predicate return type and it seems TypeScript can no longer correctly infer that this usage is safe.

## To Reproduce

```ts
import i18next, { type ExistsFunction, type TOptions } from "i18next";

let foo: ExistsFunction;

// Doesn’t work (type predicate not inferred to match target signature)
foo = (key, options) => i18next.exists(key, options);
foo = (key, options) => i18next.exists(key, { lng: "lang", ...options });
```

[TypeScript playground link](https://www.typescriptlang.org/play/?ssl=12&ssc=74&pln=12&pc=1#code/JYWwDg9gTgLgBMAjADgHYFMAeMA0cDecMAnmOnAKKbADOMNAYgK6oDGMwEqeJZcAKgHkwHLjTgBfOADMoEEHABESNFhiKA3ACgtKjNgB0wVMBgAKAJTatAG3TxpECAC5K1OoxbtOqawHo-OAB1aABrGjwAIyZ4AEk4AHcAQ1R4GAg4JIATLLgaeXIs9Gkkpht4CBEfGi1HDIBeBBR9GAMsWnp-QIARLkBMAngEsLgzXnIwKHQs4FYkmHJUCHhjaXQoSdz0uBA51gALIiSoAHN7POBj1DmmSYtapzhGs1D0YjxK0VQaC0eAPibVIZ2h5nq93lUxFZ7g0Ri83nAPtUfvV-no1G13PRQfDCDZUMdXIobCljoo8AYKYixJIoVoAsEwuJostxClFsRjMcbMQ4AA3NaRCA0ciRHmsSoc-GHE5nUBgOwgdCpOY+aGPOAAHi0cAEAGlXuI1EqsuI6FBOeqzZycNqBLFUmtIMTPgBZJJgOBG1AmhGRABW6HY6sIAG04a4rfiALquFihRYJVCSG2-My28N6g02nVUr4AflcQghXw1-Ht8ygTpVXDdYF+NuRqOa6OBWLheFx+MJxPxZLgFIMufEEihQA)

## Expected behavior

I’d like to be able to keep using the simple code we have or have an alternative

## Your Environment

- *runtime version*: Node 24.13.0
- *i18next version*: >=25.10.4
- *os*: macOS

mixelburg:
## Problem

The `ExistsFunction` interface was missing the type predicate overload that `i18n.exists` provides. This caused TypeScript errors when assigning `i18n.exists` to a variable typed as `ExistsFunction`:

```ts
import i18next, { type ExistsFunction } from 'i18next';

let foo: ExistsFunction;
foo = (key, options) => i18next.exists(key, options); // TS error!
```

TypeScript cannot reconcile a type predicate return type (`key is K & SelectorKey`) with a plain `boolean` return type on the target signature.

## Fix

Added a type predicate first overload to `ExistsFunction` matching the pattern already used on `i18n.exists`:

```ts
export interface ExistsFunction<TKeys extends string = string, TInterpolationMap extends object = > {
  <K extends TKeys>(key: K, options?: TOptions<TInterpolationMap>): key is K & SelectorKey;
  (key: TKeys | TKeys[], options?: TOptions<TInterpolationMap>): boolean;
}
```

Fixes #2425

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -223,6 +223,7 @@ export interface ExistsFunction<
   TKeys extends string = string,
   TInterpolationMap extends object = $Dictionary,
 > {
+  <K extends TKeys>(key: K, options?: TOptions<TInterpolationMap>): key is K & SelectorKey;
   (key: TKeys | TKeys[], options?: TOptions<TInterpolationMap>): boolean;
 }
 


adrai:
Thanks for the PR — but I'm going to close this without merging.

The predicate overload was *deliberately* removed from `ExistsFunction` in v26.0.8 (commit e802567) to fix exactly the issue this PR is linked to (#2425). Adding it back re-introduces the bug, because TypeScript can't infer a type predicate through an arrow-function assignment to a multi-overload type — your example

```ts
let foo: ExistsFunction;
foo = (key, options) => i18next.exists(key, options);
```

would *fail to compile* with this change, not start working. That's the same TS quirk that broke wrappers in v25.10.4 and got reported as #2425.

The current shape is intentional and documented inline on `ExistsFunction`:
- `ExistsFunction` = wrapper-friendly shape, no predicate, always-`boolean` return.
- `i18n.exists` carries the predicate so direct callers still get narrowing.
- If you want narrowing in your own code, call `i18next.exists` directly or forward it by reference (`const exists: typeof i18next.exists = i18next.exists`). Arrow-function wrappers cannot preserve a type predicate — that's a TS limitation, not something this interface can fix.

I just tightened the JSDoc on `ExistsFunction` (8fec684) to make the wrapper limitation more explicit so this is less likely to trip future readers.

PR #2431, opened 2026-05-14
fix(types): expose enableSelector on InitOptions

Faithfinder:
## Problem

`enableSelector` is a runtime option — `i18next.js` reads `opts?.enableSelector === 'strict'` inside `keysFromSelector` and applies the namespace-prefix rewrite accordingly — but it isn't declared on `InitOptions`. As a result, `i18n.init({ enableSelector: 'strict' })` fails to typecheck:

```
Object literal may only specify known properties, and
'enableSelector' does not exist in type 'InitOptions<unknown>'.
```

`CustomTypeOptions['enableSelector']` exists (used for the type-level selector machinery in `t.d.ts`), but that only configures the *compile-time* shape. Users still have to pass `enableSelector` at runtime to actually get strict-mode resolution, and the matching `InitOptions` field is missing.

Today every project enabling the selector API needs a module augmentation:

```ts
declare module "i18next" {
  interface InitOptions {
    enableSelector?: false | true | "optimize" | "strict";
  }
}
```

## Fix

Adds the runtime-side option to `InitOptions`, slotted next to the other selector-resolution knobs (`keySeparator` / `nsSeparator`).

Includes a `test/typescript/misc/i18nInstance.test.ts` case covering each accepted value:

```ts
it('should accept `enableSelector`', () => {
  expectTypeOf(i18next.init).toBeCallableWith({ enableSelector: false });
  expectTypeOf(i18next.init).toBeCallableWith({ enableSelector: true });
  expectTypeOf(i18next.init).toBeCallableWith({ enableSelector: 'optimize' });
  expectTypeOf(i18next.init).toBeCallableWith({ enableSelector: 'strict' });
});
```

```
$ npx vitest --config vitest.workspace.typescript.mts run
 Test Files  56 passed | 1 skipped (56)
      Tests  460 passed | 2 todo (462)
Type Errors  no errors
```

### DIFF
--- a/test/typescript/misc/i18nInstance.test.ts
+++ b/test/typescript/misc/i18nInstance.test.ts
@@ -66,6 +66,21 @@ describe('i18nInstance', () => {
       });
     });
 
+    it('should accept `enableSelector`', () => {
+      expectTypeOf(i18next.init).toBeCallableWith({
+        enableSelector: false,
+      });
+      expectTypeOf(i18next.init).toBeCallableWith({
+        enableSelector: true,
+      });
+      expectTypeOf(i18next.init).toBeCallableWith({
+        enableSelector: 'optimize',
+      });
+      expectTypeOf(i18next.init).toBeCallableWith({
+        enableSelector: 'strict',
+      });
+    });
+
     it('should accept `backend` options', () => {
       expectTypeOf(i18next.init).toBeCallableWith({
         fallbackLng: 'en',
--- a/typescript/options.d.ts
+++ b/typescript/options.d.ts
@@ -642,6 +642,19 @@ export interface InitOptions<T = object> extends PluginOptions<T> {
   nsSeparator?: false | string;
 
   /**
+   * Selector-API mode. Mirrors `CustomTypeOptions['enableSelector']`.
+   *
+   * - `false` (default): selector API disabled.
+   * - `true` / `'optimize'`: selector resolution enabled.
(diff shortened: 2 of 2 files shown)


adrai:
Shipped in **i18next 26.2.0**. Thanks @Faithfinder!

jayaddison:
@adrai @Faithfinder do you know whether it'd be straightforward to expand this type coverage (I suppose for all `init` options) to the `jquery-i18next` interface?

adrai:
@jayaddison good call — turned out to be straightforward. Just shipped it in **jquery-i18next@1.3.0** — and hey, 9 years between releases, good to know jquery-i18next still has users! 😄 The package now ships its own TypeScript definitions covering all of jquery-i18next's `init` options (`tName`, `i18nName`, `handleName`, `selectorAttr`, `targetAttr`, `optionsAttr`, `useOptionsAttr`, `parseDefaultValueFromContent`), with the i18next instance typed as `i18n`. When the default handle names are kept, `$.t`, `$.i18n` and `$(selector).localize(opts)` are added to the jQuery types too. No more manual augmentation needed. 🙂

jayaddison:
That's super impressive - thank you @adrai!  I'll update `jquery-i18next` for my application soon :)

jayaddison:
@adrai that upgrade of `jquery-i18next` helped me figure out that in fact, I'd placed the `enableSelector` option in the wrong init options.  Typechecking is good!  Thanks both :)

PR #2437, opened 2026-06-23
fix: reassemble comma-in-parens formatters at any chain position

spokodev:
## Problem

Chained formatters support parenthesised options that may contain the format separator, for example `join(separator: ', ')`. When such a formatter is split on the separator, the option fragments have to be rejoined before parsing.

Today that reassembly only runs for `formats[0]`. A formatter whose parenthesised options contain the separator therefore works **only when it is first** in the chain:

```js
// works
{{v, join(separator: ', '), uppercase}}

// broken: corrupt output
{{v, uppercase, join(separator: ', ')}}
```

In the broken case the `join(separator: ', ')` fragment is split into `join(separator: ' '` and `')'`, never rejoined, and the value comes out corrupted (e.g. `ANaNB`).

## Fix

Replace the `formats[0]`-only repair with a single position-independent pass: walk the split fragments and, whenever a fragment opens a paren that is not yet closed, keep appending following fragments (re-inserting the separator) until the paren closes. This reassembles comma-in-parens options at **any** position in the chain. The downstream `formats.reduce(...)` is unchanged.

## Tests

Added a regression test in `test/runtime/i18next.translation.formatting.test.js` exercising a chain where the comma-in-parens formatter is not first (`{{myVar, upper, join(separator: ', ')}}` with `['a','b']` -> `A, B`).

- Before the fix, the new test fails (`Hello ANaNB`).
- After the fix it passes, and the full runtime suite stays green.

### DIFF
--- a/src/Formatter.js
+++ b/src/Formatter.js
@@ -117,15 +117,14 @@ class Formatter {
   format(value, format, lng, options = {}) {
     if (!format) return value;
     if (value == null) return value;
-    const formats = format.split(this.formatSeparator);
-    if (
-      formats.length > 1 &&
-      formats[0].indexOf('(') > 1 &&
-      !formats[0].includes(')') &&
-      formats.find((f) => f.includes(')'))
-    ) {
-      const lastIndex = formats.findIndex((f) => f.includes(')'));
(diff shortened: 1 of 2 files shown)


adrai:
thx.... merged and included in v26.3.2

PR #2438, opened 2026-06-26
fix(types): preserve selector returnObjects shape with context (#2398)

### LINKED ISSUE #2398: Incorrect type narrowing when combining `returnObjects: true` and `context` with selector syntax
## 🐛 Bug Report

When using the `t` function (enableSelector: true) with both `returnObjects: true` and a `context` option, the inferred return type is incorrect.

**Problem 1: Array object types are incorrectly narrowed**

Given a translation key that resolves to an array of objects, the returned type loses properties. Instead of inferring the full object shape, it produces a union of partial types. (`src/App.tsx#16`)

**Problem 2: Valid translation keys with context produce a TS error (maybe not a but and intended behavior but wanted to report anyway)**
Using the context option with a valid translation key causes a TypeScript error, even though the key and context combination is valid at runtime. (`src/App.tsx#29`)

## To Reproduce

https://codesandbox.io/p/devbox/w5qphh (wanted to do a sandbox but TS has a issue there so its a codebox)

## Expected behavior

**Problem 1.** I expect that the return type narrows to 
```
({
    transKey1: string;
    transKey2: string;
})[]
```
**Problem 2.** No TS error

## Your Environment

- *runtime version*: node 22 and node 20
- *i18next version*: 25.8.13
- *os*: Mac 15.7.4
- *TypeScript version*: 5.9.3 and 5.8.3

sauravgupta-dotcom:
## Summary

Fixes the remaining TypeScript selector regression from [#2398](https://github.com/i18next/i18next/issues/2398): `t(selector, { returnObjects: true, context })` on heterogeneous JSON arrays was inferring a union of partial element types (or `unknown`) instead of preserving full object shapes such as `{ transKey1: string; transKey2: string }[]`.

Problem 2 from the issue (valid context keys erroring) was already addressed on master; this PR targets the leftover Problem 1 case and the context + plurals array regression test.

## Changes

- **`FilterKeys`**: distribute over object unions so each JSON array element is filtered independently (avoids `keyof (A | B)` only seeing common keys).
- **`_DefinedContextKeys` / `_HasContextVariant` / `_IsContextualKey`**: ignore phantom optional `undefined` context keys that TypeScript adds when merging heterogeneous array element types (e.g. `transKey1_withContext?: undefined` on elements that do not define that key).
- **`TFunctionSelector`**: add a more specific overload for `context` + `returnObjects: true` using `const Context` + `const Fn` + `ReturnType<Fn>`, matching the non-context selector pattern so `Target` is not collapsed via `ConstrainTarget` / `ApplyTarget` → `unknown`.
- **Tests**: promote `test/typescript/selector-issue-2398/` from `.todo` to real `expectTypeOf` assertions (including context + plurals on array elements).

## Test plan

- [x] Local commit includes regression tests in `test/typescript/selector-issue-2398/t.test.ts`
- [ ] `npx vitest --config vitest.workspace.typescript.mts run test/typescript/selector-issue-2398`
- [ ] `npx vitest --config vitest.workspace.typescript.mts run` (full TypeScript workspace)
- [ ] Spot-check existing selector context cases in `test/typescript/selector/` still pass

Fixes #2398

### DIFF
--- a/test/typescript/selector-issue-2398/t.test.ts
+++ b/test/typescript/selector-issue-2398/t.test.ts
@@ -1,49 +1,36 @@
-import { describe, it } from 'vitest';
+import { describe, it, expectTypeOf } from 'vitest';
 import type { TFunction } from 'i18next';
 
-// Reproduces https://github.com/i18next/i18next/issues/2398 (Problem 1)
+// Reproduces https://github.com/i18next/i18next/issues/2398
 //
 // When a JSON array has objects with different key sets (one with a context
 // variant like `transKey1_withContext`, one without), TypeScript merges all
-// keys into a single element type. The selector overload with `context` then
-// loses `transKey1` from the inferred return type, producing:
-//   ({ transKey2: string } | { transKey1: string; transKey2: string })[]
-// instead of:
-//   { transKey1: string; transKey2: string }[]
-//
-// Root cause: FilterKeys works correctly when tested directly. The issue is
-// in the context-specific selector overload (TFunctionSelector) where
-// `Target` is inferred via `SelectorFn<Source, ApplyTarget<Target, Opts>, Opts>`.
-// With `returnObjects: true`, `ApplyTarget` resolves to `unknown`, so TypeScript
-// cannot infer a specific `Target` from the callback. The non-context overloads
-// solve this with `const Fn` + `ReturnType<Fn>`, but that approach breaks
-// namespace-ordering checks when applied to the context overload (because `Fn`'s
-// parameter depends on `Opts['context']`, creating an inference dependency).
+// keys into a single element type. The selector overload with `context` must
+// still preserve `transKey1` and `transKey2` on every element after filtering.
 
 describe('issue #2398 - returnObjects + context + array with selector', () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const t = (() => []) as unknown as TFunction;
 
-  it.todo('should return full object shape for all array elements, not a union of partial types');
-  // When fixed, replace the .todo above with:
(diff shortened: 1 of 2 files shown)


sauravgupta-dotcom:
Review request (API could not assign reviewers from this fork account — please take a look if you have a moment):

- @adrai (maintainer; tracked #2398)
- @ahrjarrett (selector API author)

Labels intended for this PR (matching the issue): `typescript`, `help wanted`.

adrai:
Merged and released in [`v26.3.3`](https://www.npmjs.com/package/i18next/v/26.3.3) 🎉 Thanks @sauravgupta-dotcom!

Before merging I verified the change end-to-end:

- **Targeted tests**: the promoted `selector-issue-2398` assertions pass (both property presence and `string` value types, including the context + plurals case).
- **Full TS workspace**: all 499 type tests across 63 files pass with no type errors, so no regression in the existing selector / plural / context / namespace cases.
- **No behavioral drift anywhere**: diffed the exact `tsc` error set across the whole `test/typescript` corpus before vs. after. The intentional negative tests (`@ts-expect-error` etc.) act as a large assertion surface, and the set is identical (valid code still checks, invalid code still errors).
- **Type-checker cost**: measured with `tsc --extendedDiagnostics`, +3.7% instantiations / +3.4% types from the extra `_DefinedContextKeys` mapped type and the `FilterKeys` union distribution, with no measurable wall-clock change. Runtime is unaffected (`.d.ts`-only, zero JS emitted).

The `FilterKeys` union-distribution fix is a nice general correctness improvement beyond just this case. Appreciate the thorough writeup and regression tests 🙏

PR #2440, opened 2026-07-08
fix: parse $t() nesting options block that spans multiple lines

spokodev:
`nest()` decides where the nested key ends and the trailing formatters begin by checking `match[1]` for an options object with `/{.*}/`. That regex dot does not cross line breaks, so a `$t(key, { ... })` options block containing a newline is treated as if it had no options. The block is then mis-split as formatters and the key is truncated, so the nested lookup runs without its options and the variables inside it never resolve.

The nesting regex already allows newlines inside `$t(...)`, so single-line and multiline options should behave the same. This adds the dotAll flag. The new test covers a multiline options block; existing single-line cases stay green.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -298,7 +298,7 @@ class Interpolator {
        *
        * Since v25.3.0 also this is possible: https://github.com/i18next/i18next/pull/2325
        */
-      const keyEndIndex = /{.*}/.test(match[1])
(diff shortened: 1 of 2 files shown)


adrai:
Thanks for the fix! Verified the bug reproduces and the fix behaves as described. Released in [v26.3.5](https://github.com/i18next/i18next/blob/master/CHANGELOG.md#2635) 🎉

PR #2441, opened 2026-07-08
fix: stop mutating the passed replace object when returning details

spokodev:
`getUsedParamsDetails` builds the returned `usedParams` from `options.replace` when a `replace` map is used, and adds `count` so consumers see it. It did so by writing `count` straight onto `options.replace`, mutating the caller's object. A caller that reuses one `replace` object across `t()` calls then carries a stale `count` into later interpolations.

This builds the params from a copy so the returned details still include `count` without touching the input. Adds a test asserting the passed `replace` object is not mutated.


adrai:
Thanks! Confirmed the stale `count` leak when reusing a `replace` object across calls — nice catch. Released in [v26.3.5](https://github.com/i18next/i18next/blob/master/CHANGELOG.md#2635) 🎉

PR #2442, opened 2026-07-08
fix: skip interpolation of placeholders inside escaped values

spokodev:
With `skipOnVariables` on (the default), after each replacement the loop advances `lastIndex` past the inserted value so a `{{...}}` carried inside that value is left alone. It advanced by the raw value length. When `escapeValue` is on (also the default) and the value contains escapable characters, the written text is longer than the raw value, so `lastIndex` lands inside the inserted text and a following `{{placeholder}}` gets interpolated, leaking another in-scope variable that should have stayed literal.

This advances by the escaped length that is actually written, and applies the regex-safe `$` transform only to the replacement argument so the `$`-doubling does not distort the length. The new test asserts a placeholder carried in a value with leading `&` characters stays literal; existing `$`, whitespace, and nesting cases stay green.

### DIFF
--- a/src/Interpolator.js
+++ b/src/Interpolator.js
@@ -189,12 +189,12 @@ class Interpolator {
       {
         // unescape if has unescapePrefix/Suffix
         regex: this.regexpUnescape,
-        safeValue: (val) => regexSafe(val),
+        safeValue: (val) => val,
       },
       {
         // regular escape on demand
         regex: this.regexp,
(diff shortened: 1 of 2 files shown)


adrai:
Thanks! Subtle one — verified the placeholder leak with escaped values and that the escaped-length advance fixes it without regressions. Released in [v26.3.5](https://github.com/i18next/i18next/blob/master/CHANGELOG.md#2635) 🎉

